Issue #60, opened 2026-02-04
Feature: Custom Seatbelt rule in profile.toml


qdrddr:
I'd like to be able to define custom Seatbelt rules in my toml profile. So I could for instance, granually be able to define Apple Events.

For example so I could send imessages using imsg tool
```
brew steipete/tap
brew install imsg
```

So from nono sbx i could run `imsg send --to "+10001114444" --text "Hi from OpenClawd"` with `(allow apple-event-send (target-bundle-id "com.apple.iChat"))` allowed in my toml profile.

rayjanoka:
Hey! nice tool!

I think I'm running into a similar issue trying to get `@playwright/mcp` working with `nono` enabled.

> nono uses macOS Seatbelt with (deny default). Chrome's rendering pipeline requires IOKit access (iokit-open,
  iokit-get-properties) for hardware queries — even in headless mode. nono's Seatbelt profile doesn't allow IOKit operations, and
  there's no CLI flag to add them. This causes Chrome to SIGSEGV when it tries to initialize the renderer.

PR #122, opened 2026-02-19
IOKit + mach-register investigation for Playwright/Chrome support

scp7:
## Summary

Investigation and reproduction of the Chromium/Playwright SIGSEGV under nono's Seatbelt sandbox, as reported by @rayjanoka in #60.

**Two findings:**

1. **IOKit is the primary blocker.** Chrome's renderer calls `iokit-open` and `iokit-get-properties` for GPU/display hardware queries — even in headless mode. Under `(deny default)` this causes `SEGV_ACCERR` (signal 11). Adding `(allow iokit-open)` + `(allow iokit-get-properties)` fixes the crash. Proposed solution: opt-in `--allow-iokit` CLI flag (never auto-enabled).

2. **`mach-register` is a separate base profile gap.** Chrome uses `bootstrap_check_in` to register a Mach port rendezvous server for its multi-process architecture. nono allows `mach-lookup` but not `mach-register`, causing `Permission denied (1100)`. This affects any multi-process tool, not just Chrome, so it belongs in the base Seatbelt profile alongside `mach-lookup`.

## What's included

- **`docs/development/iokit-support-plan.md`** — Full implementation plan: design rationale (blanket vs granular IOKit), file-by-file changes, verification steps
- **`scripts/test-iokit-step0.sh`** — Automated reproduction script (runs Chromium under two Seatbelt profiles — with/without IOKit)
- **`scripts/launch-chromium.mjs`** — Minimal Playwright launcher used by the test

No Rust code changes — this is findings + plan for review before implementation.

## Reproduction

```bash
cd scripts && npm install @playwright/test && npx playwright install chromium
cd .. && ./scripts/test-iokit-step0.sh
```

**Results** (Apple Silicon, macOS Sequoia):
- Without IOKit: `SEGV_ACCERR` signal 11 — crash
- With IOKit: exit 0 — Chrome loads `about:blank` successfully

## Questions for review

1. Is `(allow mach-register)` acceptable in the base profile? It's low-risk (normal IPC, `mach-priv*` still denied), but widens the default surface.
2. Should `--allow-iokit` be a standalone flag (proposed) or bundled into a `--allow-browser` / `--browser-compat` meta-flag?

Relates to #60

### DIFF
--- a/docs/development/iokit-support-plan.md
+++ b/docs/development/iokit-support-plan.md
@@ -311,10 +311,10 @@ cd nono && cargo build
 ./target/debug/nono run --help | grep iokit
 
 # FAIL: run without --allow-iokit (same crash as Step 0)
-./target/debug/nono run --allow . -- npx playwright chromium --headless about:blank
+./target/debug/nono run --allow . -- node scripts/launch-chromium.mjs
 
 # PASS: run with --allow-iokit
-./target/debug/nono run --allow . --allow-iokit -- npx playwright chromium --headless about:blank
+./target/debug/nono run --allow . --allow-iokit -- node scripts/launch-chromium.mjs
 
 # Verify dry-run output shows IOKit status
 ./target/debug/nono run --allow /tmp --allow-iokit --dry-run -- echo test


lukehinds:
really good analysis @scp7 , we can look to bring these into seatbelt and then shape a policy around them.

PR #136, opened 2026-02-20
Add network activity tracking to nono learn

EItanya:
Extend the learn command to capture outbound connections and listening ports alongside filesystem accesses. Network activity is discovered by tracing connect, bind, and sendto syscalls via strace.

Hostname resolution uses a three-tier strategy:
1. Timing correlation — attaches the hostname from the preceding DNS query directly to each connect() call, handling DNS round-robin
2. Forward DNS — resolves captured hostnames to build IP→hostname map
3. Reverse DNS — fallback for IPs with no other mapping

Supports both direct DNS (sendto to port 53) and systemd-resolved (Varlink JSON protocol over Unix socket) for hostname capture.

Adds --no-rdns flag to skip all DNS resolution. Output includes network sections in both summary and JSON formats, with a hint suggesting --net-block when network activity is detected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -575,6 +575,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "displaydoc"
+version = "0.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97369cbbc041bc366949bc74d34658d6cda5621039731c6310521892a3a20ae0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "dns-lookup"
 version = "2.1.1"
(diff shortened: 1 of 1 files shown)


lukehinds:
I like where this is going, and can see it's got valuable utility - this will fit nicely with the supervisor based proxy , for testing if filtering works!

lukehinds (approved):
I think we can get this, we cross-streamed while landing the network code, and while no conflicts we could add a few things - I don't mind if we do it in this PR or follow up with another:

There is now a  `network-policy.json` policy that lists allowed hosts (and denied will be included) learn could reference a host seen in a trace is within the network policy. We may also be able to dump the commands output to a network policy file (json). Someone could leave learn running and get it to populate an policy list. 

Overall it's a clean fit - learn discovers, proxy enforces.

PR #184, opened 2026-02-28
fix: writes setup profiles to the correct directory on macOS

sit:
On macOS, setup.rs uses dirs::config_dir() which returns ~/Library/Application Support while the profile loader uses resolve_user_config_dir() which returns a hard-coded ~/.config. Thus profiles generated by `nono setup --profiles` were never found by `nono run --profile`.

Replace dirs::config_dir() in setup.rs with the same resolve_user_config_dir() the loader uses. Add round-trip test that runs setup_profiles() then loads the result by name.

Users who previously ran `nono setup --profiles` on macOS should remove the stale directory: rm -rf ~/Library/Application\ Support/nono

---

There are a few other places that use `dirs::config_dir()` which might have similar issues but I didn't run into / check those.

### DIFF
--- a/crates/nono-cli/src/setup.rs
+++ b/crates/nono-cli/src/setup.rs
@@ -515,6 +515,9 @@ mod tests {
     /// `resolve_user_config_dir()` returning `~/.config`. This test catches that.
     #[test]
     fn test_setup_profiles_loadable_by_name() {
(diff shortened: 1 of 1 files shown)


lukehinds:
Thanks for the fix @sit , geminis point is worth patching - I actually hit this before in tests (maybe gemini has a memory).

PR #189, opened 2026-03-01
feat!: add AccessMode::Interactive, restrict IoctlDev to TTY devices

### LINKED ISSUE #192: Unify terminology for interactive execution mode (--exec vs interactive: true)
Currently, nono has two different ways to configure an application to run in Direct execution mode (which preserves TTY control for rich CLI apps by bypassing the monitor process):

1. Via the CLI flag: `--exec`
2. Via the profile JSON config: `"interactive": true`

This dual terminology is confusing for users. There is no obvious mental mapping between the `--exec` CLI flag and the `"interactive"` profile setting, even though they control the exact same underlying execution strategy (`ExecStrategy::Direct`). Furthermore, the term "exec" describes the underlying system call mechanism (execve) rather than the intended user experience (running an interactive TUI).

Proposal:

We should unify the terminology around the word interactive to clearly describe the intended use case.

1. Rename the CLI Flag: Change the primary CLI flag from `--exec `to `--interactive` (and perhaps add a -i shorthand).
2. Keep the Profile Field: Leave `"interactive": true `unchanged in the JSON profiles.
3. Maintain Backwards Compatibility: Keep `--exec` as a hidden alias in the clap parser `(#[arg(long, alias = "exec")])` so existing user scripts do not break, but remove it from the primary help documentation.
4. Update Documentation: Update the CLI help text to clarify the behavior. Example:
   > `--interactive:` Run in interactive mode. Preserves TTY and terminal control for rich CLI apps (e.g., Claude Code, vim). Disables the diagnostic footer. (Equivalent to interactive: true in profiles).

This change will make the mental model much simpler: "If my app has a complex UI, I need to tell nono it's interactive, either via the flag or the profile."

Yus314:
## Summary

- Add `AccessMode::Interactive` variant (ReadWrite + IoctlDev on Landlock V5+)
- Remove `IoctlDev` from `AccessMode::Read` — restricts device ioctl to TTY paths only
- Rename `--exec` CLI flag to `--interactive` (`-i`), keeping `--exec` as hidden alias
- Add `#[non_exhaustive]` to `AccessMode` enum

Addresses review feedback from @lukehinds on the original approach (granting IoctlDev
to all Read paths was too broad).

Closes #192

## Problem

The original approach added `IoctlDev` to `AccessMode::Read`, which granted device
ioctl permissions to **every** read-accessible path. This was too broad — only TTY/PTY
devices (`/dev/tty`, `/dev/pts/*`) actually need ioctl for interactive apps.

## Solution

1. **New `AccessMode::Interactive`**: ReadWrite + IoctlDev (Linux only). Applied only
   to TTY device paths via `CapabilitySet::upgrade_tty_to_interactive()`.
2. **User intent protection**: Paths with `CapabilitySource::User` are never
   auto-upgraded — if a user explicitly set `/dev/tty` to `Read`, that's respected.
   Users who want interactive access on a specific path can set it via profile config.
3. **CLI unification**: `--exec` → `--interactive` (`-i`). The profile field
   `interactive: true` already existed, so terminology is now consistent.
4. **Activation**: `upgrade_tty_to_interactive()` is called only when `--interactive`
   flag is passed or profile sets `interactive: true`. The library never auto-upgrades
   on its own — the CLI owns this policy decision.

### TTY path scope

Only `/dev/tty` and `/dev/pts/*` are upgraded, per @lukehinds' guidance. `/dev/ptmx`
is excluded because it is the multiplexer (opens new PTYs) rather than an existing
terminal device — granting `IoctlDev` on it would be overly broad. `/dev/console` is
excluded as it is rarely present in container environments.

## Security Considerations

- **Tighter than before**: IoctlDev is no longer on all Read paths; only on TTY devices
  when `--interactive` is active
- **IoctlDev permits ALL ioctls**: including TIOCSTI (terminal input injection). The
  `--interactive` flag doc warns about weakened isolation
- **macOS**: `file-ioctl` is already granted to all allowed paths in the existing
  Seatbelt profile generation. `Interactive` does not change this behavior — it maps
  to ReadWrite for the read/write/map-exec classification only. This PR's IoctlDev
  isolation is Linux-specific.
- **Landlock < V5**: `restrict_self()` returns `PartiallyEnforced` (same pattern as
  `Truncate` on < V3). On older kernels, `--interactive` will not grant ioctl and
  interactive apps requiring terminal ioctl may fail. This is a fail-secure degradation.

### Why `#[non_exhaustive]` now?

Adding `Interactive` is already a breaking change (downstream exhaustive matches break).
Adding `#[non_exhaustive]` at the same time means future variants won't cause another
breaking bump — a single migration cost for downstream instead of repeated ones.

## Breaking Changes

- `AccessMode` is now `#[non_exhaustive]` — downstream exhaustive matches need `_ =>` arm
- New `AccessMode::Interactive` variant
- `--exec` flag renamed to `--interactive` (`--exec` remains as hidden alias)
- FFI: `NONO_ACCESS_MODE_INTERACTIVE = 3` added

Downstream bindings (nono-py, nono-ts) will need updates.

## Files Changed (21)

| Area | Files | Changes |
|------|-------|---------|
| Library | `capability.rs` | Interactive variant, `#[non_exhaustive]`, `upgrade_tty_to_interactive()`, dedup logic |
| Library | `sandbox/linux.rs` | Remove IoctlDev from Read, add Interactive arm |
| Library | `sandbox/macos.rs` | Interactive = ReadWrite in all match arms |
| Library | `query.rs`, `diagnostic.rs`, `state.rs` | Interactive support |
| CLI | `cli.rs` | `--interactive` / `-i` (alias `--exec`) |
| CLI | `main.rs` | TTY upgrade calls, remove `direct_exec` field |
| CLI | `output.rs`, `terminal_approval.rs`, `sandbox_state.rs`, `query_ext.rs`, `exec_strategy.rs`, `policy.rs`, `profile/mod.rs` | Interactive in all AccessMode matches |
| FFI | `types.rs`, `nono.h` | `NONO_ACCESS_MODE_INTERACTIVE = 3` |
| Docs | `flags.mdx`, `execution-modes.mdx`, `network-proxy.mdx`, `profiles-groups.mdx` | --exec → --interactive |

## Test Plan

- [x] `make ci` passes (clippy + fmt + all tests)
- [x] New tests: dedup (Interactive absorbs Read/Write/ReadWrite), TTY upgrade,
  user intent preservation, idempotency, Landlock conversion, display
- [x] Read no longer includes IoctlDev
- [x] ReadWrite does not include IoctlDev
- [x] Interactive = ReadWrite | IoctlDev
- [ ] Manual: `nono run --interactive --allow /dev/tty -- vim` on Landlock V5 kernel

### DIFF
--- a/bindings/c/include/nono.h
+++ b/bindings/c/include/nono.h
@@ -24,6 +24,8 @@
 
 #define NONO_ACCESS_MODE_READ_WRITE 2
 
+#define NONO_ACCESS_MODE_INTERACTIVE 3
+
 /**
  * Sentinel value returned on error (NULL pointer, out-of-bounds index).
  */
--- a/bindings/c/src/types.rs
+++ b/bindings/c/src/types.rs
@@ -14,6 +14,7 @@ use std::os::raw::c_char;
 pub const NONO_ACCESS_MODE_READ: u32 = 0;
 pub const NONO_ACCESS_MODE_WRITE: u32 = 1;
 pub const NONO_ACCESS_MODE_READ_WRITE: u32 = 2;
+pub const NONO_ACCESS_MODE_INTERACTIVE: u32 = 3;
 /// Sentinel value returned on error (NULL pointer, out-of-bounds index).
 pub const NONO_ACCESS_MODE_INVALID: u32 = u32::MAX;
 
@@ -25,6 +26,7 @@ pub fn validate_access_mode(raw: u32) -> Option<nono::AccessMode> {
         NONO_ACCESS_MODE_READ => Some(nono::AccessMode::Read),
         NONO_ACCESS_MODE_WRITE => Some(nono::AccessMode::Write),
         NONO_ACCESS_MODE_READ_WRITE => Some(nono::AccessMode::ReadWrite),
+        NONO_ACCESS_MODE_INTERACTIVE => Some(nono::AccessMode::Interactive),
         _ => None,
     }
 }
@@ -36,6 +38,8 @@ pub fn access_mode_to_raw(mode: nono::AccessMode) -> u32 {
         nono::AccessMode::Read => NONO_ACCESS_MODE_READ,
         nono::AccessMode::Write => NONO_ACCESS_MODE_WRITE,
         nono::AccessMode::ReadWrite => NONO_ACCESS_MODE_READ_WRITE,
+        nono::AccessMode::Interactive => NONO_ACCESS_MODE_INTERACTIVE,
+        _ => NONO_ACCESS_MODE_INVALID,
     }
 }
 
--- a/crates/nono-cli/src/cli.rs
+++ b/crates/nono-cli/src/cli.rs
@@ -335,8 +335,10 @@ pub struct RunArgs {
 
     /// Preserve TTY for interactive apps (e.g., Claude Code, vim, htop).
     /// Without this flag, nono monitors output which can break interactive UIs.
-    #[arg(long = "exec")]
-    pub direct_exec: bool,
+    /// Also upgrades TTY device paths to Interactive access mode (adds IoctlDev
+    /// on Linux Landlock V5+). WARNING: this weakens isolation for TTY devices.
+    #[arg(long = "interactive", short = 'i', alias = "exec")]
+    pub interactive: bool,
 
     /// Enable atomic rollback snapshots for the session.
     /// Takes content-addressable snapshots of writable directories so you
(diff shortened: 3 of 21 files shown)


lukehinds:
will close this now as is now obsolete relative to current main - current code selectively adds IoctlDev only for actual device paths under /dev when  the capability is Write or ReadWrite

PR #243, opened 2026-03-04
feat: experimental localhost relay with --allow-port

### LINKED ISSUE #235: Feature: expose additional localhost ports to sandbox with network proxy
To support running AI agents, MCP servers, and other dev tools in separate sandboxes, we need to allow them to communicate with each other. Currently, network proxy works in three modes: allow all, `--net-block` to block all, or `--proxy-allow example.com` to allow domains based on DNS resolution.

Proposed solution: add a `--allow-port <port>` CLI argument that exposes `127.0.0.1:<port>` between the host and sandbox process without passing through the proxy. Naming is hard, so feel free to suggest a better CLI flag.

josephgimenez:
## Summary
- add experimental localhost relay mode for `nono run` with explicit `--allow-port <port>` grants
- route localhost HTTP and CONNECT traffic through the parent proxy and enforce per-port allow checks
- keep relay scope intentionally narrow for MVP (`--net-block` required, incompatible with profile/credential/external proxy options)
- override `NO_PROXY/no_proxy` and set `NONO_LOCALHOST_RELAY=1` so localhost requests are actually filtered by relay policy

## Why Relay/Bridging (Linux + macOS)
Issue #235 asks for sandbox-to-sandbox localhost communication without broad network access.

Relay/bridging through the parent proxy was chosen because it gives one security model that works on both Linux and macOS:

- Unified enforcement point:
  - The child only gets OS permission to connect to `localhost:<proxy-port>` (`ProxyOnly` mode).
  - The proxy then enforces which localhost destination ports are allowed (`--allow-port`).
- Linux compatibility:
  - Landlock can do per-port TCP rules, but direct per-port child access would bypass existing proxy controls (host filtering, deny logic, audit path).
  - Reusing proxy mediation keeps policy behavior consistent with existing network controls.
- macOS compatibility:
  - Seatbelt supports the `ProxyOnly` pattern (`allow network-outbound remote tcp localhost:PORT`), which is already how nono enforces filtered outbound networking.
  - This lets us provide the same user-facing semantics on macOS without a Linux-only feature split.
- Security properties preserved:
  - Default deny: only explicitly allowed localhost destination ports pass.
  - Session token auth still protects proxy access.
  - Clearing `NO_PROXY/no_proxy` in relay mode prevents localhost bypasses around policy.

### 1) Connect to a host-local service from a sandbox
```bash
# Host process listening on 11435
python3 -m http.server 11435

# Sandbox can only reach localhost:11435 (and no general outbound network)
nono run \
  --allow-cwd \
  --net-block \
  --experimental-localhost-relay \
  --allow-port 11435 \
  -- curl http://localhost:11435/
```

### 2) Sandbox-to-sandbox communication
```bash
# Sandbox A (server)
nono run \
  --allow-cwd \
  --network-profile minimal \
  --allow-bind 11439 \
  -- python3 -m http.server 11439

# Sandbox B (client) can connect only to allowed localhost port
nono run \
  --allow-cwd \
  --net-block \
  --experimental-localhost-relay \
  --allow-port 11439 \
  -- curl http://localhost:11439/
```

### 3) Expected deny behavior
```bash
# Only 11440 is allowed, so access to 11439 is denied by relay policy
nono run \
  --allow-cwd \
  --net-block \
  --experimental-localhost-relay \
  --allow-port 11440 \
  -- curl -i http://localhost:11439/
# => HTTP/1.1 403 Forbidden
```

## Manual Testing (Detailed)
Environment:
- OS: macOS 26 (arm64)
- Binary: `target/debug/nono` (built from this branch)
- Date: 2026-03-04 UTC

Automated checks:
- `make check`
- `cargo test -p nono-proxy`
- `cargo test -p nono-cli -- --test-threads=1`

Manual matrix:

| Case | Command (abbrev) | Expected | Observed |
|---|---|---|---|
| Host positive | `nono run --net-block --experimental-localhost-relay --allow-port 11435 -- curl http://localhost:11435/` | allow | `HTTP 200`, body `host-relay-ok` |
| Host negative | same target, but `--allow-port 11436` | deny | `HTTP 403` |
| No relay baseline | `nono run --net-block -- curl http://localhost:11435/` | blocked | curl exit code `7` |
| Relay env override | `nono run ... -- sh -c 'echo ...'` | localhost forced through proxy | `NO_PROXY=` / `no_proxy=` empty, `NONO_LOCALHOST_RELAY=1` |
| Sandbox-to-sandbox positive | client `--allow-port 11439` to sandbox server on 11439 | allow | `HTTP 200`, body `sandbox-relay-ok` |
| Sandbox-to-sandbox negative | client `--allow-port 11440` to sandbox server on 11439 | deny | `HTTP 403` |
| Old flag rejection | `--expose-port` | fail parse | `unexpected argument '--expose-port'` |
| Shell scope guard | `nono shell --experimental-localhost-relay --allow-port ...` | rejected | config parse error: supported only for `nono run` |

Closes #235

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,6 +142,21 @@ jobs:
             echo "No integration tests found, skipping"
           fi
 
+  audit:
+    name: Cargo Audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable
+
+      - name: Install cargo-audit
+        run: cargo install cargo-audit
+
+      - name: Run cargo audit
+        run: cargo audit
+
   docs-checks:
     name: Docs Checks
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,40 @@
 # Changelog
 
+## [0.12.0] - 2026-03-07
+
+### Bug Fixes
+
+- Resolve dirfd-relative paths in seccomp-notify handler (#262) (#277) ([#277](https://github.com/always-further/nono/pull/277))
+
+- Show platform-correct path in user-level policy warning (#263) ([#263](https://github.com/always-further/nono/pull/263))
+
+- Enforce macOS signal isolation via Seatbelt (#264) ([#264](https://github.com/always-further/nono/pull/264))
+
+- *(profile)* Allow clearing inherited network profiles (#252) ([#252](https://github.com/always-further/nono/pull/252))
+
+
+### Documentation
+
+- *(readme)* Update latest release note (#253) ([#253](https://github.com/always-further/nono/pull/253))
+
+
+### Features
+
+- Add port_allow to profile JSON NetworkConfig (#254) (#276) ([#276](https://github.com/always-further/nono/pull/276))
+
+- Context-aware diagnostic banner for sandbox failures (#275) ([#275](https://github.com/always-further/nono/pull/275))
+
+- *(cli)* Add --net-allow override (#251) ([#251](https://github.com/always-further/nono/pull/251))
+
+- Add macOS learn mode using fs_usage and profile save prompt (#244) ([#244](https://github.com/always-further/nono/pull/244))
+
+
+### Miscellaneous
+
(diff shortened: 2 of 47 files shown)


josephgimenez:
Hey @lukehinds no worries at all!

PR #243 was trying to solve a smaller, more specific problem: localhost relay and per-port control with --allow-port, but it sounds like we already have this covered now.

It wasn’t meant to solve phantom credential injection for CONNECT, which turns out to be a separate and more involved problem.

Given that, I think it’s totally reasonable to close this PR for now and revisit it later if the localhost relay piece is still something we want. Then we can look at phantom credential support for CONNECT as its own design discussion, with a clearer sense of the options and tradeoffs.

PR #248, opened 2026-03-05
feat: add --allow-port for bidirectional localhost IPC between sandboxes

lukehinds:
feat: add --allow-port for bidirectional localhost IPC between sandboxes
    
Allow sandboxed processes to communicate over specific localhost TCP
ports. The sandboxed process can both connect to and bind/listen on
127.0.0.1:<port>, regardless of network mode (Blocked, ProxyOnly,
AllowAll)

Use case: running AI agents, MCP servers, and dev tools in separate
nono sandboxes that need to talk to each other.
    
Platform behavior:
- Linux: per-port ConnectTcp + BindTcp via Landlock
- macOS: per-port outbound via Seatbelt, blanket bind/inbound
(same tradeoff as --allow-bind)
    
 Signed-off-by: Luke Hinds <lukehinds@gmail.com>

### DIFF
--- a/crates/nono-cli/src/capability_ext.rs
+++ b/crates/nono-cli/src/capability_ext.rs
@@ -153,6 +153,11 @@ impl CapabilitySetExt for CapabilitySet {
             });
         }
 
+        // Localhost IPC ports
+        for port in &args.allow_port {
+            caps.add_localhost_port(*port);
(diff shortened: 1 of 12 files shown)


scp7 (approved):
Looks good

PR #283, opened 2026-03-08
docs: fix broken links and stale examples

josephgimenez:
## Summary
- fix public docs links that were pointing at non-existent `docs.nono.sh` routes
- update stale CLI docs for `nono learn --json`, profile-based `env_credentials`, and current network profiles
- clarify the difference between `--allow-cwd` and `--allow .` so read+write current-directory examples use `--allow .`
- clarify trust docs around `AGENTS.md`

### `--allow-cwd` vs `--allow .`
Why `--allow-cwd` exists even though it does not escalate permissions:
- It tells nono: if this run would auto-share the working directory, do it without asking me.
- It uses the access level already defined by policy:
  - the profile's `workdir.access` if a profile is active
  - otherwise the default CWD behavior, which is read-only

Why that matters:
- In interactive runs, it skips the CWD-sharing prompt.
- In non-interactive runs, the prompt cannot be answered, so without `--allow-cwd` the CWD may be omitted.
- In `--silent` runs, nono errors instead of prompting, so `--allow-cwd` is required if you want automatic CWD sharing.

When to use each flag:
- Use `--allow-cwd` when you want to accept the existing profile/default workdir behavior without an interactive prompt.
- Use `--allow .` when you want to explicitly grant read+write access to the current directory.

Original doc example:
```bash
nono run --allow-cwd -- sh -c "echo test > ./allowed.txt"
```

Tested from a disposable temp directory under the repo:
```console
$ cargo run -q --manifest-path /Users/josephgimenez/workspace/security/nono/Cargo.toml -p nono-cli -- run --allow-cwd -- sh -c 'echo test > ./allowed.txt'

Capabilities:
  Filesystem:
    /Users/josephgimenez/workspace/security/nono/.tmp-doc-test.cl9d99 [read] (dir)
    + 33 system/group paths (use -v to show)
  Network:
    outbound: allowed

Supervised mode: child sandboxed, parent manages supervisor.

Applying Kernel sandbox protections.
/bin/sh: ./allowed.txt: Operation not permitted

[nono] Command exited with code 1. This may be due to sandbox restrictions.
[nono]
[nono] Sandbox policy:
[nono]   Allowed paths:
[nono]     /Users/josephgimenez/workspace/security/nono/.tmp-doc-test.cl9d99 (read, dir)
[nono]     + 33 system/profile path(s)
[nono]   Network: allowed
[nono]
[nono] To grant additional access, re-run with:
[nono]   --allow <path>     read+write access to directory
[nono]   --read <path>      read-only access to directory
[nono]   --write <path>     write-only access to directory
```

Updated doc example:
```bash
nono run --allow . -- sh -c "echo test > ./allowed.txt"
```

Tested from a disposable temp directory under the repo:
```console
$ cargo run -q --manifest-path /Users/josephgimenez/workspace/security/nono/Cargo.toml -p nono-cli -- run --allow . -- sh -c 'echo test > ./allowed.txt' && ls -l allowed.txt && cat allowed.txt

Capabilities:
  Filesystem:
    /Users/josephgimenez/workspace/security/nono/.tmp-doc-test.IXiBeC [read+write] (dir)
    + 33 system/group paths (use -v to show)
  Network:
    outbound: allowed

Supervised mode: child sandboxed, parent manages supervisor.

Applying Kernel sandbox protections.
-rw-r--r--@ 1 josephgimenez  staff  5 Mar  8 10:56 allowed.txt
test
```

Additional checks:
- `cargo run -q -p nono-cli -- run --help`
- `cargo run -q -p nono-cli -- shell --help`
- `cargo run -q -p nono-cli -- why --help`
- `cargo run -q -p nono-cli -- trust --help`
- `cargo run -q -p nono-cli -- rollback --help`
- `cargo run -q -p nono-cli -- audit --help`
- `cargo run -q -p nono-cli -- setup --help`
- `cargo run -q -p nono-cli -- learn --help`

## Notes
- `make ci` still fails on clean `main` because `cargo test -p nono-cli` currently has a pre-existing failing test in `crates/nono-cli/src/main.rs` (`test_check_sensitive_path`). This PR does not change code or test behavior.

### DIFF
--- a/crates/nono-cli/README.md
+++ b/crates/nono-cli/README.md
@@ -43,6 +43,9 @@ nono run --allow-cwd --net-block -- command
 # Use a built-in profile
 nono run --profile claude-code -- claude
 
+# Use the Codex profile
+nono run --profile codex -- codex
+
 # Keep a profile but temporarily allow unrestricted network
 nono run --profile claude-code --net-allow -- claude
 
@@ -61,6 +64,7 @@ nono run --allow-cwd --dry-run -- command
 | Profile | Command |
 |---------|---------|
 | Claude Code | `nono run --profile claude-code -- claude` |
+| Codex | `nono run --profile codex -- codex` |
 | OpenCode | `nono run --profile opencode -- opencode` |
 | OpenClaw | `nono run --profile openclaw -- openclaw gateway` |
 
--- a/docs/cli/features/credential-injection.mdx
+++ b/docs/cli/features/credential-injection.mdx
@@ -1,6 +1,6 @@
 ---
 title: Credential Injection
-description: Keep API keys out of the sandbox with proxy-based credential injection or environment variable injection from the system keystore or 1Password
+description: Keep API keys out of the sandbox with proxy-based credential injection or environment variable injection from the system keystore, 1Password, or Apple Passwords
 ---
 
 nono provides two ways to keep credentials out of the sandboxed process:
@@ -79,7 +79,7 @@ echo -n "your-gemini-key" | secret-tool store --label="nono: gemini_api_key" \
 
 ## 1Password Integration
 
-nono supports [1Password](https://1password.com/) `op://` URIs as a credential source anywhere you would use a keyring account name. This works with both `--env-credential` (CLI) and profile-based credentials (`env_credentials`, `custom_credentials`). The 1Password CLI (`op`) must be installed and authenticated.
+nono supports [1Password](https://1password.com/) `op://` URIs as a credential source anywhere you would use a keyring account name. For CLI env injection, use `--env-credential-map <op://...> <ENV_VAR>`. This also works in profile-based credentials (`env_credentials`, `custom_credentials`). The 1Password CLI (`op`) must be installed and authenticated.
 
 ### Finding Your Secret Path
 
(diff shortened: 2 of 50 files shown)


crates/nono-cli/README.md
```diff
@@ -29,7 +29,7 @@ cargo build --release
 
 ```bash
 # Allow read+write to current directory
-nono run --allow-cwd -- command
```
lukehinds, line 32: I think --allow-cwd is still a working flag?

Image

docs/cli/features/credential-injection.mdx
```diff
@@ -135,7 +135,7 @@ In a profile's `env_credentials` section, use an `op://` URI as the key instead
 ```
 
 ```bash
-nono run --profile my-agent --env-credential -- my-agent
```
lukehinds, line 136: good catch, that's quite sloppy!

josephgimenez: I have to admit -- I just asked gpt 5.4 to go through all of our docs and test every command, so I did not by any means catch these myself.  i wonder if there's a way to have gemini actually compare the PR changes with what's in our docs and automatically ensure they're in sync.  Maybe it does already and this is pre-gemini PR review stuff.

crates/nono-cli/README.md
josephgimenez: oof, im validating this now and trying to figure out why this was detected as needing fixed.  My apologies for the false negative here.

josephgimenez: I updated the PR description to reflect this, but it appears that `--allow-cwd` does **not** override any permissions for the cwd, which contradicts the docs, which says it sets read+write permissions.  It appears that what the docs are referring to is what `--allow .` actually does.

josephgimenez: Thinking about this for a minute, part of the confusion is probably in the flag name itself.  

Based on the behavior we have for `--allow` where it enables r/w permissions, maybe we should update the name to `--allow-rw`, which would make the intended behavior more clear and further differentiate it from `--allow-cwd` when comparing:

- `--allow .`
- `--allow-cwd`

it's kind of ambiguous without the `-rw` suffix in `--allow-rw`, i think?

josephgimenez: @lukehinds based on this, if you're cool with it, i can open a separate PR to support `--allow-rw` and then update the docs to be more explicit about when to use one or the other.

I kind of want to take it one step further and remove `--allow-cwd` altogether, and we can consider:
- `--allow .` - skips prompt, uses permissions from profile
- `--allow-rw .` - skips prompt, elevates permissions to r/w even if profile is read-only.

lukehinds: right, got it now! `--allow-cwd` means "share the current directory automatically, with whatever access level the profile/defaults already decided and don't nag me with a prompt" - I had forgotten

I think you're onto something, allow-rw is right. `--allow .` does still read like a permission grant - it needs a connotation of skip somehow , something like `--no-cwd-prompt` (that feels too long)

josephgimenez: so maybe something like:
- `--allow-rw <path>` 
  - originally `--allow` and concretely defines that we want r/w access, regardless of base permissions
-  `--include-workdir`
   - originally `--allow-cwd` but this makes it clear this is not a permission grant specifically; it could be read or read-write, depending upon workdir permissions (or lack thereof) in the profile.
   - this also improves upon using `cwd`, which could be inaccurate, since the workdir can be overridden to be any particular directory and not just cwd? (pointed out by codex)

lukehinds: sounds good! do you want me to merge this, or just make the doc changes in the next PR?

josephgimenez: I think it makes sense to merge this, and then i'll have a follow-up PR that modifies the flags to match the pattern above.

PR #302, opened 2026-03-09
refactor(cli): standardize flags to verb-noun ordering

josephgimenez:
## Why
The network-related CLI flags were the main holdouts from nono's otherwise verb-noun command style (`--allow-port`, `--allow-bind`, `--block-command`, etc.). Renaming them makes the interface easier to scan and remember, and it removes an inconsistent naming pocket from the docs, diagnostics, and examples.

## Summary
- rename the canonical network flags to `--block-net` and `--allow-net`
- rename the canonical proxy host flag to `--allow-proxy`
- update docs, examples, diagnostics, setup text, and integration tests to use the new spellings
- add a regression test covering `--allow-net` vs `--allow-proxy` conflict handling

## Compatibility
- keep the old flag spellings (`--net-block`, `--net-allow`, `--proxy-allow`) as hidden clap aliases
- normalize legacy `NONO_NET_BLOCK` / `NONO_NET_ALLOW` env vars to the new canonical names before parsing

## Testing
- `cargo test -p nono-cli -p nono`
- `cargo test -p nono-cli`
- `cargo test -p nono-cli cli::tests::test_allow_net_conflicts_with_allow_proxy`
- `cargo build --release && export NONO_BIN="$PWD/target/release/nono" && export PATH="$PWD/target/release:$PATH" && bash tests/integration/test_network.sh && bash tests/integration/test_shell.sh && bash tests/integration/test_policy_queries.sh && bash tests/integration/test_client_startup.sh`

### DIFF
--- a/crates/nono-cli/src/cli.rs
+++ b/crates/nono-cli/src/cli.rs
@@ -279,6 +279,7 @@ pub struct SandboxArgs {
             "allow_proxy",
             "proxy_credential",
             "external_proxy",
+            "external_proxy_bypass",
             "proxy_port"
         ]
     )]
@@ -320,8 +321,19 @@ pub struct SandboxArgs {
 
     /// Chain through an external (enterprise) proxy.
     /// Format: host:port (e.g., squid.corp.internal:3128)
-    #[arg(long, value_name = "HOST:PORT")]
+    #[arg(long, value_name = "HOST:PORT", env = "NONO_EXTERNAL_PROXY")]
     pub external_proxy: Option<String>,
+
+    /// Domains to route directly instead of through the external proxy.
+    /// Supports exact hostnames and wildcards (e.g., *.internal.corp).
+    /// Can be specified multiple times. Requires --external-proxy (or profile equivalent).
+    #[arg(
+        long,
+        value_name = "HOST",
+        env = "NONO_EXTERNAL_PROXY_BYPASS",
+        value_delimiter = ','
+    )]
+    pub external_proxy_bypass: Vec<String>,
 
     /// Fixed port for the credential injection proxy (default: OS-assigned).
     /// Use this when the sandboxed application requires a known proxy port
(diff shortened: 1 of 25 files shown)


crates/nono-cli/src/cli.rs
```diff
@@ -252,32 +252,34 @@ pub struct SandboxArgs {
...
+        long = "allow-net",
+        alias = "net-allow",
+        env = "NONO_ALLOW_NET",
         value_parser = clap::builder::BoolishValueParser::new(),
         action = clap::ArgAction::SetTrue,
         conflicts_with_all = [
-            "net_block",
+            "block_net",
             "network_profile",
             "proxy_allow",
```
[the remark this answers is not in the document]
josephgimenez: I verified this against the actual parser before changing anything: `target/debug/nono run --allow /tmp --allow-net --allow-proxy api.openai.com --dry-run echo` already fails with the expected conflict. To pin that behavior down in the suite, I added `test_allow_net_conflicts_with_allow_proxy` in `crates/nono-cli/src/cli.rs`.

lukehinds: should this one be `allow_proxy` too?

lukehinds: * line 276

josephgimenez: welp, this is somewhat embarrassing.  you have the eyes of an eagle, luke.  thanks!

josephgimenez: I updated the CLI arg field and direct call sites from `proxy_allow` to `allow_proxy`, while leaving the profile schema (`network.proxy_allow`) unchanged since this PR is only standardizing the CLI surface.

josephgimenez: ^ this makes me wonder if we should adjust the policy schema too.  maybe not if we're aiming to migrate to a Nonofile pattern in the RFC you posted.

lukehinds: I don't think the nonofile will happen for a while mate, I want to keep things stable for users for a while and hopefully let the numbers swell, so you are all good buddy

PR #320, opened 2026-03-10
fix: add pnpm paths to policy.json

wrgore:
Adding pnpm paths to policy.json. This allows pnpm support for MacOS and Linux.


wrgore:
If this is acceptable, I will update the profile-groups.mdx and claude-code.mdx documentation to reflect the change.

PR #399, opened 2026-03-16
feat: support multiple base profiles in extends field

### LINKED ISSUE #391: Makes extends support multiple groups

extends is `Option<String>`, so it only supports a single base profile. It doesn't accept an array.

You'd need to pick one base and add the other's groups manually:

```json
{
  "extends": ["claude-code", "opencode"]
  .....
  }
}
```

the ability to chain groups

_No response_

### How is this blocking you?

Just an idea — low urgency

### Additional context

_No response_

kipz:
Accept both `"extends": "single"` and `"extends": ["a", "b"]` via serde untagged enum deserialization. Multiple bases are resolved recursively and fold-merged left-to-right before the child overrides.

Diamond inheritance is intentionally rejected — the visited set is shared across sibling bases to prevent confusing permission accumulation.

Closes #391

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "ahash"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


kipz:
@lukehinds yeah, the left-to-right fold semantics is what I was worried about above. My guess was that folks would not expect the ordering in the `extends` list to affect the result, but if you're cool with it, I'm happy to fix up the PR as you suggest.

PR #488, opened 2026-03-24
fix: block Unix socket connections via add_deny_access; add add_deny_commands

Austio:
Closes https://github.com/always-further/nono/issues/483

## Problem

`add_deny_access` on a Unix socket path doesn't actually block connections to it in mac.

On macOS, when a process calls connect(2) on a socket, Seatbelt treats that as a network-outbound operation (not a file operation) so the file deny rules we emit never fire.  This means a sandboxed process with /var/run/docker.sock in its deny list could still talk to the Docker daemon and do whatever it wanted with containers.  In practice this gives the sandbox a vector to reach into the host if it can find an open socket that it can communicate with.

## Fix

1. Emit a network-outbound deny for socket paths                                                                                                                                                       
   
emit_deny_rules now emits a fourth Seatbelt rule alongside the existing three.  For example with an `add_deny_access` including `\private\var\run\docker.sock` we would get this:

```                                                                                                                                                                                          
  (deny network-outbound (path "/private/var/run/docker.sock"))                                                                                                                                          
```

A few things worth noting about this rule:                                                                                                                                                             
 - It uses (path ...) for an exact match rather than (subpath ...) — socket connections match on the precise path, not a prefix              
 - Seatbelt evaluates this at connect(2) time, not when the sandbox initializes, so it blocks connections to sockets that don't exist yet when the sandbox starts

2. Add add_deny_commands to profiles                                                                                                                                                                   
                                                                                                                                                                                                         
Added add_deny_commands to PolicyPatchConfig as defense-in-depth — you can now block specific binaries directly from a profile without touching policy.json:                                           

```                                                                                                                                                                                                       
{                                                                                                                                                                                                      
    "extends": "claude-code",                                                                                                                                                                          
    "policy": {                                                                                                                                                                                          
      "add_deny_access": ["/var/run/docker.sock"],
      "add_deny_commands": ["docker", "docker-compose", "podman", "kubectl"]                                                                                                                             
    }                                                                                                                                                                                                  
}
```

Wired up through merge logic, CapabilitySet::from_profile, policy show, policy diff, JSON output, and the profile new template.                                                                                                                          

## Verification of `add_deny_access`

Summary: Proves can access docker socket without a policy and denied with policy applied.

### Given this policy

```
{                                                                                                                                                                                                      
  "meta": { "name": "deny-docker-test" },
  "policy": {                                                                                                                                                                                          
    "add_deny_access": [                                                                                                                                                                             
      "/var/run/docker.sock"                                                                                                                                                                           
    ]
  }                                                                                                                                                                                                    
}  
```

### Validate Can Access Docker w/o Policy

```
./nono run -- docker ps
...

CONTAINER ID   IMAGE              COMMAND              CREATED       STATUS        PORTS                                         NAMES
```

### Show Policy

```
./nono policy show $(pwd)/deny-docker.json 
nono policy: profile '/Users/austin/projects/dox/nono/target/debug/deny-docker.json'

  Security groups:
    deny_credentials
    ...other stuff

  Policy patches:
    add_deny_access:
      /var/run/docker.sock
```

### No access to Docker

```
./nono run --profile $(pwd)/deny-docker.json -- docker ps

permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.51/containers/json": dial unix /var/run/docker.sock: connect: operation not permitted
```

## Verification of `add_deny_command`

```
{                                                                                                                                                                                                      
  "meta": { "name": "deny-docker-test" },
  "policy": {                                                                                                                                                                                          
    "add_deny_commands": [                                                                                                                                                                             
      "docker"                                                                                                                                                                           
    ]
  }                                                                                                                                                                                                    
}
```

```
./nono policy show $(pwd)/deny-docker.json 

  Security groups:
    deny_credentials
    ....other stuff

  Policy patches:
    add_deny_commands: docker
```

```
$(pwd)/deny-docker.json -- docker ps
────────────────────────────────────

2026-03-24T16:50:02.680824Z ERROR Command 'docker' is blocked: This command is blocked by default due to destructive potential. Use --allow-command to override if you understand the risks.
nono: Command 'docker' is blocked: This command is blocked by default due to destructive potential. Use --allow-command to override if you understand the risks
```

### DIFF
--- a/crates/nono-cli/src/policy.rs
+++ b/crates/nono-cli/src/policy.rs
@@ -559,10 +559,7 @@ pub(crate) fn add_deny_access_rules(
             // Use (path ...) not (subpath ...) — socket connections match on the exact
             // path, not a prefix. Both symlink and canonical paths are covered because
             // emit_deny_rules is called for each form by the caller.
-            caps.add_platform_rule(format!(
-                "(deny network-outbound (path \"{}\"))",
-                escaped
-            ))?;
+            caps.add_platform_rule(format!("(deny network-outbound (path \"{}\"))", escaped))?;
             Ok(())
         };
 
@@ -1523,7 +1520,9 @@ mod tests {
         let expected = real_dir_canonical.join("future.sock");
 
         // Only returns Some when the resolved path differs from the original
-        if link_dir.canonicalize().ok().as_deref() != Some(&*real_dir_canonical) || link_dir != real_dir {
+        if link_dir.canonicalize().ok().as_deref() != Some(&*real_dir_canonical)
+            || link_dir != real_dir
+        {
             assert_eq!(result, Some(expected));
         }
     }
@@ -1542,7 +1541,11 @@ mod tests {
         if canonical == file {
             assert_eq!(result, None, "no parent symlinks means None");
         } else {
-            assert_eq!(result, Some(canonical), "parent symlinks means Some(canonical)");
+            assert_eq!(
+                result,
+                Some(canonical),
+                "parent symlinks means Some(canonical)"
+            );
         }
     }
 
@@ -1560,7 +1563,10 @@ mod tests {
         std::os::unix::fs::symlink(&real_dir, &link_dir).expect("create symlink");
 
         let socket_path = link_dir.join("daemon.sock");
-        assert!(!socket_path.exists(), "test precondition: socket must not exist");
+        assert!(
+            !socket_path.exists(),
+            "test precondition: socket must not exist"
+        );
 
         let mut caps = CapabilitySet::new();
         let mut deny_paths = Vec::new();
@@ -1571,7 +1577,8 @@ mod tests {
         let real_dir_canonical = real_dir.canonicalize().expect("canonicalize real_dir");
         let resolved_socket = real_dir_canonical.join("daemon.sock");
 
-        let parent_is_symlinked = link_dir.canonicalize().ok().as_deref() != Some(&*real_dir_canonical)
+        let parent_is_symlinked = link_dir.canonicalize().ok().as_deref()
+            != Some(&*real_dir_canonical)
             || link_dir != real_dir;
 
         if parent_is_symlinked && resolved_socket != socket_path {


crates/nono-cli/src/policy.rs
```diff
@@ -474,6 +474,20 @@ pub(crate) fn add_deny_access_rules(
...
+            // evaluated at syscall time, not at sandbox_init time, so it covers sockets
+            // that do not yet exist. For non-socket paths the rule is a harmless no-op.
+            // Use (path ...) not (subpath ...) — socket connections match on the exact
+            // path, not a prefix. Both symlink and canonical paths are covered because
+            // emit_deny_rules is called for each form by the caller.
+            caps.add_platform_rule(format!(
+                "(deny network-outbound (path \"{}\"))",
+                escaped
+            ))?;
             Ok(())
```
lukehinds: This new rule is correct for the path it receives, but the surrounding logic still only emits the canonical `/private/...` form when full-path `canonicalize()` succeeds. For not-yet-created sockets under symlinked parents like `/var/run/...,` that fails, so the claim that this covers future sockets is not yet there fully. We would need to resolve parent symlinks even when the leaf socket does not exist.

I think a helper function should do it, codex recommended this:

```
fn resolve_parent_symlinks(path: &Path) -> Result<Option<PathBuf>> {
    let mut suffix = Vec::new();
    let mut cur = path;

    while !cur.exists() {
        let name = cur.file_name().ok_or_else(|| {
            NonoError::ConfigParse(format!("cannot resolve parent for {}", path.display()))
        })?;
        suffix.push(name.to_os_string());
        cur = cur.parent().ok_or_else(|| {
            NonoError::ConfigParse(format!("cannot resolve parent for {}", path.display()))
        })?;
    }

    let mut resolved = cur.canonicalize()?;
    for part in suffix.iter().rev() {
        resolved.push(part);
    }

    Ok((resolved != path).then_some(resolved))
}
```

Then in crates/nono-cli/src/policy.rs, line 445, emit rules for both:

  - the original path
  - the resolved-parent path, even if the leaf socket does not exist yet

for example

  - requested deny path: `/var/run/future.sock`
  - `future.sock` does not exist yet
  - but `/var/run` does exist
  - on macOS, `/var/run` resolves to `/private/var/run`

  So you can derive:

  - original rule: `/var/run/future.sock`
  - resolved rule: `/private/var/run/future.sock`

Hope that makes sense, happy to chat it through on discord or push on top of your branch.

Austio: Great catch, will give it a go

Austio: https://github.com/always-further/nono/pull/488/changes/0b209f30fc410d359274da70f18532dc45f6affe - build locally and verified the root functionality

PR #516, opened 2026-03-27
feat(proxy): auto-detect credential format from inject_header

RobertWi:
Addresses discussion in #424

## Summary

Auto-detects the credential format based on the `inject_header` value, so `credential_format` doesn't need to be set explicitly.

- `Authorization` header → `Bearer {}`
- All other headers (e.g. `PRIVATE-TOKEN`, `X-API-Key`) → `{}`

Explicit `credential_format` in the profile still overrides the default.

## Production context

Running for a week in a Kubernetes cluster with 9 sandboxed agents. Multiple credential types (raw token headers, Bearer auth) all auto-detected correctly.

## Test plan

- [x] `Authorization` header defaults to `Bearer {}`
- [x] Other headers default to `{}`
- [x] Explicit `credential_format` overrides the default
- [x] All existing tests pass

### DIFF
--- a/crates/nono-proxy/src/credential.rs
+++ b/crates/nono-proxy/src/credential.rs
@@ -100,11 +100,21 @@ impl CredentialStore {
                     Err(e) => return Err(ProxyError::Credential(e.to_string())),
                 };
 
-                // Format header value based on mode
+                // Format header value based on mode.
+                // When inject_header is not "Authorization" (e.g., "PRIVATE-TOKEN",
+                // "X-API-Key"), the credential is injected as-is unless the user
+                // explicitly set a custom format. The default "Bearer {}" only
+                // makes sense for the Authorization header.
+                let effective_format = if route.inject_header != "Authorization"
+                    && route.credential_format == "Bearer {}"
+                {
+                    "{}".to_string()
+                } else {
+                    route.credential_format.clone()
+                };
+
                 let header_value = match route.inject_mode {
-                    InjectMode::Header => {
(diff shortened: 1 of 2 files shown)


lukehinds (commented):
Thanks @RobertWi, I think there’s a mismatch with the current reverse-proxy pipeline here that I didn’t notice on first review.

I can see your PR makes it easier for custom user profiles to use headers beyond the built-in ones in network-policy.json, for example PRIVATE-TOKEN / MY-TOKEN, which makes sense and is
useful. But in crates/nono-proxy/src/reverse.rs, validation and injection are route-aware while header stripping is still hardcoded:

- Validation uses cred.header_name in `crates/nono-proxy/src/reverse.rs` (line 574) and `crates/nono-proxy/src/reverse.rs` (around line 329)
- Stripping removes only `Authorization, x-api-key, and x-goog-api-key` in `crates/nono-proxy/src/reverse.rs` (line 372)
- Injection uses `cred.header_name` in `crates/nono-proxy/src/reverse.rs` (line 835)

So for a custom header enabled by this PR, the proxy can validate the phantom token but fail to strip it before adding the real credential, which means the upstream request could contain both, e.g.

```
PRIVATE-TOKEN: phantom123
PRIVATE-TOKEN: real-secret
```

Built-in headers mostly keep working because they happen to be in the strip list, but new/custom headers can be forwarded incorrectly by the current request-rewrite path.

I think the fix is to make header stripping route-aware and remove the configured cred.header_name instead of relying on a hardcoded set.

PR #517, opened 2026-03-27
feat(proxy): oauth2 client-credentials token exchange with cache

RobertWi:
Addresses discussion point 3 in #424

## Summary

Adds OAuth2 `client_credentials` token exchange to the credential proxy. The agent never sees client secrets or access tokens — it makes plain HTTP requests and the proxy handles authentication transparently.

Profile config (mutually exclusive with `credential_key`):
```json
{
  "custom_credentials": {
    "mcp": {
      "upstream": "https://mcp.example.com",
      "auth": {
        "token_url": "https://keycloak.example.com/realms/app/protocol/openid-connect/token",
        "client_id": "file:///vault/secrets/mcp-client-id",
        "client_secret": "file:///vault/secrets/mcp-client-secret",
        "scope": "openid"
      },
      "inject_header": "Authorization"
    }
  }
}
```

- Token exchange at startup, auto-refresh when within 30s of expiry
- `client_id` and `client_secret` support `file://` and `env://` URI schemes
- Token cache with TTL — no redundant exchanges
- Composes with `file://` credential source (PR #515)

## Production context

Running for a week in a Kubernetes cluster with Landlock V4 enforcement. Used for Keycloak OIDC authentication to internal MCP services. Agents authenticate transparently without seeing any credentials.

## Security considerations

- Client secrets are read at startup before the sandbox activates
- Access tokens are held in proxy memory only, never exposed to the child process
- Token refresh happens inside the proxy — the agent has no visibility into the OAuth2 flow
- Mutual exclusion with `credential_key` prevents ambiguous credential config

## Test plan

- [x] OAuth2 config deserializes from profile JSON
- [x] `client_id` / `client_secret` support `file://` and `env://`
- [x] Mutual exclusion with `credential_key` validated
- [x] Token cache respects TTL
- [x] All existing tests pass

### DIFF
--- a/crates/nono-cli/data/nono-profile.schema.json
+++ b/crates/nono-cli/data/nono-profile.schema.json
@@ -381,9 +381,9 @@
     },
     "CustomCredentialDef": {
       "type": "object",
-      "description": "Custom credential route definition for the reverse proxy. Defines how to route requests and inject credentials for a service.",
+      "description": "Custom credential route definition for the reverse proxy. Defines how to route requests and inject credentials for a service. Must have either credential_key or auth (mutually exclusive).",
       "additionalProperties": false,
-      "required": ["upstream", "credential_key"],
+      "required": ["upstream"],
       "properties": {
         "upstream": {
           "type": "string",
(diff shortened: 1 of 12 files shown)


iNecas:
This fixes https://github.com/always-further/nono/issues/466, right? Adding just to cross-link to the issue for better discoverability.

lukehinds:
apologies @iNecas , for late reply - it does 

@RobertWi I can get this rebased if you're short on availability

PR #639, opened 2026-04-11
feat(gpu): add Linux and WSL2 GPU access via --allow-gpu

### LINKED ISSUE #405: macOS: support opt-in non-filesystem Seatbelt capabilities (IOKit, Apple Events, WebKit, GPU)
## Summary

nono's macOS Seatbelt sandbox currently only exposes filesystem and network controls to profiles. Several workloads require **non-filesystem kernel capabilities** that cannot be unblocked with `--allow` or `--read` flags. There is no profile-level mechanism to opt into these.

## Related issues

- #131 — OpenClaw gateway browser control blocked (`network-bind`, subprocess exec, Unix domain sockets)
- #347 — Metal GPU access blocked (`iokit-open`, `iokit-get-properties`)
- #371 — WKWebView fails (`IORegisterForSystemPower`, `launchservicesd`, fonts, `webkit.mach-bootstrap`)
- #122 — Chromium/Playwright investigation (`iokit-open`, `mach-register`)

## Additional use case: Apple Events

Sending iMessages via `imsg` requires `(allow apple-event-send (target-bundle-id "com.apple.iChat"))`. This is another non-filesystem capability that profiles cannot currently express.

## Proposal

The internal plumbing already exists — `CapabilitySet::add_platform_rule()` accepts arbitrary S-expressions and validates them. The gap is exposing this to users. Three approaches, not mutually exclusive:

### Option A: Named policy groups

Add predefined groups to `policy.json` (e.g. `allow-gpu`, `allow-webkit`, `allow-apple-events-ichat`) that profiles opt into via `security.groups`:

```json
{
  "security": { "groups": ["allow-gpu"] }
}
```

**Pros:** Auditable, consistent with existing group system, each group can be carefully scoped and reviewed.
**Cons:** Every new capability requires a policy change — doesn't help users with bespoke needs (e.g. Apple Events to a specific bundle ID).

### Option B: Structured capability enum

A curated `security.macos.capabilities` field with known capability names, following lukehinds' suggestion in #347:

```json
{
  "security": {
    "macos": {
      "capabilities": ["gpu", "apple-events", "webkit"]
    }
  }
}
```

**Pros:** Discoverable, self-documenting, can be validated against a known set. Pairs naturally with CLI flags (`--allow-gpu`).
**Cons:** Still requires adding each new capability to the enum. Doesn't cover granular parameterization (e.g. Apple Events to a *specific* bundle ID).

### Option C: Raw platform rules (power-user escape hatch)

Expose `platform_rules` in profiles for arbitrary Seatbelt S-expressions:

```json
{
  "policy": {
    "platform_rules": [
      "(allow apple-event-send (target-bundle-id \"com.apple.iChat\"))"
    ]
  }
}
```

**Pros:** Maximum flexibility, covers all current and future use cases without code changes. Existing `validate_platform_rule()` already blocks root-level file access grants and validates syntax.
**Cons:** Hardest to audit, runs against nono's "no escape hatch" philosophy, users need to know Seatbelt DSL.

### Recommended approach

Options A or B as the primary interface (covering common cases like GPU, WebKit, Apple Events), with Option C gated behind an additional opt-in (e.g. `security.allow_raw_platform_rules: true`) for power users with bespoke needs. All should follow the `--allow-launch-services` precedent: profile must permit it, CLI flag activates it, warning is printed.

## Considerations

- **CLI flags:** `--allow-iokit` / `--allow-gpu` as shortcuts for the most common case (PR #122 proposes this)
- **Warning output:** print a clear warning when non-filesystem capabilities are active, matching `--allow-launch-services` pattern
- **Base profile gaps:** `mach-register` may belong in the base Seatbelt profile alongside the existing `mach-lookup` allow (PR #122 finding)
- **Linux:** these are macOS-only; fields should be silently ignored on Linux

scp7:
## Summary

Implements GPU device access for Linux and WSL2 sandboxes via `--allow-gpu`.

### Linux GPU support

- Grant read/write access to DRM render nodes (`/dev/dri/renderD*`) for GPU compute
- Grant read/write access to NVIDIA compute devices (`/dev/nvidia*`, `nvidiactl`, `nvidia-uvm`, `/dev/nvidia-caps`)
- Grant read/write access to AMD KFD device (`/dev/kfd`) for ROCm/HIP
- Grant read-only access to Vulkan/Mesa ICD paths and `/sys/class/drm`
- Deliberately exclude display-control devices (`/dev/dri/card*`, `nvidia-modeset`)

### WSL2 GPU support

WSL2 exposes GPU access through `/dev/dxg` (paravirtualized DirectX) rather than standard Linux device nodes:

- Detect `/dev/dxg` as a GPU device (read+write access)
- Grant read access to `/usr/lib/wsl/lib/` (CUDA/D3D12 shared libraries mounted by WSL2 init)
- Without this, `--allow-gpu` always fails on WSL2 since no standard device nodes exist

### Tests

- Integration tests for Linux GPU device access (`/dev/dri`, `/dev/nvidia`, `/dev/kfd`)
- WSL2 GPU tests: deny without flag, allow with flag, `/usr/lib/wsl/lib` readable, `nvidia-smi` inside sandbox
- End-to-end ollama inference tests verifying GPU compute works within the sandbox
- Unit tests for headless Linux setups without `/dev/dri`
- Profile merging tests for `allow_gpu` field

### Docs

- Updated WSL2 feature matrix: `--allow-gpu` listed as Full (118 features total)
- Updated WSL2 compatibility table with DirectX passthrough note
- Updated `--allow-gpu` CLI help and profile schema for cross-platform support

### Key files

- [sandbox_prepare.rs](https://github.com/scp7/nono/blob/issue-405/crates/nono-cli/src/sandbox_prepare.rs) — GPU device detection and capability grants
- [test_gpu.sh](https://github.com/scp7/nono/blob/issue-405/tests/integration/test_gpu.sh) — GPU integration tests
- [test_gpu_ollama.sh](https://github.com/scp7/nono/blob/issue-405/tests/integration/test_gpu_ollama.sh) — Ollama inference tests

## Test plan

- [x] `nvidia-smi` works inside sandbox on WSL2 (RTX 3080, CUDA 12.6)
- [x] Ollama tinyllama inference succeeds with GPU inside sandbox (5/5)
- [x] GPU test suite passes on WSL2 (7/7, 3 skipped for missing native devices)
- [x] `/dev/dxg` correctly denied without `--allow-gpu`
- [x] WSL2 integration tests unaffected (21/21)
- [x] Clippy, rustfmt, cargo test clean

Tested on:
- Ubuntu 24.04 WSL2, rolling kernel 6.19.11, Landlock V6
- NVIDIA GeForce RTX 3080, Driver 560.94, CUDA 12.6

Closes #405


lukehinds (approved):

agnosticlines:
@scp7 the tests seem to use the old rules before the tightening passes

scp7:
Thanks @agnosticlines, now in #645

Kexin-xu-01:
## Follow-up: issues found testing on NVIDIA driver 570 / CUDA 12.8 (A100)

Tested this PR on a native Linux system (not WSL2) with driver 570, CUDA 12.8, Landlock V4, and found two issues.

---

### 1. `nvidia-uvm-tools` not included in GPU device allowlist

The filter in `maybe_enable_gpu()` matches `nvidia-uvm` explicitly but not `nvidia-uvm-tools`:

```rust
n == "nvidiactl"
    || n == "nvidia-uvm"
    || (n.starts_with("nvidia")
        && n[6..].bytes().all(|b| b.is_ascii_digit())
        && n.len() > 6)
```

`nvidia-uvm-tools` has a non-digit suffix so it falls through all three branches and is never granted.

NVIDIA driver 570 opens `/dev/nvidia-uvm-tools` (minor 1, alongside `nvidia-uvm` minor 0) during UVM initialization. When `open()` returns `EACCES` from Landlock, the driver treats it as a fatal `cudaErrorOperatingSystem` (Error 304) — even when `/dev/nvidia-uvm` itself is accessible. The device exists on these systems as `crw-rw-rw- 1 root root 235, 1 /dev/nvidia-uvm-tools`.

**Fix**: add an explicit match, mirroring how `nvidia-uvm` is listed:

```rust
n == "nvidiactl"
    || n == "nvidia-uvm"
    || n == "nvidia-uvm-tools"   // driver 570+ opens this during UVM init
    || (n.starts_with("nvidia")
        && n[6..].bytes().all(|b| b.is_ascii_digit())
        && n.len() > 6)
```

---

### 2. CUDA Error 304 from `/proc/self/task/<tid>/comm` write being blocked

NVIDIA driver 570 names its internal threads by writing to `/proc/self/task/<tid>/comm` during driver init. If the sandbox grants `/proc` as read-only, this write returns `EACCES`, which the driver treats as fatal Error 304.

This manifests even after Fix 1 above: `torch.cuda.device_count()` returns 1 (device enumeration works) but `torch.cuda.get_device_capability()` / `cuInit(0)` fails with Error 304. Strace inside the sandbox confirms the only `EACCES` hit is the `write` to `/proc/self/task/<tid>/comm`.

Older drivers (tested on 560) do not exhibit this — they ignore the failed write. Driver 570 made it fatal.

**Fix**: profiles using `/proc` for CUDA workloads need read+write access, not read-only. For the `claude-code-autoresearch` profile specifically, `/proc` should be in `allow` (r+w) rather than `read`.

This may also be worth noting in the `--allow-gpu` documentation, since it's a non-obvious requirement that only surfaces on newer driver versions.

PR #654, opened 2026-04-13
fix(sandbox): redirect ~/.claude.json to ~/.claude/ via symlink on all unix platforms

### LINKED ISSUE #220: fix: claude-code profile is unusable on Linux due to config lock files in ~/
## Summary

The `claude-code` profile cannot be used for real Claude Code sessions on Linux. Claude Code crashes with `EACCES` when saving configuration because it creates lock and temp files in `~/` next to `~/.claude.json`.

## Root cause

Claude Code's config save mechanism (`saveConfigWithLock`) does the following:

1. Creates a lock file at `~/.claude.json.lock`
2. Creates a backup in `~/.claude/backups/`
3. Writes to a temp file `~/.claude.json.tmp.<pid>.<timestamp>`
4. Renames the temp file to `~/.claude.json`

Steps 1 and 3 require creating **new files** in `~/`. The profile uses `allow_file` for `~/.claude.json`, which creates a Landlock `PathBeneath` rule for that specific file. Creating new files in a directory requires `LANDLOCK_ACCESS_FS_MAKE_REG` on the parent directory (`~/`), which is not granted and should not be — granting write to `~/` would defeat the sandbox.

Landlock is strictly allow-list and cannot express "allow creating files in `~/` but deny write to subdirectories."

## Impact

Any Claude Code session longer than a few minutes will crash when it first tries to persist config changes. `claude -v` and very short `-p` prompts may work because they exit before a config save is triggered.

## Upstream issue

Filed as anthropics/claude-code#30372. The fix needs to happen on the Claude Code side — lock and temp files should be created in `~/.claude/` (already writable in the profile) rather than next to `~/.claude.json` in `~/`. There is also a broader XDG compliance request (anthropics/claude-code#1455) that would resolve this naturally.

## Workaround

Users can grant write access to their home directory, but this defeats the purpose of sandboxing:

```bash
nono run --profile claude-code --allow-cwd --allow ~/ -- claude
```

## Environment

- nono v0.8.0
- Claude Code 2.1.63
- Linux 6.19.5, Landlock ABI v5

advaithsujith:
## Problem

Fixes #220

Claude Code writes `~/.claude.json` atomically by creating a temp file at `~/.claude.json.tmp.<pid>.<timestamp>` in the same directory (`~/`), then renaming it. Landlock (Linux) and Seatbelt (macOS) cannot grant permission for these dynamically-named files in `~/`, so token writes silently fail and users are logged out mid-session.

This is the root cause of the workaround reported in #220 where users had to run `claude /login` outside the sandbox, then restart inside — the login itself worked but the token write failed silently, so the next session had no credentials.

## Fix

Before applying the sandbox, redirect `~/.claude.json` to `~/.claude/claude.json` via a symlink. Claude Code resolves symlinks before computing the temp file path, so temp files land in `~/.claude/` (already `readwrite` in the profile) instead of `~/` (not writable inside the sandbox).

Three cases handled:
- `~/.claude.json` is a regular file → move to `~/.claude/claude.json`, create symlink
- `~/.claude.json` doesn't exist yet → pre-create target, create symlink
- `~/.claude.json` is already a symlink → no-op

Also expands the precreate block from `#[cfg(target_os = "linux")]` to `#[cfg(unix)]` — the same atomic-write problem exists on macOS (confirmed by multiple users in #220), so the fix needs to run on both platforms.

## Testing

Tested on macOS with Seatbelt:

- Confirmed `~/` temp writes are blocked inside sandbox (`EPERM`)
- Confirmed `~/.claude/` temp writes succeed after symlink redirect
- Ran full `claude /login` flow inside the nono sandbox — browser opened, OAuth completed, token written successfully with no workaround needed
- Confirmed authenticated API calls work inside the sandbox after login (`claude --print` returns a real response)
- All 20 integration test suites pass

### DIFF
--- a/crates/nono-cli/src/sandbox_prepare.rs
+++ b/crates/nono-cli/src/sandbox_prepare.rs
@@ -1,7 +1,7 @@
 use crate::capability_ext::{self, CapabilitySetExt};
 use crate::cli::SandboxArgs;
 use crate::command_blocking_deprecation;
-#[cfg(target_os = "linux")]
+#[cfg(unix)]
 use crate::config;
 use crate::credential_runtime::load_env_credentials;
 use crate::profile;
@@ -12,7 +12,7 @@ use crate::{DETACHED_CWD_PROMPT_RESPONSE_ENV, DETACHED_LAUNCH_ENV};
 use colored::Colorize;
 use nono::{AccessMode, CapabilitySet, FsCapability, NonoError, Result, Sandbox};
 use std::collections::HashMap;
-#[cfg(target_os = "linux")]
+#[cfg(unix)]
 use std::os::unix::fs::OpenOptionsExt;
 use std::path::{Path, PathBuf};
 use tracing::{info, warn};
@@ -636,7 +636,7 @@ pub(crate) fn prepare_sandbox(args: &SandboxArgs, silent: bool) -> Result<Prepar
         command_blocking_deprecation::print_warnings(&profile_warnings, silent);
     }
 
-    #[cfg(target_os = "linux")]
+    #[cfg(unix)]
     if args.profile.as_deref() == Some("claude-code") {
         let home = config::validated_home()?;
         let home_path = std::path::Path::new(&home);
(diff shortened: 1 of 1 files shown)


ctrlok:
@advaithsujith PR broke CLAUDE_CONFIG_DIR env variable usage to override default config path.

PR #680, opened 2026-04-15
fix(sandbox): preserve symlink path when adding CWD capability on macOS

### LINKED ISSUE #662: opencode (profile) does not work with symlinked directories as cwd
### What happened?

I typically use symlinks as shortcuts to my projects, e.g.

```
mstenber@hana ~/homenetflow>ls -ld ~/homenetflow
lrwxr-xr-x 1 mstenber staff 20 Apr  3 09:35 /Users/mstenber/homenetflow -> projects/homenetflow/
```

With nono + opencode, it doesn't quite work:
```
mstenber@hana ~/homenetflow>nono run --profile opencode --allow-cwd -- opencode

  nono v0.33.0
2026-04-13T15:48:25.974729Z  WARN Profile path '$HOME/.opencode' does not exist, skipping: /Users/mstenber/.opencode
  Capabilities:
   r+w  /Users/mstenber/.config/opencode (dir)
   r+w  /Users/mstenber/.cache/opencode (dir)
   r+w  /Users/mstenber/.local/share/opencode (dir)
   r+w  /Users/mstenber/.local/share/opentui (dir)
   r+w  /Users/mstenber/.local/state/opencode (dir)
   r+w  /private/tmp (dir)
   r+w  /Users/mstenber/share/1/projects/homenetflow (dir)
       + 37 system/group paths (-v to show)
   net  outbound allowed

  mode supervised (supervisor)
  Applying sandbox...
Error: Unexpected error, check log file at /Users/mstenber/.local/share/opencode/log/2026-04-13T154827.log for more details

EPERM: operation not permitted, lstat '/Users/mstenber/homenetflow'

NONO DIAGNOSTIC
────────────────────────
The command failed. This may be due to sandbox restrictions. (exit code 1)

```

However, if I move to realpath, it is fine:
```
mstenber@hana ~/homenetflow>cd $(realpath)
mstenber@hana ~/share/1/projects/homenetflow>nono run --profile opencode --allow-cwd -- opencode

  nono v0.33.0
2026-04-13T15:49:02.270058Z  WARN Profile path '$HOME/.opencode' does not exist, skipping: /Users/mstenber/.opencode
  Capabilities:
   r+w  /Users/mstenber/.config/opencode (dir)
   r+w  /Users/mstenber/.cache/opencode (dir)
   r+w  /Users/mstenber/.local/share/opencode (dir)
   r+w  /Users/mstenber/.local/share/opentui (dir)
   r+w  /Users/mstenber/.local/state/opencode (dir)
   r+w  /private/tmp (dir)
   r+w  /Users/mstenber/share/1/projects/homenetflow (dir)
       + 37 system/group paths (-v to show)
   net  outbound allowed

  mode supervised (supervisor)
  Applying sandbox...
.. and it works ..
```

Note: codex seems to work fine in the similar symlink case (it just shows the realpath as working directory), so perhaps it is opencode misfeature?

opencode to work.

### Steps to reproduce

See above - start opencode profile in symlinked directory.

### Relevant output or logs

```shell

```

### nono version

0.33.0

### Operating system

macOS

### OS version / distro

macOS 26.4

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

advaithsujith:
## Summary

When a user \`cd\`s into a symlinked directory (e.g. \`~/project -> /real/path/project\`) and runs \`nono run\` **without** \`--workdir\`, the sandboxed process gets \`EPERM\` on \`lstat\` of the symlink path. macOS Seatbelt uses literal path matching, so both the symlink and the real path need explicit rules.

### Root cause

\`resolved_workdir()\` was calling \`std::env::current_dir()\` (which calls \`getcwd()\`) to determine the working directory when \`--workdir\` was not supplied. \`getcwd()\` always returns the **canonical real path** — it does not preserve the symlink path the user typed. So the symlink path was never captured and no Seatbelt rule was ever emitted for it.

### Fix

\`resolved_workdir()\` now prefers \`\$PWD\` over \`current_dir()\` when \`--workdir\` is not explicitly set. The shell sets \`\$PWD\` to the symlink path when the user \`cd\`s — this is the right path to use. As a safety guard, \`\$PWD\` is validated by canonicalising it and checking it matches \`getcwd()\`; a stale or spoofed \`\$PWD\` silently falls back to the real path.

The existing macOS-only block that emits a Seatbelt rule for \`workdir\` whenever \`workdir != canonical\` then covers the symlink path for both the \`--allow-cwd\` case and the profile-already-covers-canonical-path case.

Closes #662

## Test plan

- [x] Created a symlink to a directory (\`ln -s /tmp/real ~/link\`), \`cd ~/link\`, ran \`nono run -- ls\` **without** \`--workdir\` — confirmed no \`EPERM\` (reproduces #662 exactly)
- [x] Confirmed running from the real path still works
- [x] Confirmed \`--workdir <symlink-path>\` still works
- [x] Unit tests: \`resolved_workdir_prefers_pwd_symlink_over_getcwd_when_valid\` and \`resolved_workdir_ignores_stale_pwd\` pass
- [x] Integration tests: all 20 suites pass (\`bash tests/run_integration_tests.sh\`)


lukehinds:
Looks like the fix only works when the user explicitly passes `--workdir <symlink-path>` - which is referenced in the issue. The test plan ("created a symlink, cd into it…") doesn't say whether `--workdir `was  passed; be good to clarify, and ideally reproduce #662 exactly (no `--workdir`, just cd into a symlink dir).

If the goal is to fix the no `--workdir case`, the workdir source likely needs to consult $PWD (validating it resolves to the same inode as getcwd()) rather than relying on current_dir().

advaithsujith:
Thanks for the feedback, Luke. You were right — the original fix only worked when `--workdir` was explicitly passed because `resolved_workdir()` was calling `current_dir()` (`getcwd()`) which always returns the canonical real path, discarding the symlink.

Fixed in the latest commit by reading `$PWD` instead. The shell sets `$PWD` to the symlink path on `cd`, which is exactly what we need. I added a safety guard: `$PWD` is canonicalized and compared to `getcwd()` before being trusted, so stale or spoofed values fall back to the real path.

Tested without `--workdir` by reproducing #662 directly (`ln -s /tmp/real ~/link`, `cd ~/link`, `nono run -- ls`) — no longer produces `EPERM`.

lukehinds:
@advaithsujith just the merge conflict to resolve and then we can merge this.

lukehinds (approved):

PR #684, opened 2026-04-15
fix(cli): issue with claude code profiles and CLAUDE_CONFIG_DIR variable

ctrlok:
Hey! 
v0.34 broke profile extentions and CLAUDE_CONFIG_DIR support for nono.

As far as I can see, the issue is that it was already broken on linux and #654 brokes it on macOS. 

The issue is here https://github.com/always-further/nono/blob/main/crates/nono-cli/src/sandbox_prepare.rs#L1015 

```rust
if args.profile.as_deref() == Some("claude-code") {
```

The profile name shouldn't be hardcoded, while it breaks the whole flow for all extended profiles

Also, symlink may be not a correct way to do all this authentication dance, while some users (like me) can use multiple claude profiles and rely on CLAUDE_CONFIG_DIR variable.

**Proposal:** 

If CLAUDE_CONFIG_DIR variable is set, it automatically uses files from inside CLAUDE_CONFIG_DIR directory. So this variable can be set for all profiles (I guess it will be not used in different tools, still maybe there is a better way to do it)

So we can set this variable and it will force claude-code to create claude.json and locks inside CLAUDE_CONFIG_DIR (which will be ~/.claude by default) and also it will respect CLAUDE_CONFIG_DIR

### DIFF
--- a/crates/nono-cli/src/sandbox_prepare.rs
+++ b/crates/nono-cli/src/sandbox_prepare.rs
@@ -1011,6 +1011,14 @@ pub(crate) fn prepare_sandbox(args: &SandboxArgs, silent: bool) -> Result<Prepar
         command_blocking_deprecation::print_warnings(&profile_warnings, silent);
     }
 
+    // Ensure CLAUDE_CONFIG_DIR is set so Claude Code writes config
+    #[cfg(unix)]
+    if std::env::var_os("CLAUDE_CONFIG_DIR").is_none() {
+        let home = config::validated_home()?;
+        #[allow(clippy::disallowed_methods)] // Single-threaded before fork.
+        std::env::set_var("CLAUDE_CONFIG_DIR", Path::new(&home).join(".claude"));
+    }
+
     #[cfg(unix)]
     if args.profile.as_deref() == Some("claude-code") {
         let home = config::validated_home()?;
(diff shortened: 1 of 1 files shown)


lukehinds:
Nice ! good find @ctrlok and welcome to the community

It makes sense to get this fix in , but you maybe interested to know we are going to pull the profiles out of the core code and allow users to customize and host their own with full supply chain security in place. This will mean if you have any custom hooks, CLAUDE.mds you need, you can change control them in your own repo, have a nono registry use your repo as a trusted publisher and then do something like `nono pull ctrlok/ny-claude-stuff` 

To clear the DCO bot failure:

In your local branch, run: `git rebase HEAD~1 --signoff`
Force push your changes to overwrite the branch: `git push --force-with-lease origin fix-claude-profile-variable`

from then on, just add a `-s` to your commits. 

The cargo audit is not your issue, its from rustls-webpki which I can look after

ctrlok:
@lukehinds hey! thanks for the answer. 

Sorry for multiple attempts to add 'signedoff by' trailer - I'm using jujutsu instead of git and it requires some template magic. 

About moving out from a profiles - it is a great idea. I'm implementing developers AI policy for a company I work for, and we are using custom profiles anyway. 

As for PR - I think it will be useful to have this variable setted by default to simplify future integrations. Or to have something about it in documentation. Anyway, with current change everything works fine on current version of nono.

ctrlok:
btw, you saying that default profiles will be removed makes me think, and I am starting to think that maybe documentation update will be better than a code changes

I created not-extended profile 

```json
{
  "meta": {
    "name": "claude-naked",
    "version": "1.0.0",
    "description": "Standalone Claude Code profile with second config dir"
  },
  "security": {
    "groups": [
      "system_read_macos",
      "system_write_macos",
      "user_tools",
      "homebrew_macos",
      "user_caches_macos",
      "node_runtime",
      "rust_runtime",
      "python_runtime",
      "vscode_macos",
      "git_config"
    ]
  },
  "network": {
    "block": false
  }
}
```

then I created a folder `$CLAUDE_CONFIG_DIR.lock` (it is used only for lock till token is refreshed)

and then I ran

```shell
nono run --profile claude-naked --allow $CLAUDE_CONFIG_DIR --allow-cwd -v --allow $CLAUDE_CONFIG_DIR.lock -- claude
```

It allows me to authenticate and refresh tokens without any issues.

And all those stuff that currently happens in claude-code profile is obsolete.

Still, it requires user to create a directories `$CLAUDE_CONFIG_DIR` and `$CLAUDE_CONFIG_DIR.lock` if they are using variables and `~/.claude` and `~/.claude.lock` if they are not explicitly using variables.

What will be better way to format this PR? I can update documentation, but also I think group `claude_code_macos` can be updated to remove access to `/Users/sevapoliakov/Library/Keychains/` (from what I see it is not needed) and if group is setted then nono can create directories `$CLAUDE_CONFIG_DIR` and `$CLAUDE_CONFIG_DIR.lock` if they do not exist and set variable `$CLAUDE_CONFIG_DIR = ~/.claude` if it is not set.

wdyt @lukehinds

lukehinds:
so this might be the root of the login issue which is paining us as well? 

yep, inbuilt will be removed, but I think we need to get your fix in asap , especially if it solves the login issues , as that is a frustrating experience for users.  The cargo failure is fixed in main now, if you could do a rebase, lets get this in and then we can think about doc changes. 

There were just a few bits I noted on review

The new `CLAUDE_CONFIG_DIR` block sits outside the `if args.profile.as_deref() == Some("claude-code")` guard, so it might fire for every profile, not just Claude Code ones. We could scope it so it only kicks in for Claude-related profiles. 

Also the hardcoded `Some("claude-code")` check is still there on line 1015, which was the original problem. Extended profiles still won't enter the block that precreates .cache/claude-cli-nodejs. We need to widen that check to cover profiles that extend claude-code too.

Is the removal of the the precreate for `.claude.json.lock` intentional. Can you confirm that's no longer needed with the CLAUDE_CONFIG_DIR approach? I want to make sure we're not breaking lock file creation for token refreshes.

lukehinds:
@ctrlok example of what you will be able to do https://github.com/always-further/nono-packs/tree/main/claude

you can put anything you need in there for a custom profile and then: `nono pull ctrlok/claude` and its in place as your main claude profile. We will have a registry that checks provenance to the source repo is present and publishes to the repo - from there on, tag a new release in your repo, and it auto-publishes:

https://github.com/always-further/nono/issues/630

ctrlok:
Continue working in #691

Issue #688, opened 2026-04-15
Filter environment variables


rnestler:
### What problem are you trying to solve?

Sometimes I have credentials such as API tokens set in my environment, which I wouldn't want to expose to an LLM agent to not leak it accidentally.

Currently the child processes of `nono run` seem to inherit the full environment of the parent.

### What would you like to see?

Similar to `sudo` which has an `env_keep` directive to limit the environment variables that the child receives, I'd like to whitelist in profiles what variables the child inherits.

### What have you tried instead?

Manually unsetting credentials, but this is a bit error prone.

### How is this blocking you?

None

synapticarbors:
I have a wrapper bash script that I use to run nono that tries to create a minimal clean env:

```bash
#!/usr/bin/env bash
set -euo pipefail

# Allowlisted prefixes/names
ALLOWED_VARS=(
    PATH
    HOME
    USER
    SHELL
    LOGNAME
    TERM
    TERMINFO
    TERM_PROGRAM
    TERM_PROGRAM_VERSION
    COLORTERM
    LANG
    TMPDIR
    PAGER
    LESS
    XDG*
    # __CF_USER_TEXT_ENCODING
    # XPC_FLAGS
    # XPC_SERVICE_NAME
)

build_clean_env() {
    clean_env=()
    for var in $(compgen -e); do
        for pattern in "${ALLOWED_VARS[@]}"; do
            if [[ "$var" == $pattern ]]; then
                clean_env+=("$var=${!var}")
                break
            fi
        done
    done
}

# --- Main ---

NONO_PROFILE="claude-code"

usage() {
    cat <<EOF
Usage:
  $(basename "$0") claude [args...]       Launch Claude Code in sandbox
  $(basename "$0") nono <command> [args...] Run any nono command with clean env
EOF
    exit 1
}

[[ $# -lt 1 ]] && usage

MODE="$1"
shift

build_clean_env

case "$MODE" in
claude)
    exec env -i "${clean_env[@]}" \
        nono run --network-profile "$NONO_PROFILE" --profile "$NONO_PROFILE" -- ~/.local/bin/claude "$@"
    ;;
nono)
    [[ $# -lt 1 ]] && {
        echo "Error: 'nono' requires a subcommand"
        usage
    }
    exec env -i "${clean_env[@]}" \
        nono "$@"
    ;;
*)
    usage
    ;;
esac
```

lukehinds:
[...]

PR #694, opened 2026-04-16
feat: implements environment variables filtering #688

klassm:


klassm:
Review comments implemented.

crates/nono-cli/src/profile_runtime.rs
```diff
@@ -206,6 +207,16 @@ fn prepare_profile_with_options(
             &args.override_deny,
             workdir,
         ),
+        allowed_env_vars: loaded_profile.as_ref().and_then(|profile| {
+            profile.environment.as_ref().map(|env_config| {
+                if let Some(err) =
+                    crate::exec_strategy::validate_allow_vars_pattern(&env_config.allow_vars)
+                {
```
lukehinds, line 214: we tend to fail fatal when any security configurations load wrongly -  you can use `NonoError::ConfigParse` here.

lukehinds (commented):
really solid patch and thank you for taking this on, just one small change needed to fail on the config load. oh and thank you for cleaning up the clippy failure on the if let → match guard

crates/nono-cli/src/profile_runtime.rs
lukehinds: sorry, messed up my code line highliting:

```rust
if let Some(err) = validate_allow_vars_pattern(&env_config.allow_vars) {
    eprintln!("Warning: {}", err);
}
```

here you should be able to make the validator return `Result`, then `?` it at the call site. The `Option<String>` return shape already reads like a `Result`

rnestler:
@lukehinds is this included in the 0.37.1 release? Didn't see anything in the CHANGELOG.

PR #699, opened 2026-04-18
fix(hooks): invoke bash via env

xeals:
The read hook hardcodes the path to bash, which means it fails on distributions like NixOS that use a non-standard filesystem layout (that doesn't include symlinks for compatibility, like every distribution using the /usr layout).

This can be worked around by updating the shebang locally, but nono will overwrite it on every new session, removing the workaround.


lukehinds:
thanks @xeals !

lukehinds (approved):

Issue #594, opened 2026-04-05
Proposal: Clarify profile/policy/group terminology for better user comprehension


ranguard:
# Proposal: Clarify profile/policy/group terminology for better user comprehension

I've walked this through with claude a few times, getting it to look at the code and merging our thoughts on best user interface. This would be quite disruptive (changes in profile.json files) but figured sooner rather than later if you agree with it.. you could limit it to just the cli changes.

## Summary

The current CLI and profile JSON format overloads the term "policy" in three different ways, splits profile-related commands across two namespaces, and separates related concerns (group include/exclude, command allow/deny) into different sections. This creates unnecessary cognitive load for new users.

This proposal consolidates the user-facing terminology around two concepts: **profiles** (what users create and configure) and **groups** (the building blocks nono ships). "Policy" becomes purely internal.

## Key changes

**CLI:** Move all profile inspection commands (`show`, `list`, `diff`, `validate`, `groups`) into `nono profile`. Remove `nono policy` as a top-level command.

```
nono profile init <name>       # scaffold a new profile
nono profile list              # list all profiles (built-in and user)
nono profile show <name>       # show a fully resolved profile
nono profile validate <file>   # validate a profile file
nono profile diff <a> <b>      # compare two profiles
nono profile schema            # print the profile JSON Schema
nono profile guide             # print the profile authoring guide
nono profile groups [name]     # list built-in groups managed by nono
```

**Profile JSON:** Restructure so related concerns live together.

| Before | After | Why |
|--------|-------|-----|
| `security.groups` + `policy.exclude_groups` | `groups.include` + `groups.exclude` | Include/exclude in one place |
| `security.allowed_commands` + `policy.add_deny_commands` | `commands.allow` + `commands.deny` | Allow/deny in one place |
| `policy.add_allow_read/write/readwrite` | Merged into `filesystem.read/write/allow` | Semantically identical, no reason for two locations |
| `policy.add_deny_access` | `filesystem.deny` | Lives with other filesystem rules |
| `policy.override_deny` | `filesystem.bypass_protection` | Name conveys safety implications |

**CLI flag:** `--override-deny` becomes `--bypass-protection`.

## Motivation

- `nono policy validate` validates a *profile* file, not a "policy" — confusing
- `nono policy profiles` lists profiles, but `nono profile` has no `list` — split is arbitrary
- The `"policy"` key inside profile JSON means something different from `nono policy` — overloaded
[...]

PR #704, opened 2026-04-19
feat(cli): consolidate 'nono policy' subcommands under 'nono profile'…

ranguard:
… with deprecation alias (#594)

Move the five nono policy subcommands — groups, profiles, show, diff, validate — into the nono profile namespace as list, show, diff, validate, and groups. Output, flags, exit codes, and JSON shapes are preserved byte-for-byte; stdout now uses the unified 'nono profile:' label.

The old `nono policy <sub>` form continues to work via a dedicated shim at crates/nono-cli/src/deprecated_policy.rs: every invocation prints a one-line deprecation warning to stderr and delegates to the canonical handler. The shim re-exports arg types via 'pub use' aliases so there are no parallel type definitions to drift. Top-level 'nono policy --help' is labelled [deprecated]. Removal steps for when the deprecation window closes are recorded in the shim's header comment.

Includes new parser and integration tests for every migrated command, a byte-grep unit test on the embedded authoring guide to prevent future drift, a renamed integration-test file, refreshed site docs (policy-introspection.mdx -> profile-introspection.mdx), a CHANGELOG entry, and a rustfmt pass.


ranguard:
I mean't to open this against my fork! - sorry for the noise

PR #719, opened 2026-04-20
feat: add unsafe_macos_seatbelt_rules profile field

### LINKED ISSUE #678: enable raw macOS seatbelt rules in profiles

nono already has internal support for raw macOS platform rules via `CapabilitySet::add_platform_rule`, and the CLI uses that internally for niche cases like `(allow lsopen)`.

But there is no user-facing/profile-level way to express narrow macOS-only Seatbelt exceptions when we hit an unsupported capability gap. That means small gaps such as needing a rule like `(allow iokit-
open)` require a code change or a new first-class capability in nono, even when the user understands the risk and just needs an expert escape hatch.

Add a macOS-only, explicitly unsafe profile field for raw Seatbelt S-expressions, for example:

```yaml
unsafe_macos_seatbelt_rules:
  - "(allow iokit-open)"
```

Expected shape and constraints:

- macOS-only
- profile-only, not a casual CLI flag
- array of individual rules, not one large string blob
- clearly named as unsafe / expert-only
- surfaced prominently in inspect / load output so it is obvious a profile is using raw platform rules

This should be positioned as an escape hatch for advanced users, not the primary policy model. If a rule pattern becomes common, we should still prefer promoting it to a typed first-class capability later.

- Adding or considering new first-class capabilities for each missing Seatbelt operation
- Relying on internal add_platform_rule support in code
- Carrying one-off local patches

advaithsujith:
Closes #678

## Summary

- Adds `unsafe_macos_seatbelt_rules: Vec<String>` to the profile schema — a macOS-only expert escape hatch for raw Seatbelt S-expressions
- Rules are applied via the existing `add_platform_rule` path in `sandbox_prepare`, gated with `#[cfg(target_os = "macos")]`
- Malformed S-expressions are rejected at sandbox setup time with a clear error message
- A `WARN` log is emitted whenever rules are active so it is always visible in logs
- `nono policy show` surfaces the rules under a yellow **"Raw Seatbelt rules (unsafe_macos_seatbelt_rules):"** header — both in human and `--json` output
- Merged via `dedup_append` in profile inheritance (child rules appended after base)
- JSON schema updated with field description and usage guidance

## Test plan

- [x] `cargo build` passes cleanly
- [x] `nono policy show <profile>` displays rules under the yellow header
- [x] Valid rule `(allow iokit-open)` runs successfully inside sandbox
- [x] Malformed rule `"not an s-expression"` produces a clear `ConfigParse` error at startup
- [x] All 20 integration test suites pass (`bash tests/run_integration_tests.sh`)

### DIFF
--- a/crates/nono-cli/src/sandbox_prepare.rs
+++ b/crates/nono-cli/src/sandbox_prepare.rs
@@ -1110,7 +1110,7 @@ pub(crate) fn prepare_sandbox(args: &SandboxArgs, silent: bool) -> Result<Prepar
     #[cfg(target_os = "macos")]
     if let Some(ref profile) = loaded_profile {
         if !profile.unsafe_macos_seatbelt_rules.is_empty() {
-            warn!(
+            info!(
                 "Profile uses {} raw Seatbelt rule(s) via unsafe_macos_seatbelt_rules — review carefully",
                 profile.unsafe_macos_seatbelt_rules.len()
             );
--- a/crates/nono-cli/src/sandbox_prepare.rs
+++ b/crates/nono-cli/src/sandbox_prepare.rs
@@ -1106,6 +1106,24 @@ pub(crate) fn prepare_sandbox(args: &SandboxArgs, silent: bool) -> Result<Prepar
         CapabilitySet::from_args(args)?
     };
 
+    // Apply raw Seatbelt rules from the profile (macOS only).
+    #[cfg(target_os = "macos")]
+    if let Some(ref profile) = loaded_profile {
+        if !profile.unsafe_macos_seatbelt_rules.is_empty() {
+            warn!(
+                "Profile uses {} raw Seatbelt rule(s) via unsafe_macos_seatbelt_rules — review carefully",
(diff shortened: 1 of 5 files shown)


crates/nono-cli/src/sandbox_prepare.rs
```diff
@@ -1106,6 +1106,24 @@ pub(crate) fn prepare_sandbox(args: &SandboxArgs, silent: bool) -> Result<Prepar
         CapabilitySet::from_args(args)?
     };
 
+    // Apply raw Seatbelt rules from the profile (macOS only).
+    #[cfg(target_os = "macos")]
+    if let Some(ref profile) = loaded_profile {
+        if !profile.unsafe_macos_seatbelt_rules.is_empty() {
+            warn!(
```
lukehinds: can you change this to `!info` (line 1113) - otherwise we get it logged on the TUI, thanks!

advaithsujith: on it

PR #742, opened 2026-04-23
feat: make system keyring optional for headless/container builds

sahuguet:
Add a system-keyring feature flag (on by default) that controls whether the OS keyring (macOS Keychain / Linux Secret Service via libdbus) is compiled in.

- nono-cli's system-keyring feature is the single control point, propagating to nono and nono-proxy
- Gate keyring-using code with #[cfg(feature = "system-keyring")] with clear error fallbacks when disabled
- Remove unused direct keyring dep from nono-proxy
- Add docker/Dockerfile-headless for dbus-free container builds

Headless build: cargo build --release -p nono-cli --no-default-features
Docker: docker buildx build -f docker/Dockerfile-headless -t nono-headless .

In headless mode, credentials must use env://, file://, or op:// URIs.

Change tested by building and running `docker/Dockerfile-headless`.


lukehinds (commented):
solid approach thanks @sahuguet , just one thing that may need attention

On main, both `crates/nono/Cargo.toml` and `crates/nono-proxy/Cargo.toml` have no `[features]` section at all. `keyring = "3"` is an unconditional dependency. The PR adds `[features] default = []` plus the optional keyring dep to both crates. I think this means the next published version of nono and nono-proxy will ship with `default = []`, so downstream consumers who depend on nono (our libraries) and use `load_secret_by_ref("keyring://...")` or bare account names will start getting the new `KeystoreAccess` error at runtime unless they add `features = ["system-keyring"]`.

I guess one way to resolve this , set `default = ["system-keyring"]` on the nono and nono-proxy crates. Headless consumers then opt out with `default-features = false`, which is the normal Cargo idiom for "turn off the OS-integration dep." The nono-cli side of this PR can stay exactly as is and already handles the defaults.

sahuguet:
Do you want me to update the PR based on your comment?

lukehinds:
yes please, does it make sense (the change)?

sahuguet:
Changes added to the PR.
All tests passed. And I managed to build my Docker image without keyring,
@lukehinds

PR #744, opened 2026-04-23
feat(cli): add --allow-connect-port for outbound TCP port allowlisting

### LINKED ISSUE #743: Add outbound TCP port allowlisting via --allow-connect-port

When running claude code inside the sandbox, I want to restrict the ports claude can use.
I want to force all traffic to be http and have the traffic routed via the HTTP proxy used by nono.

Something like this
```
# Both ports allowed — mixed requests
  nono run --allow-connect-port 443 --allow-connect-port 80 -- \
    curl -fsS https://httpbin.org/get -o /dev/null && echo "OK"

  Should FAIL — port not in allowlist:
  # HTTP on port 80 — blocked (only 443 allowed)
  nono run --allow-connect-port 443 -- curl -fsS http://example.com -o /dev/null \
    && echo "FAIL: should have been blocked" || echo "OK: blocked"
```

_No response_

### How is this blocking you?

### Additional context

_No response_

sahuguet:
Adds --allow-connect-port PORT (repeatable) to SandboxArgs and WrapSandboxArgs, wiring caps.add_tcp_connect_port() through from_args, from_profile, and add_cli_overrides.

Also adds connect_port: Vec<u16> to NetworkConfig for profile-based configuration, merged via dedup_append consistent with open_port and listen_port.

On macOS, the flag emits a warning (Seatbelt cannot filter by TCP port). OS-level enforcement requires Linux Landlock V4+ (kernel >= 6.7).

Closes #743

### DIFF
--- a/crates/nono-cli/src/capability_ext.rs
+++ b/crates/nono-cli/src/capability_ext.rs
@@ -2360,8 +2360,8 @@ mod tests {
         let workdir = tempdir().expect("workdir");
         let args = sandbox_args();
 
-        let err = from_profile_locked(&profile, workdir.path(), &args)
-            .expect_err("should fail on macOS");
+        let err =
+            from_profile_locked(&profile, workdir.path(), &args).expect_err("should fail on macOS");
         assert!(
             err.to_string().contains("not supported on macOS"),
             "unexpected error: {err}"
--- a/crates/nono-cli/src/capability_ext.rs
+++ b/crates/nono-cli/src/capability_ext.rs
@@ -537,15 +537,17 @@ impl CapabilitySetExt for CapabilitySet {
         }
 
         // Outbound TCP connect port allowlist (Linux Landlock V4+ only)
-        for port in &args.allow_connect_port {
-            caps.add_tcp_connect_port(*port);
-        }
         #[cfg(target_os = "macos")]
         if !args.allow_connect_port.is_empty() {
-            warn!(
-                "--allow-connect-port has no effect on macOS: Seatbelt cannot filter by TCP port. \
-                 Use --allow-domain for host-level filtering instead."
-            );
+            return Err(NonoError::UnsupportedPlatform(
+                "--allow-connect-port is not supported on macOS: Seatbelt cannot filter by TCP port. \
+                 Use --allow-domain for host-level filtering, or ProxyOnly mode."
+                    .to_string(),
+            ));
+        }
+        #[cfg(not(target_os = "macos"))]
+        for port in &args.allow_connect_port {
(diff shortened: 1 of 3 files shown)


lukehinds (commented):
Quick thought on the macOS path ; right now we log "no effect on macOS" but then `generate_profile` in the library rejects a non-empty tcp_connect_ports and the run crashes with `NetworkFilterUnsupported`. So the user gets a warning saying the flag is ignored, followed by a hard failure. A bit confusing.

Could we flip it to fail up front instead? Something like: if the user passes `--allow-connect-port` on macOS, return a clean error at CLI parse time ,  `"--allow-connect-port isn't supported on macOS (Seatbelt can't filter by TCP port). Try --allow-domain or ProxyOnly mode."` - and don't add the port to caps at all.

The reason I'd lean that way rather than "warn and skip": if someone runs `--allow-connect-port 443` they're expecting the sandbox to block everything else. Silently continuing without that restriction gives them a false sense of safety. Better to stop and point them at the flag that actually works on their platform.

Happy either way though - just wanted to flag that the current behavior (warn, then crash) is the one outcome we probably don't want. Being a security tool, its always ok to fail hard (safetly)

sahuguet:
Addressed the macOS behavior in the latest commit.

**Before**: on macOS, `--allow-connect-port` would log a warning ("no effect on macOS") but still add the ports to `CapabilitySet`, causing a hard crash later in `generate_profile` with `NetworkFilterUnsupported`. The user got a warning saying the flag was ignored, followed by a failure — the worst of both worlds.

**After**: on macOS, passing `--allow-connect-port` (or setting `connect_port` in a profile) now returns a clean error immediately at capability-build time, before the sandbox is ever applied:

```
--allow-connect-port is not supported on macOS: Seatbelt cannot filter by TCP port. Use --allow-domain for host-level filtering, or ProxyOnly mode.
```

This applies to both the CLI flag and the profile `connect_port` field. The rationale: if a user passes `--allow-connect-port 443` they expect the sandbox to block everything else — silently continuing without that restriction gives a false sense of safety. Failing hard and pointing to the working alternatives (`--allow-domain`, ProxyOnly mode) is the right behavior for a security tool.

Tests are gated with `#[cfg(target_os = "macos")]` / `#[cfg(not(target_os = "macos"))]` to exercise the correct behavior on each platform.

cc @lukehinds

lukehinds:
just one more @sahuguet , run `make fmt` and should be good - apologies , I am really wanting to get this merged for you.

as a little tip going forward , if you run `make ci` it will check most of the tests here are checked before you push.

sahuguet:
`make fmt` run.

PR #762, opened 2026-04-25
fix(proxy): stop adding allow_domain hosts to NO_PROXY without direct TCP grants

### LINKED ISSUE #760: Linux proxy regression
### What happened?

# Bug: exact-match `allow_domain` entries stop routing through the reverse proxy on Linux starting with v0.39.0

## Summary

Between v0.38.0 and v0.39.0 on Linux (`x86_64-unknown-linux-gnu`),
exact-match hostnames in `network.allow_domain` stopped being routed
through the reverse proxy. Wildcard entries (`*.example.com`) still
work. macOS (`aarch64-apple-darwin`) is unaffected across the same
version range.

The sandboxed process sees the entry in the resolved allowlist
(`nono policy show --json` confirms), but `curl` against that host
fails with `(7) Couldn't connect to server` — i.e. it bypasses the
proxy and tries a direct connect, which the sandbox correctly denies.

## Environment

- nono versions tested: **v0.38.0**, **v0.39.0**, **v0.40.0**, **v0.40.1**
- Targets: `x86_64-unknown-linux-gnu`, `aarch64-apple-darwin`
- Runners: GitHub Actions `ubuntu-latest` and `macos-latest`
- Binary installed from the official release tarballs on
  `https://github.com/always-further/nono/releases/`

## Observed results

|   nono    | Linux `github.com` (exact) | Linux `releases.hashicorp.com` (wildcard) | macOS `github.com` (exact) | macOS `releases.hashicorp.com` (wildcard) |
|:---------:|:--------------------------:|:-----------------------------------------:|:--------------------------:|:-----------------------------------------:|
| v0.38.0   | ✅ HTTP 200                | ✅ HTTP 200                               | ✅ HTTP 200                | ✅ HTTP 200                               |
| v0.39.0   | ❌ curl: (7)               | ✅ HTTP 200                               | ✅ HTTP 200                | ✅ HTTP 200                               |
| v0.40.0   | ❌ curl: (7)               | ✅ HTTP 200                               | ✅ HTTP 200                | ✅ HTTP 200                               |
| v0.40.1   | ❌ curl: (7)               | ✅ HTTP 200                               | ✅ HTTP 200                | ✅ HTTP 200                               |

- HTTP 200 = the sandboxed curl reached the host through the proxy and
  the origin responded successfully.
- `curl: (7)` = full error message:
  `curl: (7) Failed to connect to github.com port 443 after N ms: Couldn't connect to server`.
  The sandbox blocked a direct connect, which is correct — but the
  proxy should have intercepted this host because it is in `allow_domain`.

## Expected behavior

Exact-match entries in `network.allow_domain` should be routed through
the reverse proxy in the same way wildcard entries are — and the same
way exact entries were routed in v0.38.0 on Linux. There should be no
divergence between exact and wildcard handling, and no divergence
between Linux and macOS.

## Additional observations

- `nono policy validate` passes on all versions — the profile is
  structurally valid.
- `nono policy show --json` on all versions shows
  `"allow_domain": ["github.com", "*.hashicorp.com"]` in the resolved
  profile, so the value reaches the runtime layer.
- Swapping `github.com` with any other exact host
  (`registry.terraform.io`, `releases.hashicorp.com`,
  `accounts.google.com`, ...) reproduces the same failure on Linux
  v0.39.0+.
- Converting the same host into a wildcard (`*.terraform.io`,
  `*.hashicorp.com`, `*.google.com`) makes it work again on Linux
  v0.39.0+ — but there is no wildcard form that covers a TLD-apex
  (`github.com` itself cannot be expressed as a suffix wildcard), so
  some hosts cannot be worked around by the profile author.

## Impact

Any profile that whitelists an apex domain (`github.com`,
`example.com`, etc.) and is meant to be used on Linux with a nono
release ≥ v0.39.0 is effectively broken for that host. Users who only
develop on macOS will not see the issue locally; Linux CI or Linux
devcontainers will.

## Suspected area

The regression is between v0.38.0 and v0.39.0 on the Linux proxy /
interception path. A `git log --oneline v0.38.0..v0.39.0 -- <proxy/
network code>` should narrow it down quickly.

Same behaviour across all versions and distributions

## Minimal reproducer

`repro.json`:

```json
{
  "extends": "claude-code",
  "meta": { "name": "repro", "version": "1.0.0" },
  "network": {
    "allow_domain": [
      "github.com",
      "*.hashicorp.com"
    ]
  }
}
```

Probes (run both from the same shell):

```bash
# Exact-match entry in allow_domain:
nono run -s --allow-cwd --profile ./repro.json -- \
  curl -sS --max-time 5 --connect-timeout 3 -o /dev/null \
       -w '%{http_code}\n' https://github.com/

# Wildcard entry in allow_domain:
nono run -s --allow-cwd --profile ./repro.json -- \
  curl -sS --max-time 5 --connect-timeout 3 -o /dev/null \
       -w '%{http_code}\n' https://releases.hashicorp.com/
```

### Relevant output or logs

```shell

```

### nono version

0.39 + 

### Operating system

Linux

### OS version / distro

githubn runner ubuntu-latest

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

4x0v7:
After #720 removed ConnectTcp grants from `--allow-domain`, allowed hosts still appeared in NO_PROXY. HTTP clients that honour NO_PROXY attempted direct connections that the sandbox denied, breaking tools like curl on Linux.

Add direct_connect_ports to ProxyConfig, populated from the capability set's tcp_connect_ports. The NO_PROXY computation now only includes hosts whose port is in this set. With no connect-port grants the list is empty and all allowed-domain traffic routes through the proxy.

Also fix warn! -> info! in try_new_unix_socket_{file,dir} to match the log-level change in 024ca6f that missed these call sites added by 85708ca.

---

Forward-compatible with #744: when `--allow-connect-port` / `connect_port` lands, ports will propagate through `tcp_connect_ports()` into `direct_connect_ports`, and matching hosts will reappear in `NO_PROXY`.

Fixes #760

### DIFF
--- a/crates/nono-cli/src/capability_ext.rs
+++ b/crates/nono-cli/src/capability_ext.rs
@@ -55,7 +55,7 @@ fn try_new_unix_socket_file(
     match UnixSocketCapability::new_file(path, mode) {
         Ok(cap) => Ok(Some(cap)),
         Err(NonoError::PathNotFound(_)) => {
-            warn!("{}: {}", label, path.display());
+            info!("{}: {}", label, path.display());
             Ok(None)
         }
         Err(e) => Err(e),
@@ -71,7 +71,7 @@ fn try_new_unix_socket_dir(
     match UnixSocketCapability::new_dir(path, mode) {
         Ok(cap) => Ok(Some(cap)),
         Err(NonoError::PathNotFound(_)) => {
-            warn!("{}: {}", label, path.display());
+            info!("{}: {}", label, path.display());
             Ok(None)
         }
         Err(e) => Err(e),
--- a/crates/nono-cli/src/proxy_runtime.rs
+++ b/crates/nono-cli/src/proxy_runtime.rs
@@ -190,7 +190,8 @@ pub(crate) fn start_proxy_runtime(
         });
     }
 
-    let proxy_config = build_proxy_config_from_flags(proxy)?;
+    let mut proxy_config = build_proxy_config_from_flags(proxy)?;
+    proxy_config.direct_connect_ports = caps.tcp_connect_ports().to_vec();
     let rt = tokio::runtime::Builder::new_multi_thread()
         .worker_threads(2)
         .enable_all()
--- a/crates/nono-proxy/src/config.rs
+++ b/crates/nono-proxy/src/config.rs
@@ -47,6 +47,12 @@ pub struct ProxyConfig {
     #[serde(default)]
     pub external_proxy: Option<ExternalProxyConfig>,
 
+    /// Outbound TCP ports that the sandbox allows direct connections on
+    /// (via Landlock ConnectTcp). Hosts whose resolved port is NOT in this
+    /// set must go through the proxy and should NOT appear in NO_PROXY.
+    #[serde(default)]
+    pub direct_connect_ports: Vec<u16>,
+
     /// Maximum concurrent connections (0 = unlimited).
     #[serde(default)]
     pub max_connections: usize,
@@ -60,6 +66,7 @@ impl Default for ProxyConfig {
             allowed_hosts: Vec::new(),
             routes: Vec::new(),
             external_proxy: None,
+            direct_connect_ports: Vec::new(),
             max_connections: 256,
         }
     }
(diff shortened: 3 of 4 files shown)


lukehinds:
Thanks @4x0v7 looking at this now.

PR #774, opened 2026-04-27
fix: unify path canonicalization with ancestor-walk fallback

### LINKED ISSUE #257: Unify path canonicalization with ancestor-walk fallback
## Problem

Path canonicalization is inconsistent across the codebase. There are **three different fallback strategies** used when `std::fs::canonicalize()` fails (typically because the path doesn't exist yet):

### 1. Ancestor-walk fallback (robust, correct)
Walks up the path to find the longest existing ancestor, canonicalizes it, and appends remaining components. This correctly handles cases like `/tmp/nonexistent/child` on macOS where `/tmp` → `/private/tmp`.

Used in:
- `crates/nono/src/supervisor/never_grant.rs` — `resolve_path()`
- `crates/nono-cli/src/protected_paths.rs` — `resolve_path()` (identical copy)

### 2. Parent-only fallback (partial)
Only tries to canonicalize the immediate parent (1 level up). Misses cases where multiple path components don't exist.

Used in:
- `crates/nono-cli/src/query_ext.rs` — `query_path()`

### 3. Simple fallback to original path (weakest)
Just returns the original path unchanged if canonicalization fails. On macOS, this means `/tmp/foo` stays as `/tmp/foo` instead of resolving to `/private/tmp/foo`, leading to incorrect path comparisons.

Used in:
- `crates/nono/src/query.rs` — `query_path()` (library)
- `crates/nono-cli/src/exec_strategy/supervisor_linux.rs` — seccomp handler
- `crates/nono-cli/src/learn.rs` — path deduplication
- `crates/nono-cli/src/rollback_commands.rs` — filter matching

### Not affected (hard error is correct)
- `capability.rs:new_dir/new_file()` — path must exist, hard error appropriate
- `exec_strategy.rs:open_path_for_access()` — opening a real file, hard error appropriate
- `policy.rs` — config paths must resolve, hard error appropriate

## Impact

On macOS, the simple fallback sites can produce incorrect security decisions when paths involve symlinks (`/tmp`, `/var`, `/etc` all symlink to `/private/*`). A non-existent path under `/tmp/` won't canonicalize, so it stays as `/tmp/...` while capability paths are stored as `/private/tmp/...` — the comparison fails silently.

## Proposed Fix

1. Extract a shared `try_canonicalize()` utility function that implements the ancestor-walk fallback
2. Place it in the library (`crates/nono/src/`) since both library and CLI code need it
3. Replace all 6 inconsistent sites with calls to the shared utility
4. Remove the duplicate `resolve_path()` implementations in `never_grant.rs` and `protected_paths.rs`
5. The function signature: `pub fn try_canonicalize(path: &Path) -> PathBuf`

## Sites to update

| File | Current approach | Action |
|---|---|---|
| `nono/src/supervisor/never_grant.rs` | ancestor-walk (private fn) | Replace with shared utility |
| `nono-cli/src/protected_paths.rs` | ancestor-walk (private fn, identical copy) | Replace with shared utility |
| `nono-cli/src/query_ext.rs` | parent-only (1 level) | Replace with shared utility |
| `nono/src/query.rs` | `.ok()` fallback to raw path | Replace with shared utility |
| `nono-cli/src/exec_strategy/supervisor_linux.rs` | `.unwrap_or_else` to raw path | Replace with shared utility |
| `nono-cli/src/learn.rs` | `.unwrap_or` to raw path | Replace with shared utility |
| `nono-cli/src/rollback_commands.rs` | `.unwrap_or_else` to raw path | Replace with shared utility |

advaithsujith:
## Summary

- Extracts a shared `try_canonicalize(path: &Path) -> PathBuf` utility into `nono::path`, using the correct ancestor-walk strategy that was already implemented (but duplicated) in `protected_paths.rs`
- Replaces 7 inconsistent call sites that used weaker fallbacks (raw path or parent-only) with the shared utility
- Removes the duplicate `resolve_path` private function from `protected_paths.rs`

**Why it matters:** On macOS, `/tmp`, `/var`, and `/etc` are symlinks to `/private/tmp` etc. The weak fallbacks left these unresolved for non-existent paths, causing silent security comparison failures when capability paths (stored as `/private/tmp/...`) were compared against query paths that stayed as `/tmp/...`.

## Sites updated

| File | Old approach | New |
|---|---|---|
| `nono/src/query.rs` | `.ok()` → raw path | `try_canonicalize` |
| `nono-cli/src/query_ext.rs` | parent-only (1 level) | `try_canonicalize` |
| `nono-cli/src/exec_strategy/supervisor_linux.rs` | `.unwrap_or_else` → raw path | `try_canonicalize` |
| `nono-cli/src/learn.rs` | `.unwrap_or` → raw path | `try_canonicalize` |
| `nono-cli/src/rollback_commands.rs` | `.unwrap_or_else` → raw path | `try_canonicalize` |
| `nono-cli/src/protected_paths.rs` | private `resolve_path` (ancestor-walk) | shared `try_canonicalize` |

## Test plan

- [x] `cargo test -p nono` passes (576 tests)
- [x] `cargo build` clean with no warnings
- [x] New unit tests in `nono::path` cover: existing path, non-existent leaf, non-existent nested path, symlink resolution

Closes #257

### DIFF
--- a/crates/nono-cli/src/exec_strategy/supervisor_linux.rs
+++ b/crates/nono-cli/src/exec_strategy/supervisor_linux.rs
@@ -11,7 +11,7 @@
 
 use super::*;
 use crate::trust_intercept::TrustInterceptor;
-use nono::AccessMode;
+use nono::{try_canonicalize, AccessMode};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(super) struct InitialCapability {
@@ -231,8 +231,7 @@ pub(super) fn handle_seccomp_notification(
             return Ok(());
         }
     };
-    let canonicalized =
-        std::fs::canonicalize(&resolved_path).unwrap_or_else(|_| resolved_path.clone());
+    let canonicalized = try_canonicalize(&resolved_path);
 
     // For the initial capability match, map a grandchild's /proc/<tgid> path back to the
     // direct child's /proc/<child_pid>, because initial_caps are built from the direct
--- a/crates/nono-cli/src/learn.rs
+++ b/crates/nono-cli/src/learn.rs
@@ -5,7 +5,7 @@
 //! to be allowed in a nono profile.
 
 use crate::cli::LearnArgs;
-use nono::{AccessMode, NonoError, Result};
+use nono::{try_canonicalize, AccessMode, NonoError, Result};
 use std::collections::BTreeSet;
 use std::net::IpAddr;
 use std::path::PathBuf;
@@ -1565,7 +1565,7 @@ fn process_accesses(
 
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 fn canonicalize_existing_path(path: &Path) -> PathBuf {
-    path.canonicalize().unwrap_or_else(|_| path.to_path_buf())
+    try_canonicalize(path)
 }
 
 #[cfg(any(target_os = "linux", target_os = "macos"))]
--- a/crates/nono-cli/src/protected_paths.rs
+++ b/crates/nono-cli/src/protected_paths.rs
@@ -44,12 +44,17 @@ pub fn validate_caps_against_protected_roots(
     protected_roots: &[PathBuf],
     allow_parent_of_protected: bool,
 ) -> Result<()> {
+    // Pre-canonicalize once so the per-capability loop doesn't repeat the work.
+    let resolved_roots: Vec<PathBuf> = protected_roots
+        .iter()
+        .map(|p| try_canonicalize(p))
+        .collect();
     for cap in caps.fs_capabilities() {
         validate_requested_path_against_protected_roots(
             &cap.resolved,
             cap.is_file,
             &cap.source.to_string(),
-            protected_roots,
+            &resolved_roots,
             allow_parent_of_protected,
         )?;
     }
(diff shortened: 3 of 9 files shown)


lukehinds:
One more spot to migrate: crates/nono/src/diagnostic.rs (1676 from just checking, but might be a diffrerent line number in this PR, if so just just search for "canonicalize_query_path"

There's a function called `canonicalize_query_path` that uses the exact same parent-only fallback you just cleaned up in query_ext.rs:

```rust
  fn canonicalize_query_path(path: &Path) -> PathBuf {
      if path.exists() {
          path.canonicalize().unwrap_or_else(|_| path.to_path_buf())
      } else if let Some(parent) = path.parent() {
          if parent.exists() {
              match parent.canonicalize() { ... }
          }
          ...
      }
  }
```

This one isn't a security boundary - it's only used to format diagnostic output when a sandbox denial is reported, so leaving it won't cause the same /tmp vs /private/tmp bug class as the sites you fixed. But since the main point of this PR is "one shared helper, no more drifting copies," may as well cover this one as well.

Two options, either is fine:

1. Migrate it, replace the body with `try_canonicalize(path)` and delete the helper. Quick win, keeps the codebase consistent.
2. Keep it but add a one-line comment explaining why diagnostic formatting deliberately uses its own helper (e.g. "we want to show the user the path they typed, not a resolved one").

If neither reason applies, option 1 is the cleaner finish and a quick win.

advaithsujith:
understood, on it

advaithsujith:
Addressed in 69c55f4 — replaced the local `canonicalize_query_path` with the shared `try_canonicalize` from `crate::path` and deleted the helper. No more drifting copies.

lukehinds (approved):

PR #796, opened 2026-04-29
feat(env): add operator-controlled deny_vars to EnvironmentConfig

### LINKED ISSUE #783: env var blocking: operator-controlled deny-list
nono supports an env allow-list and a hardcoded dangerous-var blocklist, but **no operator-configurable deny-list**. 

Present (`crates/nono-cli/src/exec_strategy/env_sanitization.rs`):
- Hardcoded `is_dangerous_env_var` covering `LD_*`, `DYLD_*`, `BASH_ENV`, `PROMPT_COMMAND`, `IFS`, `NODE_OPTIONS`, `PYTHONPATH`, `PERL5OPT`, `RUBYOPT`, `JAVA_TOOL_OPTIONS`, `OP_SERVICE_ACCOUNT_TOKEN`, `OP_SESSION_*`, etc. Always stripped, not configurable.
- `EnvironmentConfig.allow_vars` in `crates/nono-cli/src/profile/mod.rs:1147` — exact names + trailing-`*` prefixes (`AWS_*`). When set, only listed vars pass through.
- `should_skip_env_var` (line 98) already accepts a `blocked_extra: &[&str]` parameter, so the plumbing for a config-driven deny-list is half there.

Missing:
- No `EnvironmentConfig.deny_vars` field. Operators who want to strip `GH_TOKEN`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY` while keeping everything else inherited have to either flip to allow-list mode (high friction — must enumerate everything) or patch the hardcoded list in source.

### Proposed shape

Extend `EnvironmentConfig` in `crates/nono-cli/src/profile/mod.rs`:

```rust
pub struct EnvironmentConfig {
    /// Existing: when non-empty, only these vars pass through.
    #[serde(default)]
    pub allow_vars: Vec<String>,

    /// New: vars matching these patterns are stripped, even if otherwise allowed.
    /// Same matcher as allow_vars (exact names or `PREFIX_*`).
    #[serde(default)]
    pub deny_vars: Vec<String>,
}
```

Precedence (documented and tested):
1. Hardcoded `is_dangerous_env_var` — always stripped (cannot be re-allowed).
2. `deny_vars` — stripped, even if matched by `allow_vars`.
3. `allow_vars` — if non-empty, only matching vars pass; if empty, all (except 1+2) pass.

Profile-merge semantics: `deny_vars` from base and child are concatenated (same as `allow_vars` via `dedup_append`).

### Use cases

```toml
# Strip GitHub-related tokens for an agent that should not see them
[env]
deny_vars = ["GH_TOKEN", "GITHUB_TOKEN", "GITHUB_*"]
```

```toml
# Strip all API keys but keep everything else
[env]
deny_vars = ["*_API_KEY", "*_TOKEN", "ANTHROPIC_API_KEY", "OPENAI_API_KEY"]
```

(Note: trailing-`*` only — match the existing pattern matcher's restriction.)

### Acceptance criteria

- [ ] `deny_vars` field added to `EnvironmentConfig`
- [ ] `validate_allow_vars_pattern` reused/renamed for both fields
- [ ] `should_skip_env_var` integrates `deny_vars` ahead of `allow_vars`
- [ ] Merge logic in `profile/mod.rs:2091` extended for `deny_vars`
- [ ] Tests: deny overrides allow; deny + dangerous compose; bare `*` rejected in `deny_vars` (or explicitly allowed — TBD)
- [ ] Docs updated: `docs/cli/features/environment.mdx`

advaithsujith:
Closes #783

## Summary

- Adds `deny_vars: Vec<String>` to `EnvironmentConfig` — strips named variables from the sandboxed process without requiring a full allow-list
- Same trailing-`*` prefix pattern syntax as `allow_vars` (`"GITHUB_*"`, `"GH_TOKEN"`, etc.)
- Precedence: built-in dangerous-var blocklist > `deny_vars` > `allow_vars` (deny wins over allow)
- Profile merge: `deny_vars` from base and child are concatenated with dedup, identical to `allow_vars`
- `validate_allow_vars_pattern` renamed to `validate_env_var_patterns(patterns, field_name)` so both fields share the same validator with correct field names in warnings
- Empty `allow_vars` is now a no-op (returns `None`, not `Some([])`) so profiles that set only `deny_vars` don't accidentally block the entire environment

## Test plan

- [x] `cargo test -p nono-cli` — 866 passed, 0 failed
- [x] `bash tests/run_integration_tests.sh` — 20/20 suites passed
- [x] Smoke tested `deny_vars: ["MY_SECRET_TOKEN"]`: token stripped, other vars pass through
- [x] Smoke tested prefix `deny_vars: ["GITHUB_*"]`: all `GITHUB_*` vars stripped
- [x] Smoke tested `deny_vars` overrides `allow_vars: ["MY_*"]`: denied var still stripped


crates/nono-cli/src/profile_runtime.rs
```diff
@@ -378,13 +379,31 @@ fn prepare_profile_with_options(
             workdir,
         ),
         allowed_env_vars: loaded_profile.as_ref().and_then(|profile| {
-            profile.environment.as_ref().map(|env_config| {
-                if let Some(err) =
-                    crate::exec_strategy::validate_allow_vars_pattern(&env_config.allow_vars)
-                {
+            profile.environment.as_ref().and_then(|env_config| {
+                if env_config.allow_vars.is_empty() {
+                    return None;
```
lukehinds: I think the early-return might flip the intent of an empty allow_vars from fail-closed to fail-open

On main, an environment block with empty allow_vars returns Some(vec![]), which exec_strategy reads as "strip every inherited var." After this change it becomes `None`, which means "no filtering." The case I'd worry about: a profile with env_credentials: [...] and no allow_vars is locked down today to only the injected secret -  after this it would also inherit every env var from the parent, silently on upgrade.

I'd keep the fail-closed behavior here. Drop the early-return and use the current approach from main:

https://github.com/always-further/nono/blob/7914fb391057bfc57329a44bcec1e613520b21b1/crates/nono-cli/src/profile_runtime.rs#L380-L391

That preserves today's semantics: environment absent → no filtering; environment present with empty allow_vars → strip all inherited vars which is. a safer approach. So pretty simple change, just knock that bit of code back to the same as before and we are sweet and ready to merge!

lukehinds (approved):

PR #804, opened 2026-04-30
fix: update example in Quickstart examples after `nono setup` is run

zaxtax:
When `nono setup` is run some examples are given. The last one is not accurate.


lukehinds:
thanks @zaxtax 🙏 great having you here.

lukehinds (approved):

PR #806, opened 2026-04-30
fix: resolve extends against sibling profiles in the same directory

### LINKED ISSUE #750: bug: user-defined profiles aren't taken into account if not saved to default directory
### What happened?

I want to create a shared profile that defines `filesystem` settings for the other two profiles in our repository. Extending it doesn't work if the user-defined shared profile is not saved in the default `.config/nono/profiles` directory.

`nono` either searches for nearby profile files or allows you to specify the profile location for the `extends` directive.

### Steps to reproduce

1. Create profiles in the `.nono` directory, for example, `filesystem.json`.  
2. In the target profile, extend it with the expected profile using `extends: ["claude-code", "filesystem"]`.  

### Relevant output or logs

```shell
nono: Profile inheritance error: base profile 'filesystem' not found
```

### nono version

0.39.0

### Operating system

macOS

### OS version / distro

macOS 2026.4.1

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

SequeI:
Profile extends resolution only searched ~/.config/nono/profiles/, pack-store, and built-ins. Project-local profiles (e.g. in .nono/) that extended each other by name failed with 'base profile not found'. Thread the parent directory as context through the inheritance chain so sibling <name>.json files are checked first before falling back to global locations.

```
nono profile show .nono/my-profile.json 2>&1

nono profile: profile '.nono/my-profile.json'
  Extends:      default, filesystem
  
2026-04-30T15:20:30.955473Z DEBUG Resolved 'filesystem' from sibling: .nono/filesystem.json
```

Closes #750 

@skoch13 please let me know if this works for you :)

### DIFF
--- a/crates/nono-cli/src/profile/mod.rs
+++ b/crates/nono-cli/src/profile/mod.rs
@@ -1784,7 +1784,7 @@ pub(crate) fn finalize_profile(mut profile: Profile) -> Result<Profile> {
 
 /// Resolve inheritance and apply implicit default-group merging for a raw profile.
 pub(crate) fn resolve_and_finalize_profile(profile: Profile) -> Result<Profile> {
-    finalize_profile(resolve_extends(profile, &mut Vec::new(), 0)?)
+    finalize_profile(resolve_extends(profile, &mut Vec::new(), 0, None)?)
 }
 
 /// Get the implicit default groups for a finalized profile.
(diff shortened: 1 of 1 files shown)


skoch13:
Yes it works, thank you! ❤️

lukehinds:
Taking a look, but first thoughts, .nono is a protected space as it contains audit and snapshots and we don't want the sandboxed child to have access to those (tamper protections) - is there a particular need for storing profiles in .nono ? 

```
❯ ls .nono
audit  rollbacks  sessions
```

providing access to .nono , gets recursive access on the above.

skoch13:
@lukehinds in my example this `.nono` folder is located in the root of the repo, so the profiles can be shared within the team. is there a better option?

lukehinds:
@skoch13 , my bad, that's my misread - I thought we were referring to `~.nono/` which is where the said files are.

so if '.nono/my-profile.json' is relative and not to $HOME - no issue.

PR #823, opened 2026-05-04
feat(profile): support env:// URI in custom_credentials credential_key

### LINKED ISSUE #820: Support env:// URI scheme in credential_key for custom credentials

I want to use env:// in credential_key for custom credentials, but it's not supported. The credential_key field currently accepts bare keyring names, op://, apple-password://, and file:// URIs — but rejects env:// even though it's already fully supported in env_credentials and in the runtime load_secret_by_ref() dispatcher.

I want to use "credential_key": "env://X_TOKEN" to reference credentials already in my environment without needing file:// with a hardcoded absolute path.

- "credential_key": "env://X_TOKEN" passes profile validation
- env_var is optional when using env:// (derived from the URI)
- env:// with empty var name or dangerous vars (LD_PRELOAD, etc.) is rejected
- Proxy phantom token injection works without explicit env_var for env:// keys

I tried using file:// but it requires absolute file path, not work with $HOME, and store ephemeral credentials into disk which is not good.

### How is this blocking you?

### Additional context

_No response_

SequeI:
- Accept env:// in credential_key for custom credentials, matching existing support in env_credentials and load_secret_by_ref()
- env_var is optional for env:// (derived from the URI), required for op://, apple-password://, and file://
- Dangerous variables (LD_PRELOAD, PATH, etc.) rejected via existing validate_env_uri blocklist
- Update schema, profile authoring guide, and credential injection docs

Closes #820 

cc @panga please let me know if this works for you now :)

### DIFF
--- a/crates/nono-cli/data/nono-profile.schema.json
+++ b/crates/nono-cli/data/nono-profile.schema.json
@@ -396,7 +396,7 @@
         },
         "credential_key": {
           "type": "string",
-          "description": "Keystore account name for the credential (e.g., \"telegram_bot_token\"), or a 1Password op:// URI, or an apple-password:// URI. Mutually exclusive with auth."
+          "description": "Keystore account name for the credential (e.g., \"telegram_bot_token\"), or a 1Password op:// URI, apple-password:// URI, file:// URI, or env:// URI (e.g., \"env://MY_TOKEN\"). Mutually exclusive with auth."
         },
         "auth": {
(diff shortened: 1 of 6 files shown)


panga:
@SequeI yes, that works great!

SequeI:
cc @lukehinds We can get this in for 0.50 if possible, will unblock this user

lukehinds:
reviewing now

lukehinds (approved):
lgtm

PR #835, opened 2026-05-05
fix(profile): skip self-references in sibling extends resolution

### LINKED ISSUE #832: bug: nono: Profile inheritance error: circular dependency detected: codex -> codex
### What happened?

Updated to the most recent version tried to start with the profile that extends `codex`

It should work as expected

### Steps to reproduce

1. Update to the most recent 0.47.0 version
2. start nono with profile that extends built-in

### Relevant output or logs

```shell

```

### nono version

0.47.0

### Operating system

macOS

### OS version / distro

macOS 26.4.1

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

SequeI:
A profile like '.nono/codex.json' extending 'codex' would match itself as a sibling, causing a circular dependency error instead of resolving to the pack-installed codex profile.

- Thread 'source_file' through 'resolve_extends' and 'load_base_profile_raw'
- Skip sibling matches that point to the source file itself
- Carry resolved path in 'ResolvedBase::Sibling' for recursive propagation

Closes #832 

cc @skoch13 let me know if this works now please if possible :)

### DIFF
--- a/crates/nono-cli/src/profile/mod.rs
+++ b/crates/nono-cli/src/profile/mod.rs
@@ -1848,7 +1848,7 @@ pub(crate) fn finalize_profile(mut profile: Profile) -> Result<Profile> {
 
 /// Resolve inheritance and apply implicit default-group merging for a raw profile.
 pub(crate) fn resolve_and_finalize_profile(profile: Profile) -> Result<Profile> {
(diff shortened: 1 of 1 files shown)


skoch13:
@SequeI it works, thank you!

PR #846, opened 2026-05-06
feat(wiring): add yaml_merge directive

### LINKED ISSUE #819: Implement  yaml_merge wiring for packages
alongside the existing  json_merge we require  yaml_merge as helper for agents whose setup requires manipulation of YAML:

e.g.

```yaml
plugins:
    enabled:
      - nono-sandbox
```

would be resolved with

```json
  {
    "type": "yaml_array_remove",
    "file": "$HOME/.hermes/config.yaml",
    "path": "plugins.enabled",
    "entries": ["nono-sandbox"]
  }
```

advaithsujith:
## Summary

- Adds `WiringDirective::YamlMerge { file, patch }` mirroring `JsonMerge` semantics for YAML files
- Parses YAML via `serde_yaml_ng` and projects onto `serde_json::Value` so `walk_and_merge` / `restore_one_leaf` / `prune_empty_object_at` are reused unchanged
- `WiringRecord::YamlMerge` reuses `JsonLeaf` for per-leaf reversal — idempotent, reversible, user-edit safe
- Strict: non-string mapping keys → error, custom YAML tags (`!foo`) → error, non-finite floats (`.inf`/`.nan`) → error (prevents silent `null` corruption)
- Lossy round-trip on comments/anchors, same as `JsonMerge`

## Test plan

- [x] `yaml_merge_and_strip_round_trip` — merge into existing file, verify patch applied, reverse, verify only merged keys removed
- [x] `yaml_merge_is_idempotent` — second run reports `changed = false`
- [x] `yaml_merge_preserves_sibling_keys_on_reverse` — user's pre-existing sibling keys survive uninstall
- [x] `yaml_merge_rejects_non_string_keys` — integer mapping key returns clear error
- [x] `yaml_merge_rejects_custom_tags` — `!MyTag` returns clear error
- [x] All 21 integration test suites pass

Closes #819

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2647,9 +2647,9 @@ dependencies = [
 
 [[package]]
 name = "serde_yaml_ng"
-version = "0.9.36"
+version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bd24347956e682cf958c95e82deb9914cad4010d3efc035d579f81f4c426038c"
+checksum = "7b4db627b98b36d4203a7b458cf3573730f2bb591b28871d916dfa9efabfd41f"
 dependencies = [
  "indexmap",
  "itoa",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1685,6 +1685,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
+ "serde_yaml_ng",
  "sha2 0.11.0",
  "shlex",
  "sigstore-sign",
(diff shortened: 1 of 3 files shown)


lukehinds:
Thanks for picking this up, just a couple of bits - its worth having a yaml equivalent of reverse_collects_failures_instead_of_swallowing that we have on json - this way if something breaks, we back out and don't leave a misformatted config , its also worth pinning like the other dependencies - I checked on crates and it seems the latest is 0.10.0

PR #847, opened 2026-05-06
fix: demote --allow-launch-services log from warn to debug

### LINKED ISSUE #816: remove warn line on --allow-launch-services enabled
we need to stop warning on TUIs,  switch to a debug - for   `warn!("--allow-launch-services enabled: allowing direct LaunchServices opens on macOS");"`

SequeI:
Closes #816 

Only one call site, so used fully qualified import name


lukehinds:
Nice, we don't need these hitting the TUI!

PR #851, opened 2026-05-07
fix(trust): reject path traversal in multi-subject bundle subject names

lukehinds:
Subject names extracted from Sigstore bundle DSSE envelopes were joined with scan_root via Path::join without validation. An attacker who controls a signing key could embed (when much worse things could happen), could embed:

    - `../../../etc/passwd` (relative traversal) – OS resolves the join to a path outside scan_root.
    - `/etc/passwd` (absolute path) – Path::join discards scan_root entirely for absolute inputs

Fixed care of: add `safe_subject_path(scan_root, name)` which rejects names before any path construction:

1. Absolute paths – detected via `Path::is_absolute()`.
2. ParentDir (`..`) components – detected by iterating
   `Path::components()`; covers both leading and embedded traversal.

  Apply the guard at all three call sites:
- `trust_scan::verify_multi_subject_bundle` (pre-exec scan path)
- `trust_cmd::verify_multi_subject_file` (trust verify command)
- `trust_cmd` caller loop (defense-in-depth after subject names
    already pass the inner guard)
    
Kudos to @hackerman70000 for raising this to us.

### DIFF
--- a/crates/nono-cli/src/trust_scan.rs
+++ b/crates/nono-cli/src/trust_scan.rs
@@ -766,14 +766,22 @@ fn verify_keyless_crypto(
 
 /// Resolve a bundle subject `name` to a path within `scan_root`.
 ///
-/// Rejects names that would escape the scan root before any filesystem I/O:
+/// Rejects names that would escape the scan root before any filesystem I/O,
+/// covering three distinct vectors:
+///
 /// - **Absolute paths** (`/etc/passwd`): `Path::join` discards `scan_root`
 ///   entirely for absolute inputs, so the join itself escapes the root.
 /// - **`..` traversal** (`../../etc/shadow`): parent-directory components
-///   climb above `scan_root` after resolution by the OS.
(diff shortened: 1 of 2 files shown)


crates/nono-cli/src/trust_scan.rs
```diff
@@ -760,6 +760,82 @@ fn verify_keyless_crypto(
...
+///
+/// If the target file does not yet exist (`canonicalize` on the joined path
+/// returns `Err`), the symlink check is skipped and the caller's own file I/O
+/// will produce the appropriate error.  `scan_root` itself must be resolvable;
+/// an error is returned immediately if it cannot be canonicalized.
+///
+/// # Errors
+///
+/// Returns a descriptive `String` error when the name is rejected.
+pub(crate) fn safe_subject_path(
```
cgwalters, line 789: I strongly recommend using https://docs.rs/cap-std/latest/cap_std/

PR #864, opened 2026-05-08
fix(cli): make 'nono why --host' aware of proxy domain filtering

### LINKED ISSUE #859: nono why reports Blocked , when it is not
### What happened?

With a minimal profile file:

```
{
  "meta": {
    "name": "minimal-network-blocking"
  },
  "network": {
    "allow_domain": [
      "ai-gateway.vercel.sh"
    ]
  }
}
```

You get

```
> nono why --profile minimal.json --host ai-gateway.vercel.sh
DENIED
  Reason: network_blocked
  Details: Network access is blocked. Connection to ai-gateway.vercel.sh:443 would be denied.
```

But when running `nono run -v --profile minimal.json -- curl https://ai-gateway.vercel.sh/v1/models`, it correctly goes through.

`nono why` should show that the domain is allowed.

### Steps to reproduce

Paste the above json into `minimal.json` and execute `nono why --profile minimal.json -- curl https://ai-gateway.vercel.sh/v1/models`.

### Relevant output or logs

```shell

```

### nono version

0.49.0

### Operating system

Linux

### OS version / distro

Arch

7.0.3-arch1-2

### Profile (if relevant)

```json

```

### Additional context

_No response_

SequeI:
## Linked Issue

Closes #859 

## Summary

`nono why --host` only checked whether network was globally blocked or allowed, ignoring proxy-based domain filtering entirely. Profiles with `allow_domain` set ProxyOnly mode (which is_network_blocked() treats as blocked), so `nono why` always reported DENIED even for allowed domains.

Rewrites query_network() to match on all three NetworkMode variants and resolves the profile's domain allowlist (from network_profile groups and allow_domain entries) so the query result reflects actual proxy behavior.

Domain matching mirrors HostFilter::check_host from nono::net_filter: case-insensitive exact match, wildcard subdomain matching where `*.example.com` matches subdomains but not the bare domain, and empty allowlist treated as allow-all (proxy active without domain filter).

```

Profile:
{
  "meta": {
    "name": "minimal-network-blocking"
  },
  "network": {
    "allow_domain": [
      "ai-gateway.vercel.sh"
    ]
  }
}

nono why --profile minimal.json --host ai-gateway.vercel.sh
DENIED
  Reason: network_blocked
  Details: Network access is blocked. Connection to ai-gateway.vercel.sh:443 would be denied.
  
./target/debug/nono why --profile minimal.json --host ai-gateway.vercel.sh
ALLOWED
  Reason: proxy_allowed
  Access: Connection to ai-gateway.vercel.sh:443 would be allowed via proxy
  Source: domain allowlist
```

## Test Plan

Tested manually + 3 tests

## Checklist

### DIFF
--- a/crates/nono-cli/src/why_runtime.rs
+++ b/crates/nono-cli/src/why_runtime.rs
@@ -53,7 +53,7 @@ pub(crate) fn run_why(args: WhyArgs) -> Result<()> {
                 WhyContext {
                     caps: state.to_caps()?,
                     overridden_paths: paths,
-                    allowed_domains: vec![],
+                    allowed_domains: state.allowed_domains.clone(),
                 }
             }
             None => {
--- a/crates/nono-cli/src/why_runtime.rs
+++ b/crates/nono-cli/src/why_runtime.rs
@@ -1,17 +1,60 @@
 use crate::capability_ext::CapabilitySetExt;
 use crate::cli::{SandboxArgs, WhyArgs, WhyOp};
-use crate::{policy, profile, query_ext, sandbox_state};
+use crate::{network_policy, policy, profile, query_ext, sandbox_state};
 use nono::{AccessMode, CapabilitySet, NonoError, Result};
+
+struct WhyContext {
+    caps: CapabilitySet,
+    overridden_paths: Vec<std::path::PathBuf>,
+    allowed_domains: Vec<String>,
+}
+
+/// Resolve the proxy domain allowlist from a profile's network config.
+fn resolve_allowed_domains(profile: &profile::Profile) -> Vec<String> {
+    let policy_json = crate::config::embedded::embedded_network_policy_json();
+    let net_policy = match network_policy::load_network_policy(policy_json) {
+        Ok(p) => p,
+        Err(_) => return profile.network.allow_domain.clone(),
+    };
+
+    let mut domains = Vec::new();
+
+    if let Some(net_profile_name) = profile.network.resolved_network_profile() {
+        if let Ok(resolved) = network_policy::resolve_network_profile(&net_policy, net_profile_name)
+        {
+            domains.extend(resolved.hosts);
+            for suffix in &resolved.suffixes {
+                let wildcard = if suffix.starts_with('.') {
+                    format!("*{}", suffix)
+                } else {
+                    format!("*.{}", suffix)
+                };
+                domains.push(wildcard);
+            }
+        }
+    }
+
+    domains.extend(network_policy::expand_proxy_allow(
+        &net_policy,
+        &profile.network.allow_domain,
+    ));
+
+    domains
+}
 
 pub(crate) fn run_why(args: WhyArgs) -> Result<()> {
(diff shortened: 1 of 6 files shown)


crates/nono-cli/src/why_runtime.rs
```diff
@@ -1,17 +1,60 @@
...
-    let (caps, overridden_paths): (CapabilitySet, Vec<std::path::PathBuf>) = if args.self_query {
+    let ctx: WhyContext = if args.self_query {
         match load_sandbox_state() {
             Some(state) => {
                 let paths = state.bypass_protection_as_paths();
-                (state.to_caps()?, paths)
+                WhyContext {
+                    caps: state.to_caps()?,
+                    overridden_paths: paths,
+                    allowed_domains: vec![],
```
lukehinds: I think this might end up with always empty and allow domain filtering to slip through: `allowed_domains: vec![],`

If the original sandbox was launched in `ProxyOnly` mode with domain filtering (e.g. allow_domain = ["api.example.com"]), `query_network` will call `HostFilter::allow_all()` and report every host as `proxy_allowed`, which would be wrong. I expect the sandbox state has info to reconstruct and pass in the domain list; I can take a dig into the code and find out if its not clear -  if it doesn't, this is a known limitation and we get a patch in.

crates/nono-cli/src/why_runtime.rs
```diff
@@ -84,7 +133,11 @@ pub(crate) fn run_why(args: WhyArgs) -> Result<()> {
         if prepared.needs_unlink_overrides {
             policy::apply_unlink_overrides(&mut caps);
         }
-        (caps, vec![])
+        WhyContext {
+            caps,
+            overridden_paths: vec![],
+            allowed_domains: vec![],
```
lukehinds, line 139: another possible silent passthru - the else branch (no profile, no self-query) also sets allowed_domains: vec![].

lukehinds (commented):
Thanks for digging into this - while I was looking at it, I surfacec another bug (pre your PR), you maybe able to patch in here or we can follow up:

`--allow` does not validate its arg to check it is actually a folder:

Image

This instead should be something like

```
❯ nono why --allow example.com --host evil.com
  error: 'example.com' is not a valid path
```

SequeI:
Resolved both issues, tested with a minimal profile, works on my end. Also the `--allow` issue where it does not make sure it is a folder should be fine now too

tested inside sdbx with:

```
{"meta":{"name":"test-domain-filter"},"network":{"allow_domain":["api.example.com"]}}

nono why --self --host ai-gateway.vercel.sh

  nono v0.52.0
  Capabilities:
   r+w  /home/<>/WORK/nono/target/debug (dir)
   r+w  /tmp (dir)
       + 46 system/group paths (-v to show)
   net  proxy localhost:0

  mode supervised (proxy, supervisor)
  Applying sandbox...

ALLOWED
  Reason: proxy_allowed
  Access: Connection to ai-gateway.vercel.sh:443 would be allowed via proxy
  Source: domain allowlist

```

lukehinds (approved):

PR #865, opened 2026-05-09
fix: prevent feature unification from linking libdbus in no-keyring builds

### LINKED ISSUE #861: feat: optional keyring feature flag to enable fully static musl / Chainguard static builds
## Summary

The keystore crate unconditionally links against `libdbus-1` on Linux for Secret Service support. This has two consequences:

1. **musl/Alpine builds require `-C target-feature=-crt-static`** - the musl Rust target defaults to fully static binaries, but there is no static `libdbus-1.a` in Alpine's package repository. The workaround is to force dynamic linking, which then requires `dbus-libs` in the runtime image.

2. **Chainguard static images are not possible** - `cgr.dev/chainguard/static` (the minimal, distroless runtime) cannot be used because the binary cannot be fully statically linked while depending on dbus.

## Possible approach

Add a Cargo feature flag (e.g. `keyring`, enabled by default) that gates the dbus/Secret Service keystore backend. When the feature is disabled (`--no-default-features` or `--features no-keyring`), the keystore falls back to environment variable resolution only, producing a binary with no dbus dependency.

This would allow:

```dockerfile
# fully static musl binary — no runtime dependencies at all
FROM cgr.dev/chainguard/static:latest
COPY --from=builder /nono/target/release/nono /usr/bin/nono
```

- Gate `libdbus-sys` / `keyring` crate behind `#[cfg(feature = "keyring")]` in `crates/nono/Cargo.toml`
- Fail hard when a clear error or fall through to `env://` resolution when keyring feature is disabled
- Add a `--no-default-features` build path to CI to verify the static build compiles cleanly
- Document the tradeoff: without the keyring feature, credentials must be supplied via `env://` or `op://` (1Password) and other backends, rather than the system keychains (apple / linux).

SequeI:
## Linked Issue

Closes #861 

## Summary

nono-ffi depended on nono without 'default-features = false', causing Cargo feature unification to enable system-keyring (and transitively libdbus-1) even when nono-cli was built with --no-default-features.

Also handles KeystoreAccess errors gracefully in the proxy credential loader: when the system keyring is unavailable, routes now warn and skip credential injection instead of hard-failing the entire sandbox. Removes the duplicate tracing::error!() in main.rs that printed the same error twice on stderr.

```
cargo build --release --no-default-features
ldd ./target/release/nono | grep dbus
❯  ------ (no link)

❯ ./target/release/nono run --credential openai -- echo hello

  nono v0.50.1
 
2026-05-09T00:18:39.369069Z  WARN Credential 'openai' for route 'openai' requires the system keyring, which is not available. Use env://, file://, or op:// credential references instead. Requests will proceed without credential injection.
  Applying sandbox...

hello
```

## Checklist

### DIFF
--- a/crates/nono-proxy/src/credential.rs
+++ b/crates/nono-proxy/src/credential.rs
@@ -134,6 +134,14 @@ impl CredentialStore {
                         );
                         continue;
                     }
+                    Err(nono::NonoError::KeystoreAccess(msg)) => {
+                        warn!(
+                            "Credential '{}' for route '{}' could not be loaded: {}. \
+                             Requests will proceed without credential injection.",
+                            key, normalized_prefix, msg
+                        );
+                        continue;
+                    }
                     Err(e) => return Err(ProxyError::Credential(e.to_string())),
                 };
 
@@ -207,7 +215,8 @@ impl CredentialStore {
                 let client_id =
                     match nono::keystore::load_secret_by_ref(KEYRING_SERVICE, &oauth2.client_id) {
                         Ok(s) => s,
-                        Err(nono::NonoError::SecretNotFound(msg)) => {
+                        Err(nono::NonoError::SecretNotFound(msg))
+                        | Err(nono::NonoError::KeystoreAccess(msg)) => {
                             warn!(
                                 "OAuth2 client_id not available for route '{}': {}. \
                                  Managed-credential requests on this route will be denied.",
@@ -223,7 +232,8 @@ impl CredentialStore {
                     &oauth2.client_secret,
                 ) {
                     Ok(s) => s,
-                    Err(nono::NonoError::SecretNotFound(msg)) => {
+                    Err(nono::NonoError::SecretNotFound(msg))
+                    | Err(nono::NonoError::KeystoreAccess(msg)) => {
                         warn!(
                             "OAuth2 client_secret not available for route '{}': {}. \
                              Managed-credential requests on this route will be denied.",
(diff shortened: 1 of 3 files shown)


lukehinds:
just some rebasing thanks @SequeI

SequeI:
done :) @lukehinds

crates/nono-proxy/src/credential.rs
```diff
@@ -134,6 +134,14 @@ impl CredentialStore {
                         );
                         continue;
                     }
+                    Err(nono::NonoError::KeystoreAccess(msg)) => {
```
lukehinds, line 138: “Requests will proceed without credential injection” is inaccurate here. The route still exists in RouteStore, and reverse.rs rejects managed-credential routes with no loaded credential as 503 ManagedCredentialUnavailable. We can change this warning to say requests will be denied, not forwarded without injection.

Something like...

```rust
Err(nono::NonoError::KeystoreAccess(msg)) => {
  warn!(
      "Credential '{}' not available for route '{}': {}. \
       Managed-credential requests on this route will be denied until the credential is available.",
      key, normalized_prefix, msg
  );
  continue;
}
```

Worth also doing a quick edit to amend the load() doc comment to not reference keystore access errors “hard failures,” since this PR changes that behavior for configured credentials.

https://github.com/always-further/nono/blob/917b8cba71251554dffe34c290103ef412a221e1/crates/nono-proxy/src/credential.rs#L110-L111

SequeI: Makes sense, changes made. Thanks!

lukehinds (approved):

PR #878, opened 2026-05-10
fix(pty): stop clearing terminal scrollback on exit for normal-mode sessions

### LINKED ISSUE #877: Terminal does not output `nono run ...` correctly on 0.52.0
### What happened?

```
❯ nono --version
nono 0.52.0
❯ nono run -s --allow-cwd -- date
❯ nono run -s --allow-cwd -- date
```
No output within my terminal as you can see. Also having issues with running `nono` commands and output ends up truncated etc. Or it clears the whole screen. Shows up for a split second, disappears etc etc.

For the output to be logged correctly

### Steps to reproduce

```
nono 0.52.0

nono run -s --allow-cwd -- date
```
On GNOME, terminal 

### Relevant output or logs

```shell

```

### nono version

0.52.0

### Operating system

Linux

### OS version / distro

Fedora

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

SequeI:
## Linked Issue

Closes #877

## Summary

The PTY proxy unconditionally sent CSI ?1049l (exit alternate screen buffer) on every session teardown, even when the outer terminal was never put into alternate screen mode. On VTE-based terminals (GNOME Terminal, Tilix) and some iTerm2 versions, an unsolicited alt-screen exit restores an empty saved buffer, destroying the user's scrollback history. This made it appear as though nono wiped the terminal on exit.

Introduce TERMINAL_RESTORE_NORMAL, a restore escape that resets mouse modes, bracketed paste, focus events, and keyboard enhancements but omits CSI ?1049l and CSI 2J. All teardown paths now check the child's actual alternate screen state (via the vt100 parser) before choosing the restore escape: alt-screen sessions (vim, htop) still get the full CSI ?1049l restore, while normal-mode sessions (shells, Claude Code, curl) get the safe variant that preserves scrollback.

0.52.0 output below vs this PR output as comparison.

```
❯ nono run --allow-cwd -- date

  nono v0.52.0
  Capabilities:
  
  
❯ ./target/debug/nono run --allow-cwd -- date

  nono v0.52.0
  Capabilities:
    r   /home/<>/WORK/nono (dir)
       + 47 system/group paths (-v to show)
   net  outbound allowed

  Applying sandbox...

Mon 11 May 2026 01:09:38 IST
```

## Test Plan

Would appreciate input and testing from others to verify the fix, it works for me now but extra validation is best. cc @lukehinds @quinncomendant

## Checklist

### DIFF
--- a/crates/nono-cli/src/exec_strategy.rs
+++ b/crates/nono-cli/src/exec_strategy.rs
@@ -1754,8 +1754,8 @@ fn detach_client_for_session(pty: &mut crate::pty_proxy::PtyProxy) -> bool {
     pty.detach()
 }
 
-fn restore_terminal_after_detach() {
-    crate::pty_proxy::write_detach_terminal_reset(libc::STDOUT_FILENO);
+fn restore_terminal_after_detach(in_alt_screen: bool) {
+    crate::pty_proxy::write_detach_terminal_reset(libc::STDOUT_FILENO, in_alt_screen);
     crate::pty_proxy::write_detach_notice(libc::STDERR_FILENO);
 }
 
@@ -1797,8 +1797,13 @@ fn handle_pty_detach_request(
     if in_band_detach_requested {
         info!("PTY detach requested via in-band key sequence");
     }
-    if (pause_requested || in_band_detach_requested) && pty.is_some_and(detach_client_for_session) {
-        restore_terminal_after_detach();
+    if let Some(p) = pty {
+        if pause_requested || in_band_detach_requested {
+            let in_alt_screen = p.in_alt_screen();
+            if detach_client_for_session(p) {
+                restore_terminal_after_detach(in_alt_screen);
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


lukehinds:
This may have originally been a result of trying to fix --detached mode, so its worth testing we have a happy medium between both

lukehinds:
I am seeing the same on mac as well - just testing your PR out.  This i think was introduced in f782ddcd2ae70acb663587024725de5ba4d43d0e and I missed the regression as I was overly focused on testing in a TUI

lukehinds:
Looks good! before merging, any chance we can add the two keyboard-mode resets to TERMINAL_RESTORE_NORMAL:

"\x1b[?1l", // disable application cursor keys
"\x1b>",    // normal keypad mode

Those were previously sent by the detach reset path through TERMINAL_RESTORE_AND_CLEAR_ESCAPE, so keeping them preserves existing terminal cleanup behavior without reintroducing a the alt-screen exit or screen clear issues we had before.

There is a nit issue , we should follow up for this, not block merging -a  normal child exit can still clear output from commands without a trailing newline, is worth handling as a follow-up. That needs a slightly broader teardown split between “release for parent prompt” and “release after normal child exit”, so it is better as a focused second PR

You can see this by running `nono run -s --allow-cwd -- printf cannot-see-me`

SequeI:
cc @lukehinds Added both resets, agreed other issue better in a follow up while we unblock users with this

Would be good for one last testing run, just to ensure nothing is broken

lukehinds (approved):

PR #916, opened 2026-05-13
feat(profiles): verify pack signer identities and allow use of org/repo pull format

lukehinds:
This change introduces new verification logic to ensure that installed packages were signed by the expected identity. It strengthens the security of package consumption by comparing the signer identity extracted from a package's Sigstore bundle against a pinned signer identity stored in the lockfile.

If a signer identity mismatch is detected, an error is returned, prompting the user to re-evaluate the package's provenance.

- Added `pinned_signer` to `verify_stored_bundles` to enable the comparison.
- Introduced `canonical_signer` to normalize signer URIs before comparison, focusing on repository/workflow rather than specific git refs.
- Ensured that if a profile is addressed directly by a package reference (e.g., `nono run some/pack`), that specific package is always included in the verification process, even if not explicitly listed in the profile's `packs` array.
- Made `is_registry_ref` public within the `nono-cli` crate to support the updated profile verification logic.

## Checklist

### DIFF
--- a/crates/nono-cli/src/profile/mod.rs
+++ b/crates/nono-cli/src/profile/mod.rs
@@ -1632,7 +1632,7 @@ pub fn load_profile_extends(name_or_path: &str) -> Option<Vec<String>> {
 
     // Pack-store: any installed pack that declares a profile artifact with
     // matching `install_as`.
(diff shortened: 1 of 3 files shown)


SequeI (approved):
lgtm!

PR #921, opened 2026-05-13
fix(proxy): honor explicit credential_format on custom inject headers

### LINKED ISSUE #919: Proxy strips Bearer prefix from credential injection on non-Authorization headers
### What happened?

In [crates/nono-proxy/src/credential.rs:154-160](https://github.com/always-further/nono/blob/fc5c9553b11631f8ec9157b43c3a032f1cc946a6/crates/nono-proxy/src/credential.rs#L149-L160), there's an auto-override that strips the Bearer prefix from any header that isn't Authorization.

Any custom_credentials route that uses a non-Authorization header with `credential_format: "Bearer {}"` silently injects the bare `token` instead of `Bearer <token>`. 

Do not change credential_format if I a custom is specified.

### Steps to reproduce

 Affected config pattern:
 
```json
"litellm": {
    "upstream": "https://litellm",
    "credential_key": "env://LITELLM_TOKEN",
    "inject_header": "x-litellm-api-key",
  }
```

Upstream returns 500 because the credential value is malformed. Hard to debug because the proxy logs show the request succeeding (allowed, forwarded), the credential format mismatch is invisible.

### Relevant output or logs

```shell

```

### nono version

0.53.0

### Operating system

macOS

### OS version / distro

macOS

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

SequeI:
## Linked Issue

Closes #919 

## Summary

Treat credential_format as optional (Option<String>) so omitted vs explicit 'Bearer {}' are distinct. Resolve defaults only when absent: Authorization -> Bearer {}, other headers -> bare {}. Manifest export and profile validation use the same resolution.

cc @panga lmk if this works now :)

## Checklist


panga (approved):
Perfect.

lukehinds (approved):

PR #937, opened 2026-05-15
docs(cli-security): add isolation scope and deployment model

lukehinds:
This commit adds a new "Isolation Scope and Deployment Model" section to the security model documentation.

The new content clarifies nono's role as a fine-grained, kernel-enforced capability control mechanism. It distinguishes nono from monolithic isolation stacks like containers or microVMs, detailing:

- What nono is (a syscall-level sandbox) and what it is not (a hypervisor or container runtime).
- How nono's effectiveness is proportional to the accuracy of the policy applied to the environment it runs in.
- How nono can be combined with hardware isolation (containers or microVMs) for the strongest possible guarantees.
- A comparison table illustrating different deployment models and their respective perimeters and granularities.
- Recommended postures for high-assurance deployments versus typical development, CI, and local agent use cases.

@SequeI let me know how this reads pls

### DIFF
--- a/docs/cli/internals/security-model.mdx
+++ b/docs/cli/internals/security-model.mdx
@@ -418,7 +418,7 @@ Supervised mode on macOS provides rollback snapshots (content-addressable filesy
 
 ## Isolation Scope and Deployment Model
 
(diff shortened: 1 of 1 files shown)


SequeI (approved):
lgtm, this read nice and concise!

PR #944, opened 2026-05-15
fix(pty): prevent keyboard mode reset in multiplexers

### LINKED ISSUE #941: nono running pi inside tmux breaks ESC hotkey
### What happened?

When running pi agent inside tmux with nono, the escape key inside pi stops working. When running outside tmux, everything is working as expected

I do run tmux with the following settings:
```
set-option -g allow-passthrough on
set-option -g extended-keys on
set -s extended-keys-format csi-u
```

Escape key should be forwarded correctly to pi.

### Steps to reproduce

1. Create profile called `pi` (JSON in the profile section)
2. Start tmux
3. Run `nono run --profile pi -- pi`
4. Hit escape key twice and see that nothing happens (it should open the tree picker)

### nono version

nono 0.54.0

### Operating system

macOS

### OS version / distro

macOS 26.3

### Profile (if relevant)

```json
{
  "extends": "default",
  "interactive": true,
  "meta": {
    "name": "pi",
  },
  "filesystem": {
    "allow": [
      "$HOME/.pi",
      "$HOME/.agents",
      "$TMPDIR",
    ],
    "read": [
      "$HOME/.cargo",
      "$HOME/go",
      "$HOME/.local/share/mise/installs"
    ],
  },
  "network": {
    "block": false
  },
    "(allow user-preference-read)"
  ],
  "groups": {
    "include": [
      "user_caches_macos",
      "node_runtime",
      "unlink_protection"
    ]
  }
}
```

### Additional context

I noticed that it stops working starting with version `0.52.0` of nono. 

I am running tmux `3.6a` and pi `0.74.0`.

lukehinds:
Previously, the `TERMINAL_RESTORE_NORMAL` sequence included `DECCKM off` (`\x1b[?1l`) and `DECPNM` (`\x1b>`), which disable application cursor keys and reset the keypad to normal mode.

Emitting these sequences directly to the outer PTY when running inside a terminal multiplexer (such as tmux, GNU screen, or Zellij) corrupts the multiplexer's internal keyboard-mode tracking. This issue leads to broken extended-key delivery for the running application (e.g., `ESC` via `CSI-u`).

To prevent this interference:

- The `DECCKM off` and `DECPNM` sequences have been moved into a new constant, `KEYBOARD_MODE_RESET`.
- A new `is_inside_multiplexer` function detects common terminal multiplexers by checking for environment variables like `TMUX`, `STY`, and `ZELLIJ`.
- The `KEYBOARD_MODE_RESET` sequences are now only emitted if the application is not in an alternative screen and is *not* detected as running inside a terminal multiplexer.

New unit tests confirm the correct separation and content of these terminal control sequence constants.

Fixes: #941

### DIFF
--- a/crates/nono-cli/src/pty_proxy.rs
+++ b/crates/nono-cli/src/pty_proxy.rs
@@ -102,8 +102,6 @@ const TERMINAL_RESTORE_AND_CLEAR_ESCAPE: &[u8] = concat!(
     "\x1b[?1005l\x1b[?1006l\x1b[?1015l", // disable mouse encodings
     "\x1b[?1004l", // disable focus events
     "\x1b[?2004l", // disable bracketed paste
-    "\x1b[?1l", // disable application cursor keys
-    "\x1b>",   // normal keypad mode
     "\x1b[?1049l", // exit alternate screen
     "\x1b[?25h", // show cursor
     "\x1b[2J\x1b[H", // clear screen + cursor home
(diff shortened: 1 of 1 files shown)


SequeI (approved):
LGTM! Thanks for cleaning up the `TERMINAL_RESTORE_NORMAL`  issue after me 😄

PR #948, opened 2026-05-16
fix(pty): forward bare ESC immediately in filter_client_input

### LINKED ISSUE #941: nono running pi inside tmux breaks ESC hotkey

caiocdcs:
## Linked Issue

Closes #941 

## Summary

When ESC is pressed alone inside tmux with `extended-keys-format csi-u`, `filter_client_input` buffers the bare `\x1b` byte waiting to see if a CSI-u detach sequence follows. It gets held until the next keystroke — at that point `\x1b` + the next byte arrive together, which the terminal interprets as Alt+key instead of plain ESC. This broke ESC in TUI apps like pi and opencode, introduced as a regression in v0.52.0.

To fix this:
- `filter_client_input` now uses `enumerate()` to look ahead in the current read batch.
- `\x1b` is only buffered for enhanced detach matching when `[` immediately follows in the same batch.
- A bare ESC with no `[` following is forwarded immediately.

I used AI to help me investigate the issue and reviewing the changes, they are plausible.

## Test Plan

Two regression tests added:
- `filter_client_input_forwards_bare_esc_immediately` — bare ESC is forwarded right away
- `filter_client_input_forwards_esc_not_paired_with_next_key` — ESC + non-`[` byte are both forwarded as-is

All existing `filter_client_input` tests pass, including chunked and full CSI-u detach sequences.

Also, tested with opencode and pi on ghostty/wezterm and it seems to be working.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes
+
+- *(pty)* Forward bare ESC immediately instead of buffering for CSI-u detach match, fixing ESC key in TUI apps inside tmux with `extended-keys-format csi-u` (#941)
+
 ### Notes
 
 - Socket grant state now records explicit socket scope. New subtree socket
--- a/crates/nono-cli/src/pty_proxy.rs
+++ b/crates/nono-cli/src/pty_proxy.rs
@@ -1035,7 +1035,7 @@ impl PtyProxy {
     }
     fn filter_client_input(&mut self, bytes: &[u8]) -> Vec<u8> {
         let mut forwarded = Vec::with_capacity(bytes.len());
-        for &byte in bytes {
+        for (i, &byte) in bytes.iter().enumerate() {
             if self.maybe_consume_enhanced_detach_byte(byte, &mut forwarded) {
                 continue;
             }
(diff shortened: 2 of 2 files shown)


lukehinds (approved):
Tested out well ! thanks for the PR

PR #977, opened 2026-05-20
feat(keyring): add NONO_KEYRING_TIMEOUT_SECS for keychain access

### LINKED ISSUE #967: Do not timeout as quickly when accessing keychain

I am using custom_credentials, which are loaded from the system keychain.
However when starting nono this means unlocking the keychain.
This takes some time for me, during which nono has always already timed out.

Previously it would just abort and I would start it again, at which time the keychain was still unlocked and I could allow the permissions.
Now however, the issue has been aggravated a bit, because nono simply starts without setting up the custom_credentials and this fact can be overlooked.

For me personally it would be good if nono could just wait longer, but I am not sure about other use-cases.
Maybe it is possible to notice, when the keychain is still being unlocked, and wait only in that case.

_No response_

### How is this blocking you?

### Additional context

Using keepassxc as my linux system keychain here.

caiocdcs:
Closes #967

## Summary

Adds `NONO_KEYRING_TIMEOUT_SECS` to bound how long nono will block on the
system keyring (macOS Keychain / Linux Secret Service / KeePassXC). Default
is 120 seconds -- long enough for a real keychain-unlock workflow, finite
enough that an unattended nono cannot hang forever. `0` preserves the old
"wait forever" behaviour for users who want it.

Also surfaces credential-load failures on the proxy path with proper
redaction: previously a locked keychain caused the proxy to start silently
without the credential, which the issue author flagged as easy to overlook.

Key design decisions:

1. **Default 120 s** -- covers a real human unlock workflow (notice prompt,
   switch windows, type a master password, click approve). Documented in
   the source with rationale.
2. **`0` means "wait forever"** -- explicit opt-in to the old behaviour.
3. **Invalid env value falls back to default** with a warning -- a typo
   here should not block startup.
4. **No thread overhead when not needed** -- when `keyring_timeout()` returns
   `None`, `call_with_keyring_timeout` calls the closure inline.
5. **Detached thread on timeout** -- documented. Rust cannot cancel a
   blocking syscall; the spawned thread stays alive until the underlying
   call returns or the process exits. No secret material leaks because
   the `Zeroizing<String>` result is dropped on the thread side once the
   channel receiver is gone.
6. **Proxy warnings use existing redaction helpers** -- credential refs are
   never echoed raw in warning output.

## Changes

- `crates/nono/src/keystore.rs`: `KEYRING_TIMEOUT_SECS = 120`,
  `keyring_timeout()` env reader, `call_with_keyring_timeout()` wrapper.
  Applied at `load_single_secret` and `load_from_keyring_uri`, gated on
  `feature = "system-keyring"`.
- `crates/nono-proxy/src/credential.rs`: `redact_credential_ref()` helper
  covering `op://`, `apple-password://`, `keyring://`, `bw://`, and
  `file://` URI schemes. Warning messages on all three skip branches in
  `CredentialStore::load` (static `KeystoreAccess`, OAuth2 `client_id`,
  OAuth2 `client_secret`) now include the timeout env var hint.

## Test Plan

- Unit tests in `keystore.rs` (`keyring_timeout_tests` submodule) covering
  env-var parsing (unset / 0 / valid value / invalid), the inline path when
  timeout is `None`, the fast-success path, and the timeout-fires path.
  All env-var tests serialise via an in-module `ENV_LOCK` + `EnvGuard`.
- Manual verification on Linux VM with KeePassXC: locked-DB timeout fires,
  unlock-in-time succeeds, `0` waits forever, invalid value falls back to
  default.
- `make ci`.

_AI-assisted development. All code reviewed and tested by the author._

### DIFF
--- a/crates/nono-proxy/src/credential.rs
+++ b/crates/nono-proxy/src/credential.rs
@@ -139,13 +139,8 @@ impl CredentialStore {
                         let redacted = redact_credential_ref(key);
                         warn!(
                             "Credential '{}' not available for route '{}': {}. \
-                             Managed-credential requests on this route will be denied until the credential is available.",
-                            redacted, normalized_prefix, msg
-                        );
-                        eprintln!(
-                            "[nono] Credential '{}' not available for route '{}': {}\n\
-                             \x20      Managed-credential requests on this route will be denied.\n\
-                             \x20      Set NONO_KEYRING_TIMEOUT_SECS=N (default 120) to wait longer for keychain unlock; 0 disables the timeout.",
+                             Managed-credential requests on this route will be denied until the credential is available. \
+                             Set NONO_KEYRING_TIMEOUT_SECS=N (default 120) to wait longer for keychain unlock; 0 disables the timeout.",
                             redacted, normalized_prefix, msg
                         );
                         continue;
@@ -222,13 +217,8 @@ impl CredentialStore {
                         let redacted = redact_credential_ref(&oauth2.client_id);
                         warn!(
                             "OAuth2 client_id '{}' not available for route '{}': {}. \
-                                 Managed-credential requests on this route will be denied.",
-                            redacted, route.prefix, msg
-                        );
-                        eprintln!(
-                            "[nono] OAuth2 client_id '{}' not available for route '{}': {}\n\
-                                 \x20      Managed-credential requests on this route will be denied.\n\
-                                 \x20      Set NONO_KEYRING_TIMEOUT_SECS=N (default 120) to wait longer for keychain unlock; 0 disables the timeout.",
+                                 Managed-credential requests on this route will be denied. \
+                                 Set NONO_KEYRING_TIMEOUT_SECS=N (default 120) to wait longer for keychain unlock; 0 disables the timeout.",
                             redacted, route.prefix, msg
                         );
                         continue;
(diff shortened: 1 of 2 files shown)


crates/nono-proxy/src/credential.rs
```diff
@@ -136,10 +136,17 @@ impl CredentialStore {
...
                             "Credential '{}' not available for route '{}': {}. \
                              Managed-credential requests on this route will be denied until the credential is available.",
-                            key, normalized_prefix, msg
+                            redacted, normalized_prefix, msg
+                        );
+                        eprintln!(
+                            "[nono] Credential '{}' not available for route '{}': {}\n\
+                             \x20      Managed-credential requests on this route will be denied.\n\
+                             \x20      Set NONO_KEYRING_TIMEOUT_SECS=N (default 120) to wait longer for keychain unlock; 0 disables the timeout.",
+                            redacted, normalized_prefix, msg
```
lukehinds: Please use the NonoError handlers - , the nono-proxy crate is a library used by the language SDKs alongside the nono-CLI - so we tend to handle  - you should see lots of examples in credential.rs:

For example

```rust
match nono::keystore::load_secret_by_ref(KEYRING_SERVICE, &oauth2.client_id) {
      Ok(s) => s,
      Err(nono::NonoError::SecretNotFound(msg))
      | Err(nono::NonoError::KeystoreAccess(msg)) => {
          warn!(
              "OAuth2 client_id not available for route '{}': {}. \
               Managed-credential requests on this route will be denied.",
              route.prefix, msg
          );
          continue;
      }
      Err(e) => return Err(ProxyError::Credential(e.to_string())),
  }
```

The rest looks good and thank you for contributing! great having you on the nono bus!

lukehinds (commented):
Just a change to use inbuilt error handling, otherwise looking good!

lukehinds:
Is this meant to be in draft still @caiocdcs , or ready for review?

caiocdcs:
This is still not ready @lukehinds. I added some comments in the original issue, but one upstream dependency is still timing out and ignoring any configurations. I'm still investigating.

caiocdcs:
After the work done on #1049 this was change now works as the async zbus allows us to put some timeout in the task. So, now this is ready to review @lukehinds.

lukehinds (approved):
lgtm,  The secret-lifetime reasoning is sound, it's fail-secure, and the redaction is actually strictly better than before, so kudos there. nice PR and well thought out!

PR #992, opened 2026-05-22
chore(deps): update sigstore crates to 0.8.0

lukehinds:
Upgrade `sigstore-*` dependencies to version `0.8.0` in `nono-cli` and `nono`. This also updates `Cargo.lock` with numerous new and updated transitive dependencies, including `reqwest`, so may need to be mindful of this when released. 

I also had to tweak  minor `Cow` dereferencing adjustments in `src/scrub.rs` from the bump in sigstore.


SequeI (approved):
lgtm!

Issue #658, opened 2026-04-13
GPU sandbox: sweep remaining --allow-gpu gaps (sysfs, multi-GPU, IPC, diagnostics)


scp7:
Follow-up to #651. That PR makes `--allow-gpu` functional for single-GPU CUDA init on Linux (adds `/dev/nvidia-uvm-tools` to the device filter, grants `/proc/self` and `/proc/driver`). Several gaps remain that are out of scope for #651 because each has its own design/security tradeoff or requires hardware we don't currently test on. Tracking here so they don't get lost.

**Verified baseline**: after #651, `nono run --allow-gpu` works for PyTorch CUDA on Ubuntu 22.04 / kernel 6.8 / driver 570 / A100-SXM4-40GB with no additional `--allow` flags.

## Gaps to sweep

1. **Sysfs for topology / NUMA**
   - `/sys/module/nvidia` — driver module introspection
   - `/sys/bus/pci/devices` — PCI topology (broad; needs design — maybe narrow to `/sys/bus/pci/drivers/nvidia`)
   - `/sys/devices/system/node` — NUMA topology for multi-socket training

2. **Multi-GPU device nodes**
   - `nvidia-nvswitch*`, `nvidia-nvlink` — NVLink fabric for 8×A100/H100 boxes
   - `nvidia-vgpu*` — vGPU workloads
   - Extend `is_nvidia_compute_device` predicate + unit tests.
   - Different threat surface than single-GPU (peer-to-peer DMA) — warrants its own review.

3. **IPC surface**
   - `/dev/shm` — CUDA IPC, NCCL SHM transport
   - `/dev/infiniband/*` — NCCL RDMA for multi-node training
   - `nvidia-persistenced` socket, MPS control socket
   - Real security implications; should not be granted unconditionally.

4. **ROCm sysfs**
   - `/sys/class/kfd` (the `/dev/kfd` device is already granted).
   - Blocked on having an AMD GPU to test.

5. **Diagnostics / UX**
   - Preflight: warn if the user is not in the `video` / `render` group.
   - Structured logging of which devices were granted (currently only a count).

6. **Caches (profile-layer policy, not `--allow-gpu`)**
   - `~/.nv/ComputeCache` (CUDA JIT cache), `~/.cache/torch`, `~/.triton/cache`.
   - Arguably belongs in a profile (`claude-code-gpu` / `pytorch`), not in the device-allowlist code path.

[...]

PR #651, opened 2026-04-12
fix(gpu): add nvidia-uvm-tools to GPU device allowlist

Kexin-xu-01:
## Summary

Follow-up to #639, fixing an issue found in real-world testing on NVIDIA driver 570 / CUDA 12.8 (A100, native Linux, Landlock V4).

### `nvidia-uvm-tools` not included in GPU device allowlist

The filter in `maybe_enable_gpu()` matches `nvidia-uvm` explicitly but not `nvidia-uvm-tools` — the non-digit suffix causes it to fall through all three branches and never be granted.

NVIDIA driver 570 opens `/dev/nvidia-uvm-tools` (minor 1, alongside `nvidia-uvm` minor 0) during UVM initialization. When `open()` returns `EACCES` from Landlock, the driver treats it as a fatal `cudaErrorOperatingSystem` (Error 304) — even when `/dev/nvidia-uvm` itself is accessible.

```rust
// Before
n == "nvidiactl"
    || n == "nvidia-uvm"
    || (n.starts_with("nvidia") && n[6..].bytes().all(|b| b.is_ascii_digit()) && n.len() > 6)

// After
n == "nvidiactl"
    || n == "nvidia-uvm"
    || n == "nvidia-uvm-tools"   // driver 570+ opens this during UVM init
    || (n.starts_with("nvidia") && n[6..].bytes().all(|b| b.is_ascii_digit()) && n.len() > 6)
```

## Test plan

- [x] `torch.cuda.get_device_capability()` succeeds inside sandbox (previously Error 304)
- [x] `torch.cuda.device_count()` returns correct count

Tested on: Ubuntu 22.04, Landlock V4, NVIDIA A100-SXM4-40GB, driver 570, CUDA 12.8

PR #993, opened 2026-05-22
fix(sandbox): widen /proc/self/task for grandchild CUDA init

### LINKED ISSUE #924: Fix: CUDA blocked writing to /proc/self/task/<tid>/comm for thread naming
## Problem

CUDA (tested with CUDA 580) calls `cuInit()` which creates a thread and attempts to name it by writing to `/proc/self/task/<tid>/comm`. The nono sandbox's Landlock rules resolve `/proc/self` to the **startup PID** of the nono process — so when a Python subprocess with a different PID tries to write to its own `/proc/<pid>/task/<tid>/comm`, the write is blocked with `EACCES`.

This causes `cuInit` to return error 304, preventing GPU access entirely.

## Root Cause

- Landlock rule for `/proc/self` is locked to the nono startup PID (e.g. `/proc/81199/task`)
- Child processes spawned by the sandbox have different PIDs
- Their writes to `/proc/<pid>/task/*/comm` are blocked because the allow rule doesn't cover them

The profile only had `/proc/self` in the `read` list, not in `allow` (read+write), and only for the startup PID.

`/proc/self/task` alone is insufficient because nono resolves the symlink to the startup PID — child processes need access under their own numeric PID path.
It is possible to add an allow-list with dynamic path?

I have tried enabling write access to /proc but this is dangerous.

### How is this blocking you?

None

### Additional context

  OS
  - Ubuntu 22.04.5 LTS (Jammy), kernel 6.8.0-1053-gcp
  
  Hardware
  - 12 vCPUs
  - 83 GiB RAM (54 GiB free), no swap
  - 485 GB root disk (442 GB available)
  - 1× NVIDIA A100-SXM4-40GB (40 GB HBM2)
  
  GPU / CUDA
  - Driver: 580.126.20
  - CUDA: 13.0 (runtime), CUDA toolkit 12.9 installed at /usr/local/cuda-12.9

  Python
  - Python 3.10.12 (system)

  Services
  - JupyterHub (/opt/jupyterhub)
  - containerd

scp7:
## Linked Issue

Closes #924

## Summary

`--allow-gpu` grants `/proc/self/task` as ReadWrite on NVIDIA systems so the CUDA driver can write `/proc/self/task/<tid>/comm` during `cuInit()`. In Monitor mode that rule is canonicalised against the direct child's PID inode before Landlock is installed, so any grandchild process (e.g. `claude → python → cuInit()`) writing to its own `/proc/<gpid>/task/<tid>/comm` hits `EACCES` and surfaces as CUDA Error 304 (`cudaErrorOperatingSystem`).

The read-side widening in `widen_procfs_self_to_proc()` already covers this case for `/proc/self`. This PR extends the same widening to `/proc/self/task`, preserving the original ReadWrite access mode.

### Trade-off

Landlock cannot express "any `/proc/<pid>/task`" without covering the `/proc` parent — it is a filesystem-tree allowlist, not a glob. So the widened rule necessarily covers other writable per-process knobs (`oom_score_adj`, `coredump_filter`, …). The kernel's own per-file ownership check remains the effective limit on which writes succeed: a process can only write to its own `/proc/<pid>/*` writable entries. The trade-off and the security reasoning are documented inline on `widen_procfs_self_to_proc()`.

Related history:
- #651 / #673 added the original `--allow-gpu` NVIDIA procfs grants but only validated the direct-child case.
- #658 tracks the broader `--allow-gpu` sweep; this fix is scoped narrowly to the write-widening gap and intentionally kept separate.

## Test Plan

**Unit tests** (new, in `crates/nono/src/capability.rs::procfs_remap_tests`):
- `widen_procfs_self_to_proc_widens_read_rule` — existing `/proc/self` Read widening, now under explicit coverage.
- `widen_procfs_self_to_proc_widens_proc_self_task_write_rule` — regression for #924: the RW grant widens to `/proc` and keeps `AccessMode::ReadWrite`.
- `widen_procfs_self_to_proc_merges_after_dedup_on_linux` — both widenings + dedup leave no stale PID-pinned rule.
- `widen_procfs_self_to_proc_leaves_proc_self_fd_alone` — defence-in-depth: `/proc/self/fd` is per-PID by design and must not be widened.

**Integration test** (added to `tests/integration/test_gpu.sh`): `grandchild /proc/self/task/<tid>/comm write allowed with --allow-gpu (#924)` runs the comm-write under `sh -c "python3 -c …"` so the writer PID differs from the direct child.

**Manual verification on Ubuntu 22.04 / NVIDIA A100-SXM4-40GB / driver 580.126.20 / CUDA 13** — matches the reporter's environment:

1. Minimal grandchild comm write — direct child `sh` pid `435468`, grandchild `python3` pid `435470`, write to `/proc/self/task/435470/comm` succeeded.
2. Real `cuInit()` via `libcuda.so.1` from a grandchild — `OK cuInit succeeded; 1 CUDA device(s) visible`. Pre-fix the same call returned 304.
3. Defence-in-depth — `/proc/driver` parent still denied even with `--allow-gpu`; grandchild comm write still denied without `--allow-gpu`.

Local CI:
- `cargo fmt --check` ✅
- `cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::unwrap_used` ✅
- `cargo test` ✅

## Checklist

- [x] Public-facing changes are paired with documentation updates (function doc on `widen_procfs_self_to_proc()` rewritten; no user-facing CLI/flag changes)
- [ ] Release note has been added to `CHANGELOG.md` if needed — happy to add one on request; the user-visible behaviour is "CUDA init now works for grandchild processes under `--allow-gpu`", which is arguably a bug fix that does not need a separate note.

## Agent Disclosure

This PR was prepared by an AI coding agent (Claude) operating under the maintainer's direction (@scp7). The fix follows the implementation sketch the maintainer posted in #924's comment thread: extending `widen_procfs_self_to_proc()` to widen `/proc/self/task` alongside `/proc/self`.

Files / sections consulted:
- `crates/nono/src/capability.rs` — `widen_procfs_self_to_proc()`, `remap_procfs_self_references()`, `AccessMode`, `deduplicate()`.
- `crates/nono-cli/src/sandbox_prepare.rs` — `grant_nvidia_gpu_procfs()`, `maybe_enable_gpu()` (no changes; reviewed for grant shape).
- `crates/nono-cli/src/exec_strategy.rs` — Monitor-mode call sequence (`remap_procfs_self_references` then `widen_procfs_self_to_proc`).
- `crates/nono/src/sandbox/linux.rs` — `access_to_landlock()` to understand the AccessFs flags `Write` expands into.
- `proj/invariants.yaml` — `least-privilege-scope`, `separate-read-write`, `landlock-strict-allowlist`, `test-new-capability-types`, `test-on-both-platforms`.
- `CLAUDE.md` and `.github/pull_request_template.md`.

## Agent Compliance Check

- [x] An issue already exists (#924)
- [x] I disclosed that I am an agent in this PR (the issue discussion already contains the maintainer's implementation sketch)
- [x] I described my intent and approach in the issue discussion (see #924 maintainer comment)
- [x] I provided required attribution for reused or adapted code (#651, #673, #658 cross-references)
- [x] I did not use forbidden patterns such as unwrap/expect (test-only `.expect()` only, gated by clippy cfg)
- [x] I used `NonoError` where required (no new fallible paths added)
- [x] I validated and canonicalized all relevant paths (widening preserves canonical `/proc` resolution; access mode preserved)

### DIFF
--- a/crates/nono/src/capability.rs
+++ b/crates/nono/src/capability.rs
@@ -1325,28 +1325,58 @@ impl CapabilitySet {
         self.deduplicate();
     }
 
-    /// Widen `/proc/<pid>` READ-only Landlock rules to `/proc` so that
-    /// grandchild processes can access their own procfs entries.
+    /// Widen `/proc/<pid>` Landlock rules to `/proc` so that grandchild
+    /// processes can access their own procfs entries.
     ///
-    /// This is needed because Landlock rules are fixed at sandbox setup time with
-    /// the direct child's PID. When a grandchild (e.g. nono→sh→bun) forks, it
-    /// gets a new PID and its `/proc/self` resolves to a different inode than the
-    /// direct child's `/proc/<sh_pid>`. By widening to `/proc`, we allow any
-    /// descendant to read its own procfs entries.
+    /// This is needed because Landlock rules are fixed at sandbox setup time
+    /// with the direct child's PID. When a grandchild (e.g. nono→sh→bun) forks,
+    /// it gets a new PID and its `/proc/self` resolves to a different inode
+    /// than the direct child's `/proc/<sh_pid>`. By widening to `/proc`, we
+    /// allow any descendant to access its own procfs entries.
     ///
-    /// Only applies to READ capabilities at the `/proc/self` level (not
-    /// subdirectories like `/proc/self/fd` which may have write access).
+    /// Two narrow rewrites are performed:
+    ///
+    /// 1. **`/proc/self` (Read)** → `/proc` (Read). Reads under `/proc/<pid>`
+    ///    are essentially free to widen: the kernel already permits cross-PID
+    ///    reads on most procfs entries (subject to `hidepid` and
+    ///    `ptrace_scope`), so coarsening the Landlock rule does not hand out
+    ///    anything the kernel would not already allow.
+    ///
+    /// 2. **`/proc/self/task` (ReadWrite)** → `/proc` (ReadWrite). This rule
+    ///    is granted by `--allow-gpu` on NVIDIA systems so the CUDA driver
+    ///    can write `/proc/self/task/<tid>/comm` (the thread name) during
+    ///    `cuInit()`. Without widening, only the direct child can perform
+    ///    that write; a grandchild's `/proc/self/task/<tid>/comm` resolves
+    ///    to a different inode and gets `EACCES`, surfacing as CUDA Error 304
+    ///    (`cudaErrorOperatingSystem`). See issue #924.
+    ///
+    ///    Trade-off: widening to `/proc` necessarily covers other writable
+    ///    per-process knobs under `/proc/<pid>/` (e.g. `oom_score_adj`,
+    ///    `coredump_filter`). Two factors bound the impact:
+    ///
+    ///    - Landlock cannot express "any `/proc/<pid>/task`" without covering
+    ///      the parent — it is a filesystem-tree allowlist, not a glob.
+    ///    - The kernel still enforces per-file ownership: a process can only
+    ///      write to its own `/proc/<pid>/*` writable knobs, not another
+    ///      process's. So the widened Landlock rule defers to kernel
+    ///      ownership checks rather than handing out cross-process write.
+    ///
+    /// Other capabilities (e.g. `/proc/self/fd` writes, profile-specific
+    /// grants) are unaffected.
     pub fn widen_procfs_self_to_proc(&mut self) {
         for cap in &mut self.fs {
-            if cap.access == AccessMode::Read {
-                let is_proc_self_dir = cap
-                    .original
-                    .to_str()
-                    .map(|s| s == "/proc/self" || s == "/proc/self/")
-                    .unwrap_or(false);
-                if is_proc_self_dir {
-                    cap.resolved = std::path::PathBuf::from("/proc");
-                }
+            let original_str = cap.original.to_str().unwrap_or("");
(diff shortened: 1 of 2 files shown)


scp7:
Closing this PR unmerged. The fix as written widens the Landlock allowlist to `/proc` (RW), not just `/proc/self/task` (RW), because Landlock is filesystem-tree-based and cannot express "any `/proc/<pid>/task/<tid>/comm`". The kernel's per-file ownership check bounds the effective impact, but at the Landlock layer this expands write access to every writable per-process procfs entry (`oom_score_adj`, `coredump_filter`, `clear_refs`, `attr/*`, etc.) — that's a sandbox privilege expansion and shouldn't land via `--allow-gpu`.

The right primitive for the precise pattern (`^/proc/\d+/task/\d+/comm$` with writer-TGID/TID cross-check) is **seccomp-notify in Supervised mode**, which nono already has infrastructure for. Plan:

- Drop the `/proc/self/task` Landlock RW grant entirely.
- Add a seccomp filter that notifies on `openat()` to `/proc/*` with write modes.
- Supervisor inspects the path, validates `<pid>` == calling thread's TGID and `<tid>` ∈ caller's TIDs, only then allows.
- Auto-enable Supervised mode when `--allow-gpu` is requested on NVIDIA so users don't have to know.

A new issue will track the precise implementation. #924 stays open until that ships.

Refs: #924, #651, #673, #658.

PR #995, opened 2026-05-22
feat(policy): add java_runtime group and java-dev profile

### LINKED ISSUE #994: Java support

When running Java-based agents or tools inside a nono sandbox, the JVM cannot locate its runtime because user-space version managers (SDKMAN) and build tool caches (Maven, Gradle) are not included in any default  policy group. Users must manually allowlist paths like ~/.sdkman, ~/.m2/repository, and ~/.gradle/caches every time.

Additionally, on macOS with Apple Silicon, the JVM fails to start because Seatbelt denies the sysctl-write call to kern.grade_cputype, which the JVM uses to query CPU translation state (Rosetta 2).

A java_runtime policy group and a java-dev profile, following the existing pattern of python_runtime/python-dev, node_runtime/node-dev, etc.

Manually adding --allow-read ~/.sdkman --allow-read ~/.m2/repository etc. to every invocation and the unstrusted macos seatbelt rule

### How is this blocking you?

### Additional context

_No response_

panga:
# Linked Issue

Closes #994

# Summary

Add a java_runtime policy group and java-dev profile to support Java development inside nono sandboxes.

java_runtime group grants read access to:

- JVM paths: /usr/lib/jvm (Linux)
- SDKMAN: ~/.sdkman
- Maven (safe subdirs): ~/.m2/repository, ~/.m2/wrapper
- Gradle (safe subdirs): ~/.gradle/caches, ~/.gradle/wrapper

Maven/Gradle credential files (settings.xml, gradle.properties) remain inaccessible by design — narrow subdirectory grants avoid the Landlock deny-within-allow limitation.

java-dev profile extends default with java_runtime, network access, and readwrite workdir — matching python-dev, node-dev, go-dev, rust-dev.

Seatbelt sysctl rule: unconditionally allows `(allow sysctl-write (sysctl-name "kern.grade_cputype"))` so the JVM starts correctly on Apple Silicon (required to query CPU translation state).

# Test Plan

- make build — confirms policy.json parses at compile-time embedding
- make clippy — no warnings
- cargo test -p nono -- sandbox::macos::tests — all 57 Seatbelt profile generation tests pass
- cargo test -p nono-cli -- test_setup_profiles_loadable_by_name — confirms java-dev profile loads

# Checklist

[x] - An issue exists and is linked above
[x] - All commits are signed-off, using [DCO](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin)
[x] - All new code follows the project's coding standards and is covered by tests
[] - Public-facing changes are paired with documentation updates
[] - Release note has been added to CHANGELOG.md if needed

### DIFF
--- a/crates/nono/src/sandbox/macos.rs
+++ b/crates/nono/src/sandbox/macos.rs
@@ -521,8 +521,6 @@ fn generate_profile(caps: &CapabilitySet) -> Result<String> {
 
     // Allow specific system operations
     profile.push_str("(allow sysctl-read)\n");
-    // JVM on Apple Silicon requires this sysctl write to query CPU translation state
-    profile.push_str("(allow sysctl-write (sysctl-name \"kern.grade_cputype\"))\n");
 
     // Mach IPC: allow service resolution. Deny Keychain/security services by default.
     // If a keychain DB is explicitly granted, skip these denies so profiles that
--- a/crates/nono/src/sandbox/macos.rs
+++ b/crates/nono/src/sandbox/macos.rs
@@ -521,6 +521,8 @@ fn generate_profile(caps: &CapabilitySet) -> Result<String> {
 
     // Allow specific system operations
     profile.push_str("(allow sysctl-read)\n");
+    // JVM on Apple Silicon requires this sysctl write to query CPU translation state
+    profile.push_str("(allow sysctl-write (sysctl-name \"kern.grade_cputype\"))\n");
 
     // Mach IPC: allow service resolution. Deny Keychain/security services by default.
     // If a keychain DB is explicitly granted, skip these denies so profiles that
--- a/crates/nono-cli/data/policy.json
+++ b/crates/nono-cli/data/policy.json
@@ -413,6 +413,19 @@
         ]
       }
     },
+    "java_runtime": {
+      "description": "Java runtime and build tool paths (SDKMAN, Maven, Gradle)",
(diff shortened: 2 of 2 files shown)


crates/nono/src/sandbox/macos.rs
```diff
@@ -521,6 +521,8 @@ fn generate_profile(caps: &CapabilitySet) -> Result<String> {
 
     // Allow specific system operations
     profile.push_str("(allow sysctl-read)\n");
+    // JVM on Apple Silicon requires this sysctl write to query CPU translation state
+    profile.push_str("(allow sysctl-write (sysctl-name \"kern.grade_cputype\"))\n");
```
lukehinds: I'd prefer this to be set into platform_rules instead

If it sits in the nono crate it becomes universal sandbox primitive on every macOS sandbox (including the language SDKs., which cuts against the library/CLI boundary

What you can do is add a raw string to a profile:

```
    "(allow sysctl-write (sysctl-name \"kern.grade_cputype\"))"
  ]
```

panga: This string was hard to find and is specific to Java on MacOS. Do you want this PR to add in policy platform rules ? I can make the change.

panga: Fixed, moved to unstrusted rules.

lukehinds (approved):

PR #997, opened 2026-05-22
feat: support fine-grained method+path restrictions in allow_domain

### LINKED ISSUE #960: Support fine-grained method + path restrictions

When using allow_domain with a custom_credential for a service like GitHub's API, the agent gets access to all endpoints on that domain. There's no way to restrict which specific paths or HTTP methods the agent can call.

For example, if I configure a GitHub credential with allow_domain: ["api.github.com"], the agent can access any organization, repository, or administrative endpoint — not just the specific org I intend it to work with. I want to grant access to GET /orgs/my-org/** without implicitly allowing GET /orgs/other-org/**.

This also applies to cases where no credential injection is needed at all. An agent might need to call a public API or an internal service where authentication is handled by the agent itself (e.g., a token already in its context). In these cases, you still want to restrict which endpoints the agent can reach — limiting blast radius isn't only about protecting secrets, it's about constraining what actions the agent can take. For example, allowing an agent to read from a public data API at GET /v1/data/** while preventing it from hitting POST /v1/admin/** on the same host.

Following the principle of least privilege: the domain-level granularity is too coarse for services that expose multi-tenant or multi-scope APIs behind a single hostname.

Extend allow_domain to accept objects alongside plain strings, allowing endpoint-level restrictions per domain:

```json
  {
    "network": {
      "allow_domain": [
        "api.openai.com",
        {
          "domain": "api.github.com",
          "endpoints": [
            { "method": "GET", "path": "/orgs/my-org/**" },
            { "method": "GET", "path": "/repos/my-org/**" },
            { "method": "POST", "path": "/repos/my-org/*/issues" }
          ]
        },
        {
          "domain": "internal-service.example.com",
          "endpoints": [
            { "method": "GET", "path": "/v1/data/**" }
          ]
        }
      ]
    }
  }
```

When endpoints is present, the behavior is default-deny — only matching method+path combinations are allowed. The existing glob patterns (* = one segment, ** = zero or more segments) from endpoint_rules in proxy route config would apply. Endpoint restrictions work independently of whether a custom_credential is configured — they constrain reachability regardless of how the agent authenticates.

There no workaround.

### How is this blocking you?

### Additional context

The proxy already has the internal machinery for this (EndpointRule matching with glob patterns, percent-decoding, method matching). The main work is surfacing it in the profile/manifest schema and wiring the  profile-level config through to the proxy's route config during startup. For domains without a credential, routing through the reverse proxy (needed for path inspection) would also require the sandboxed process to trust the proxy CA — which is already handled for credential routes.

panga:
## Linked Issue

Closes #960

## Summary

Extends `allow_domain` to accept objects with fine-grained method+path endpoint restrictions alongside plain hostname strings. When an entry includes `endpoints`, the proxy performs TLS interception for L7 filtering (default-deny) — only requests matching at least one rule are forwarded.

**Example profile:**

```json
{
  "network": {
    "allow_domain": [
      "api.openai.com",
      {
        "domain": "api.github.com",
        "endpoints": [
          { "method": "GET", "path": "/repos/my-org/**" },
          { "method": "POST", "path": "/repos/my-org/*/issues" }
        ]
      }
    ]
  }
}
```

**Key design decisions:**

- Reuses existing `EndpointRule` + `CompiledEndpointRules` + TLS interception machinery from the reverse proxy — no new proxy logic needed
- Endpoint-restricted domains become `RouteConfig` entries (prefix `_ep_<domain>`) without credential injection
- Profile `extends` **appends** endpoint rules for the same domain (never replaces), matching the user's expectation for additive inheritance
- Proxy mode is automatically enabled when endpoint rules are present
- `NONO_CAP_FILE` includes `domain_endpoints` field for introspection by sandboxed processes
- `nono why --host` reports endpoint restrictions when a domain is allowed

## Agent Disclosure

This PR was co-authored by an AI agent (Claude). Files consulted:
- `crates/nono-proxy/src/config.rs` (existing `EndpointRule`, `RouteConfig`, `CompiledEndpointRules`)
- `crates/nono-proxy/src/route.rs` (route loading, `requires_intercept` logic)
- `crates/nono-cli/src/profile/mod.rs` (profile merge, `NetworkConfig`)
- `crates/nono-cli/src/network_policy.rs` (credential resolution, proxy config building)
- `crates/nono-cli/src/proxy_runtime.rs` (proxy launch pipeline)

## Test Plan

- 9 new unit tests covering deserialization, merge behavior, and partition logic
- `cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::unwrap_used` passes
- `cargo fmt --all -- --check` passes
- 1042/1043 tests pass (1 pre-existing failure unrelated to this change)
- Manual verification: profile with mixed `allow_domain` entries produces correct `ProxyConfig` with plain hosts in `allowed_hosts` and endpoint routes in `routes`

## Checklist

- [x] An issue already exists


lukehinds (approved):
This is a really nice change thanks @panga

PR #1005, opened 2026-05-24
refactor(pack-hints): refresh in detached process to avoid threads

### LINKED ISSUE #1004: Avoid pre-fork pack update hint thread in supervised mode
`nono run` can fail in supervised mode with:

```
Cannot fork: process has 8 threads (max 7 with crypto pool)
```

This happens because pack update hint refresh currently starts a detached
background thread during sandbox preparation, before supervised execution
reaches its fork() safety check.

## Expected Behavior

Pack update hint refresh should not add threads to the supervised parent
process before fork().

## Proposed Fix

Move stale pack update hint refresh out of the parent process. For example:

- keep cached hint printing synchronous and local
- refresh stale cache entries in a detached helper process instead of a
  detached thread
- add an opt-out env var for users who do not use packs:

`NONO_NO_PACK_UPDATE_HINTS=1`

The existing global opt-out, NONO_NO_UPDATE_CHECK=1, should continue to disable this too.

lukehinds:
The pack update hint refresh mechanism now uses a detached child process instead of a background thread. This ensures that the `nono run` command, when operating in supervised mode, does not create an additional thread in the parent process before forking the supervisor.

- Introduce `pack-update-hint-helper` subcommand for out-of-process refresh.
- Add `NONO_NO_PACK_UPDATE_HINTS` environment variable to specifically disable pack update hints, distinct from the general update check opt-out.
- Disable pre-execution update checks for the new helper command to prevent recursion.
- Update relevant documentation to reflect these changes.

Fixes: #1004

### DIFF
--- a/crates/nono-cli/src/app_runtime.rs
+++ b/crates/nono-cli/src/app_runtime.rs
@@ -118,6 +118,7 @@ fn dispatch_command(
             run_command_with_update(update_handle, silent, || package_cmd::run_outdated(args))
         }
         Commands::OpenUrlHelper(args) => run_open_url_helper(args),
+        Commands::PackUpdateHintHelper(args) => crate::pack_update_hint::run_refresh_helper(args),
         Commands::Completions(args) => run_completions(args),
     }
 }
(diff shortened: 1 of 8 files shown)


SequeI (approved):
lgtm!

PR #1020, opened 2026-05-26
fix: correct Quick Start profile reference in README

guzalv:
## Summary

- The Quick Start example in the root README used `--profile always-further/claude-code`, but there is no pack called `always-further/claude-code`. The correct fully qualified registry pack name is `always-further/claude`.
- Changed to `--profile always-further/claude` per maintainer feedback that the project is moving towards fully qualified registry pack names.

## Test plan

- [x] Verified all source code references use `always-further/claude` as the pack ref
- [x] Confirmed via `migration.rs` tests that `always-further/claude` is the canonical pack ref
- [x] Confirmed direction with maintainer in PR comments

🤖 Generated with [Claude Code](https://claude.com/claude-code)


guzalv:
Hi there! Claude generated this PR. I would have written `nono run --profile always-further/claude -- claude` instead, but Claude's explanation seems solid.

SequeI:
HI!

We are moving towards using the fully qualified registry pack name, in which case your intuition was right : D

It should be `always-further/claude`

SequeI (approved):

SequeI:
Amazing thank you and welcome to the community!

PR #1034, opened 2026-05-27
feat(credential-broker): add macos keychain credential broker

### LINKED ISSUE #1026: Supervised macOS Keychain Broker
coding agents typically always need to read, create, update, or delete credentials in the macOS Keychain during login and token refresh flows. Today this often requires granting direct sandbox access to the user's keychain database:

```json
"read_file": [
  "~/Library/Keychains/login.keychain-db"
]
```

That is the uncomfortable security boundary. The sandboxed process only needs narrow credential operations, not filesystem visibility into the keychain database. Direct keychain file grants also make it difficult to restrict one application profile from accessing credentials intended for another application profile.

The feature must work opaquely for existing agents. We cannot expect upstream agent developers to call a nono-specific credential API. Existing tools use multiple access paths:

native `Security.framework` APIs, usually through language libraries
`/usr/bin/security` subprocess calls from shell, Node, Python, or other runtimes
Both paths need to be mediated without product-specific branches or service-name assumptions.

lukehinds:
Introduces a new system for brokering macOS Keychain access, that allows `nono` profiles to define granular permissions for `security` CLI interactions. This brings in a new `credential-helper` internal subcommand processes credential requests via supervisor IPC. On macOS, a `security` shim is installed in `PATH` when `NONO_EXPERIMENTAL_KEYCHAIN_BROKER` is set, redirecting keychain operations through `nono`. Profiles can now specify `credential_access` grants, including `provider`, `classes`, `operations`, `services`, `service_prefixes`, and `accounts` to control access.


lukehinds:
Still need some work, but pleased to see no dreaded red labels from gemmy.

TBeijen:
Ran some tests as we discussed in Discord.

See: https://github.com/TBeijen/nono-pocs/blob/main/broker-test/RESULT-20260531-0820CEST.txt (and the `.sh` files around it)

TL;DR:

* Intercepts CLI keychain requests that are allowed, triggers OS popup where I need to confirm by entering pw.
* Blocks CLI keychain requests not explicitly allowed
* Only intercepts `security` cli, does not intercept `gh` retrieving credentials from Keychain

In its current form could work, by allowing only access to claude credentials (via cli shim) while blocking access to Keychain in general. As long as Claude doesn't start using `Security.framework`.

I have no SSO account for Claude to actually test its way of storing those credentials.

(🤖 / 👨):
GitHub CLI uses `go-keyring`, calling `SecItemCopyMatching/SecItemAdd` via `Security.framework`. So that would still require blocking KeyChain access entirely.

No originally Claude believed claude used keytar to access Keychain in a similar way to gh, based on it diggin through itself using `strings`. But when pressed harder, since [keytar is archived](https://github.com/atom/node-keytar), it came to the following conclusion:

```
I was wrong. Claude Code does use the security CLI, not native SecItem* APIs directly.

This is definitive. Claude Code shells out to the security CLI for all Keychain operations:

  - Read: security find-generic-password -a "<account>" -w -s "<service>"
  - Write: security add-generic-password -U -a <account> -s <service> -X <hex-data> (or via security -i stdin for large payloads)
  - Delete: security delete-generic-password -a <account> -s <service>
  - Probe: security show-keychain-info

  The SecItemCopyMatching/SecItemAdd symbols I found earlier are just in the compiled security binary strings or framework metadata baked into the single binary — they're not called
   directly by Claude Code's Node.js runtime.

  This is great news for the credential broker. It means nono's existing shim approach (replacing security in PATH) would work for Claude Code's OAuth tokens. The architecture you
  asked about earlier — block Keychain by default, intercept via shim — is exactly what would work here:

  1. Block Keychain at Mach level (deny_keychains_macos)
  2. Shim security CLI in PATH (broker already does this)
  3. Claude's /login OAuth calls go through the shim → broker validates against credential_access policy → supervisor (unsandboxed) executes the real security command

  The gh CLI remains the odd one out since it uses go-keyring (native APIs). But for Claude itself, the broker approach is viable. The --env-credential / GH_TOKEN injection handles
  gh.
```

TBeijen:
If we phrase the goal as:

> Don't mind claude accessing it's own SSO credentials, but don't grant it access to anything else in Keychain

Then there might be some pointers in thise OpenShell issue which tackles the same problem: https://github.com/NVIDIA/OpenShell/issues/620

PR #1049, opened 2026-05-30
feat: remove libdbus dependency on linux

### LINKED ISSUE #1047: Switch Linux keyring backend to pure-Rust zbus (async-secret-service) so default builds don't link libdbus
The default Linux release binary fails to start on hosts without libdbus

`nono: error while loading shared libraries: libdbus-1.so.3: cannot open shared object file`

This is all from `keyring` is configured with `features = ["sync-secret-service"]`, which pulls
`dbus-secret-service` → `libdbus-sys`. `libdbus-sys` emits `cargo:rustc-link-lib=dbus-1`, baking a
`DT_NEEDED libdbus-1.so.3` entry into the ELF. That entry is resolved by `ld.so` at process startup —
**before `main()` runs** — so the binary won't launch on a libdbus-free host, even though the keyring
is never touched.

The `.tar.gz` artifacts carry no dependency metadata (unlike the `.deb`, where cargo-deb's `$auto` runs
`dpkg-shlibdeps` and adds `libdbus-1-3`), so tarball users on minimal or non-Debian systems hit this.

why the existing `system-keyring` flag (#741 / #861) isn't enough...

`--no-default-features` removes libdbus but also removes native Secret Service / gnome-keyring support.
That forces a binary choice between *portable* and *keyring-capable*. This issue aims for **both** in
the default build.

I think a better approach is to switch the Linux backend from `sync-secret-service` (C libdbus) to keyring's pure-Rust path, which speaks the D-Bus wire protocol over the socket via `zbus` — no `libdbus-sys`, no `DT_NEEDED`:

```toml
# crates/nono/Cargo.toml + crates/nono-cli/Cargo.toml (Linux target deps)
keyring = { version = "3", features = ["async-secret-service", "crypto-rust", "async-io"], optional = true }
```
This seems a like a nice clean work stream..

- The synchronous `Entry` API is unchanged → `keystore.rs` and `trust_keystore.rs` need no code changes.
- `async-io` (not `tokio`) runtime feature chosen deliberately to avoid "cannot `block_on` within a
runtime" panics, since a tokio runtime is already active in-process.
- Removes `dbus` / `dbus-secret-service` / `libdbus-sys`; adds `zbus` / `zvariant` / RustCrypto.
- Lets us drop `libdbus-1-dev` / `pkg-config` from `release.yml` and `docker/Dockerfile`, and libdbus
from the `.deb` dependencies.

## Acceptance criteria

- [ ] Default `cargo build -p nono-cli` on `x86_64-unknown-linux-gnu` produces a binary with no
    `libdbus-1.so.3` in `objdump -p` NEEDED entries.
- [ ] Keyring store/load still works on a Linux desktop with a running Secret Service.
- [ ] On a host with no D-Bus session, keyring calls fail *gracefully at call time* (existing messaging);
    the binary still launches.
- [ ] CI gate added to `release.yml`: fail the release if a portable artifact links libdbus.

## Prior art

This has been raised before in issues #741, #861,  and PR #740 

## Alternatives considered

- **Vendor / static-link libdbus** (`vendored` feature): self-contained, but ships C attack surface
into a security tool. Rejected.
- **Drop keyring from the default tarball**: the #741 path; loses native keyring UX. Not the goal here.

lukehinds:
Migrates the Linux Secret Service keyring backend to use a pure-Rust implementation based on `zbus`, eliminating the need for `libdbus` at both build and runtime. Made this change after seeing some many folks but heads with libdbus not being available on different platforms, especially in containerized or minimal environments where `libdbus` might not be present

## Linked Issue

Closes #1047

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,10 +98,6 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
-      - name: Install system dependencies (Ubuntu)
-        if: runner.os == 'Linux'
-        run: sudo apt-get update && sudo apt-get install -y libdbus-1-dev pkg-config
-
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@631a55b12751854ce901bb631d5902ceb48146f7 # stable
         with:
@@ -120,6 +116,21 @@ jobs:
 
       - name: Build
         run: cargo build --workspace --verbose
+
+      - name: Verify nono does not link libdbus (Linux)
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          set -euo pipefail
+          bin="target/debug/nono"
+          needed="$(objdump -p "$bin" | awk '/NEEDED/ {print $2}')"
+          echo "NEEDED libraries:"
+          echo "$needed"
+          if echo "$needed" | grep -qi '^libdbus'; then
+            echo "::error::$bin links libdbus — the Linux keyring backend regressed off the pure-Rust zbus path"
+            exit 1
+          fi
+          echo "OK: no libdbus dependency"
 
       - name: Run tests
         run: cargo test --workspace --verbose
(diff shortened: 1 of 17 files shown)


SequeI (approved):
Amazing !!

lukehinds:
Thanks @SequeI !

PR #1050, opened 2026-05-31
fix(macos): allow com.apple.SecurityServer for Secure Transport TLS on macOS 13+

zhxiaogg:
**Draft — I'm not sure this is the right way to fix the issue; feedback welcome.**

## Problem
cargo's bundled libgit2 uses macOS Secure Transport, which does a `mach-lookup com.apple.SecurityServer` during the TLS handshake to validate the server certificate. The seatbelt profile denies that service, so HTTPS git fetches fail with `errSSLBadCert (-9808)` inside the sandbox. (`curl` and system `git` work — they validate via `trustd`, which is allowed.)

## Change
Stop denying `com.apple.SecurityServer` on **macOS 13+**, where the keychain is reached via `keychaind`/`secd` (still denied) — so TLS works while keychain secrets stay sealed. On **< 13** it is itself a keychain daemon, so it's kept denied. The macOS version is read at runtime from `kern.osproductversion`.

## Why I'm unsure
- Verified only on macOS 26 (a canary keychain secret stays un-extractable with SecurityServer allowed). The `13` boundary comes from the existing daemon-era comments, not from testing on Ventura — a higher floor (e.g. 14) might be safer.
- This overlaps conceptually with the keychain escape-hatch (#616) and the broker plan (#1026). The 'right' fix might be a dedicated capability/flag rather than an unconditional allow.
- I may be missing side effects of allowing SecurityServer.

### DIFF
--- a/crates/nono/src/sandbox/macos.rs
+++ b/crates/nono/src/sandbox/macos.rs
@@ -476,6 +476,48 @@ fn push_localhost_tcp_outbound_seatbelt_rules(profile: &mut String, localhost_po
     }
 }
 
+/// The running macOS major version (e.g. `13` for Ventura) from `kern.osproductversion`,
+/// or `None` if it cannot be read. Cached: the version is fixed for the process lifetime.
+fn macos_major_version() -> Option<u32> {
+    static MAJOR: std::sync::OnceLock<Option<u32>> = std::sync::OnceLock::new();
+    *MAJOR.get_or_init(|| {
+        let name = c"kern.osproductversion";
+        let mut size: libc::size_t = 0;
+        // First query the length of the version string.
+        let rc = unsafe {
+            libc::sysctlbyname(
+                name.as_ptr(),
+                std::ptr::null_mut(),
+                &mut size,
+                std::ptr::null_mut(),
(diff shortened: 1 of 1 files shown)


lukehinds:
hi @zhxiaogg 

you can add this escape hatch to a user profile:

`nono profile init claude2 --extends always-further/claude --full`

I put claude for extends (inherits) , but you can put whatever agent profile you want to use.

You then add a top level `unsafe_macos_seatbelt_rules`

```
  {
    "meta": {
      "name": "libgit2",
      "version": "0.1.0"
    },
      "(allow mach-lookup (global-name \"com.apple.SecurityServer\"))"
    ],
    }
  }
```

Let me know if that works

panga:
I'm not sure how this would work without `--trust-proxy-ca` unreleased feature (https://github.com/always-further/nono/pull/978). Please try to see if it solves your problem.

zhxiaogg:
@lukehinds the solution works for my use case, thank you so much! I'll close this PR now.

zhxiaogg:
@panga sorry I'm not sure why, but my use case works without the fix.

Issue #1048, opened 2026-05-30
TLS CONNECT interception ignores configured upstream_proxy


lukehinds:
TLS CONNECT interception for managed credential routes appears to ignore the configured `upstream_proxy` / `external_proxy`. After nono accepts the local CONNECT and intercepts TLS, the upstream leg is opened directly to the target host instead of chaining through the configured corporate proxy.

This can make requests hang or fail in environments where the target is only reachable through the upstream proxy.

## User-visible symptom

A profile like this:

```json
{
  "security": {
    "wsl2_proxy_policy": "insecure_proxy"
  },
  "network": {
    "block": false,
    "allow_domain": ["*.internal.com"],
    "credentials": ["my_api"],
    "custom_credentials": {
      "my_api": {
        "upstream": "https://server.internal.com",
        "credential_key": "my_api_basic_auth",
        "inject_mode": "basic_auth"
      }
    },
    "upstream_proxy": "proxy.corporate.com:80",
    "upstream_bypass": []
  }
}
```

and a command like:

```bash
nono run -v --profile test ./nono-test.sh
# where nono-test.sh runs:
curl -v https://server.internal.com/api/auth
```

shows curl authenticating to the local nono proxy via `Proxy-Authorization`, then the request hangs after TLS interception. The `Proxy-Authorization` header itself is expected and is separate from upstream credential injection.

## Local reproduction

I validated this with a fake upstream proxy on `127.0.0.1:18080` that logs any accepted connection and returns `502`.

Profile used:

```json
{
  "security": {
    "wsl2_proxy_policy": "insecure_proxy"
  },
  "network": {
    "block": false,
    "allow_domain": ["example.com"],
    "credentials": ["probe_api"],
    "custom_credentials": {
      "probe_api": {
        "upstream": "https://example.com",
        "credential_key": "env://NONO_PROBE_BASIC",
        "env_var": "NONO_PROBE_BASIC",
        "inject_mode": "basic_auth"
      }
    },
    "upstream_proxy": "127.0.0.1:18080",
    "upstream_bypass": []
  }
}
```

Command:

```bash
NONO_PROBE_BASIC='user:pass' target/debug/nono run -v \
  --profile /tmp/nono-upstream-proxy-probe.json \
  -- curl -vk https://example.com/
```

Observed result:

- nono reports TLS interception active for the credential route
- curl CONNECTs to the local nono proxy successfully
- nono logs `mode=connect_intercept`
[...]

PR #1091, opened 2026-06-06
fix(proxy): respect upstream_proxy in TLS CONNECT intercept path (#1048)

caiocdcs:
The TLS intercept upstream leg (used for credential injection) always connected directly to the target host, ignoring the configured `upstream_proxy` / `external_proxy`. In enterprise environments where the target is only reachable through a corporate proxy, this caused the intercepted request to hang or fail.

Issue: https://github.com/always-further/nono/issues/1048

### Approach

Two changes in `server.rs::handle_connection` and `tls_intercept::handle::forward_inner_request`:

1. **server.rs** — before entering the intercept path, check the bypass matcher against the target host. If the host should go through the upstream proxy and auth is not configured, pass the proxy address into `InterceptCtx` as `InterceptUpstreamProxy`. This mirrors the existing bypass logic for transparent CONNECT.
2. **handle.rs** — `forward_inner_request` selects `UpstreamStrategy::ExternalProxy` when `ctx.upstream_proxy` is set, falling back to `UpstreamStrategy::Direct`. The `forward.rs` pipeline already knew how to chain through a proxy — the intercept path just never asked it to.

### Files changed

- `crates/nono-proxy/src/tls_intercept/handle.rs` — Added `InterceptUpstreamProxy` struct + `upstream_proxy` field; strategy selection in `forward_inner_request`; 3 unit tests.
- `crates/nono-proxy/src/tls_intercept/mod.rs` — Re-export `InterceptUpstreamProxy`.
- `crates/nono-proxy/src/server.rs` — Bypass \& auth gate before `InterceptCtx` construction.

### Compliance

- [x] An issue already exists (#1048)
- [ ] I described my intent and approach in the issue discussion
- [ ] I used NonoError where required (proxy crate uses `ProxyError`, not applicable to library `NonoError`)
- [ ] I validated and canonicalized all relevant paths (no file path changes in this PR)
- [x] This PR matches the disclosed issue scope

AI-assisted contribution.


SequeI (approved):
LGTM, thank you!

PR #1108, opened 2026-06-08
fix: --detached (Linux): sandboxed process denied access to /dev/null

### LINKED ISSUE #1064: --detached (Linux): sandboxed process denied access to /dev/null
# `--detached` (Linux): sandboxed process denied access to `/dev/null`

**Labels:** `bug`, `triage`

## Summary

When running `nono run --detached` on Linux, the sandboxed command cannot read
from or write to `/dev/null`. The same profile and command work correctly in
attached mode.

## Steps to reproduce

```sh
nono run --detached -- bash -c 'cat /dev/null && echo ok || echo FAIL'
nono attach <session-id>
# observe: FAIL (or the command exits 1 / bash error)
```

Any command that explicitly opens `/dev/null` inside the sandbox exhibits the
same failure. Profile does not matter — the default profile reproduces it.

## Expected behaviour

`/dev/null` is accessible (read and write) inside the sandbox regardless of
whether the session was started with `--detached`.

## Actual behaviour

`open("/dev/null")` returns `EACCES`. The Landlock sandbox has no rule for the
`/dev/null` inode, so any attempt to open the file by path is denied.

## Root cause

The bug is in `crates/nono/src/capability.rs`,
`CapabilitySet::deduplicate()`, Linux-only `original_updates` logic, triggered
by a specific interaction with `remap_procfs_self_references()` in the forked
child.

**Sequence of events:**

1. In detached mode, nono starts with `stdin = /dev/null` and
   `stdout = /dev/null` (set by `startup_runtime::run_detached_launch`).

2. When capabilities are built in this parent process, `system_read_linux_core`
   adds both an explicit `/dev/null` file entry *and* a `/dev/stdin` file
   entry. Because `FsCapability::new_file` calls `Path::canonicalize()`,
   `/dev/stdin` follows the chain `/dev/stdin → /proc/self/fd/0 → /dev/null`
   and resolves to the same inode. `system_write_linux` does the same via
   `/dev/stdout`.

3. `deduplicate()` runs with Linux keying on `(resolved, is_file)`. Both
   entries share the key `("/dev/null", true)` and collapse to one. A
   side-effect (`original_updates`) updates the surviving direct entry's
   `original` field from `/dev/null` to `/dev/stdin` (read) and `/dev/stdout`
   (write).

4. In the forked child, `remap_procfs_self_references()` iterates capabilities
   and acts on `cap.original`. It recognises `/dev/stdin` as a special alias
   and rewrites `cap.resolved` to `/proc/{child_pid}/fd/0`. Same for
   `/dev/stdout` → `/proc/{child_pid}/fd/1`.

5. `setup_child_pty()` then runs, replacing fd 0 and fd 1 with the PTY slave.

6. `Sandbox::apply()` calls `PathFd::new("/proc/{child_pid}/fd/0")` — which
   now opens the **PTY slave** inode. The Landlock rule is attached to the PTY
   slave, not `/dev/null`.

7. The sandboxed process has no Landlock rule for `/dev/null` and is denied.

In attached mode, `/dev/stdin` resolves to the real TTY device (`/dev/pts/X`),
which has a different inode from `/dev/null`. Dedup never merges them, the
explicit `/dev/null` entry keeps `original = /dev/null`, and
`remap_procfs_self_references()` ignores it.

The same issue would affect `/dev/zero`, `/dev/urandom`, etc. if they were ever
used as stdin/stdout, but `/dev/null` is the common case.

## Proposed fix

In `deduplicate()`, guard both `original_updates` push sites to skip updating
`original` when the symlink path being inherited is one of the four paths that
`remap_procfs_self_references()` handles specially: `/dev/fd`, `/dev/stdin`,
`/dev/stdout`, `/dev/stderr`.

A direct entry (e.g. `/dev/null`) whose `original` is silently renamed to one
of these will be wrongly remapped in the child. Preserving
`original = /dev/null` prevents the misdirection entirely.

```rust
/// Returns true if `path` is a special /dev alias that
/// remap_procfs_self_references() rewrites to /proc/{pid}/fd/N.
#[cfg(target_os = "linux")]
fn is_procfs_remap_original(path: &Path) -> bool {
    matches!(
        path.to_str(),
        Some("/dev/fd" | "/dev/stdin" | "/dev/stdout" | "/dev/stderr")
    )
}

// In deduplicate(), both original_updates push sites:
if existing.original == existing.resolved
    && cap.original != cap.resolved
    && !is_procfs_remap_original(&cap.original)   // <-- guard
{
    original_updates.push((existing_idx, cap.original.clone()));
}
```

A regression test `remap_preserves_dev_null_when_deduped_with_dev_stdin`
should be added to the `procfs_remap_tests` module to lock in the correct
behaviour.

scp7:
## Summary

  Fixes #1064.

  - **Root cause**: In `--detached` mode, stdin/stdout are `/dev/null`.
  `system_read_linux_core` adds both an explicit `/dev/null` file capability and a `/dev/stdin`
  symlink capability. Both resolve to the same inode, so `deduplicate()` collapses them and
  its `original_updates` logic silently renames the surviving entry's `original` from
  `/dev/null` to `/dev/stdin`. `remap_procfs_self_references()` then treats that as a fd alias
  and rewrites `resolved` to `/proc/{child_pid}/fd/0` (the PTY slave). The Landlock rule ends
  up on the wrong inode; the sandboxed process gets EACCES on any `open("/dev/null")`.

  - **Fix**: Add `is_procfs_remap_original()` and guard both `original_updates` push sites in
  `deduplicate()`. When the symlink original being inherited is one of the four `/dev/` aliases
  that `remap_procfs_self_references()` rewrites (`/dev/fd`, `/dev/stdin`, `/dev/stdout`,
  `/dev/stderr`), the inheritance is skipped and the surviving direct entry keeps its own
  `original`.

  - **Test**: `remap_preserves_dev_null_when_deduped_with_dev_stdin` in `procfs_remap_tests`
  verifies both that `deduplicate()` leaves `original = /dev/null` intact and that
  `remap_procfs_self_references()` subsequently leaves `resolved = /dev/null` unchanged.

  ## Changed files

  - `crates/nono/src/capability.rs` — guard + helper + regression test (Linux-only)

  ## Agent compliance checklist

  - [x] An issue already exists (#1064)

  *This change was authored by an AI agent (Claude Code) under human review.*

### DIFF
--- a/crates/nono/src/capability.rs
+++ b/crates/nono/src/capability.rs
@@ -1546,8 +1546,16 @@ impl CapabilitySet {
                     seen.insert(key, i);
                     // On Linux: preserve symlink original from the removed
                     // entry into the kept entry so `original` stays meaningful.
+                    // Guard: skip if the symlink original is one of the four
+                    // /dev aliases that remap_procfs_self_references() rewrites
+                    // to /proc/{pid}/fd/N — inheriting such an original would
+                    // cause a direct entry (e.g. /dev/null) to be misdirected
+                    // to a PTY slave inode when the remap runs in the child.
                     #[cfg(target_os = "linux")]
-                    if cap.original == cap.resolved && existing.original != existing.resolved {
+                    if cap.original == cap.resolved
+                        && existing.original != existing.resolved
+                        && !is_procfs_remap_original(&existing.original)
+                    {
                         original_updates.push((i, existing.original.clone()));
                     }
                     // Apply merged access to the new (kept) entry
@@ -1557,8 +1565,15 @@ impl CapabilitySet {
                 } else {
                     // On Linux: inherit symlink original from the entry
                     // being discarded into the surviving entry.
+                    // Guard: skip if the discarded entry's original is one of
+                    // the four /dev aliases that remap_procfs_self_references()
+                    // rewrites to /proc/{pid}/fd/N — see the keep_new branch
+                    // above for the rationale.
                     #[cfg(target_os = "linux")]
-                    if existing.original == existing.resolved && cap.original != cap.resolved {
+                    if existing.original == existing.resolved
+                        && cap.original != cap.resolved
+                        && !is_procfs_remap_original(&cap.original)
+                    {
                         original_updates.push((existing_idx, cap.original.clone()));
                     }
                     to_remove.push(i);
@@ -1764,6 +1779,19 @@ impl CapabilitySet {
     }
 }
 
+/// Returns `true` if `path` is any path that [`rewrite_procfs_self_reference`]
+/// would rewrite — i.e. inheriting it as an `original` in
+/// [`CapabilitySet::deduplicate`] would cause a subsequent
+/// `remap_procfs_self_references` call to misdirect the resolved inode.
+///
+/// Implemented by delegating to [`rewrite_procfs_self_reference`] so the two
+/// functions are always in sync: any path added to the rewriter is
+/// automatically covered here without a separate update.
+#[cfg(target_os = "linux")]
+fn is_procfs_remap_original(path: &Path) -> bool {
+    rewrite_procfs_self_reference(path, 0, None).is_some()
+}
+
 fn rewrite_procfs_self_reference(
     original: &Path,
     process_pid: u32,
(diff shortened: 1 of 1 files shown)


lukehinds (approved):
lgtm,  /dev/null is the one device where granting any extra access carries essentially zero risk - will merge this as I work via the backlog.

SequeI (approved):

Issue #1099, opened 2026-06-08
(network) separate proxy features from proxy activation


lukehinds:
Split the current proxy options into separate intent structs:

```text
ProxyActivation
DomainFilterIntent
CredentialProxyIntent
EndpointFilterIntent
UpstreamProxyIntent
TlsInterceptIntent
OpenUrlIntent
```

The activation rule should be obvious:

```text
proxy starts if DomainFilterIntent or CredentialProxyIntent or
EndpointFilterIntent or UpstreamProxyIntent is non-empty
```

`--proxy-port`, `--trust-proxy-ca`, and `--proxy-ca-validity` etc should be rejected unless something actually activates the proxy.

Issue #1098, opened 2026-06-08
(network) typed NetworkPlan


lukehinds:
add a typed `NetworkPlan`

Create one CLI-owned type that represents resolved user intent before any proxy is started:

```rust
enum NetworkIntent {
    Unrestricted,
    BlockAll {
        localhost_ports: Vec<u16>,
    },
    ProxyFiltered {
        domain_policy: DomainPolicy,
        credential_routes: Vec<CredentialRouteIntent>,
        endpoint_routes: Vec<EndpointRouteIntent>,
        upstream_proxy: Option<UpstreamProxyIntent>,
        listen_ports: Vec<u16>,
        localhost_ports: Vec<u16>,
        fixed_proxy_port: Option<u16>,
        tls_intercept: TlsInterceptIntent,
    },
    LinuxDirectPorts {
        connect_ports: Vec<u16>,
        bind_ports: Vec<u16>,
    },
}
```

Then have a second type for post-proxy materialization:

```rust
struct NetworkEnforcementPlan {
    mode: NetworkMode,
    localhost_ports: Vec<u16>,
    tcp_connect_ports: Vec<u16>,
    tcp_bind_ports: Vec<u16>,
    linux_seccomp_notify: LinuxNetworkNotifyMode,
    proxy_env: Vec<(String, String)>,
}
```

Benefits:

- No more `ProxyOnly { port: 0 }` placeholders in `CapabilitySet`.
- No more `network_block_requested` side channel.
- Just one place to validate incompatible flags.
- and one place to print a human-readable summary.

PR #1126, opened 2026-06-11
refactor(networking): typed NetworkIntent, decouple --credential from OS isolation

### LINKED ISSUE #1031: --credential / proxy mode implicitly blocks networking; proxy and credential injection should be independent
`--credential` and proxy-based credential injection implicitly block all outbound networking and localhost (see #588), even when network blocking is not the intent. Using credential injection forces the Seatbelt `(deny network*)` rule, which breaks DNS and restricts localhost to only the proxy port.

The proxy has three documented modes (CONNECT tunnel, reverse proxy, external proxy) that can be combined, but they all trigger the same network-blocking codepath. There's no way to use the reverse proxy for credential injection only while keeping networking open.

I would like to suggest decoupling the credential injection process from the network blocking functionality.
- `--credential` → starts the reverse proxy, injects headers
- `--block-net` / `--allow-domain` → controls network access

Without `--block-net` or `--allow-domain`, the proxy should inject credentials without dropping `(deny network*)` on the child.

**Steps to reproduce:**
1. `nono run --credential my_service -- nono why --self --host localhost`
2. Returns blocked despite no `--block-net` or `--allow-domain` being set — the credential flag alone should not imply network blocking

**Relevant output / logs:**
```
 DENIED
   Reason: network_blocked
   Details: Network access is fully blocked. Connection to localhost:443 would be denied.
   Suggested fix: --allow-domain localhost
```

**Environment:**
- macOS Tahoe 26.5
- nono v0.58.0

**Profile JSON:**
```json
{
  "network": {
    "custom_credentials": {
      "my_service": {
        "upstream": "https://api.example.com",
        "credential_key": "my_token",
      }
    }
  }
}
```

SequeI:
## Summary

Introduce NetworkIntent (Unrestricted / BlockAll / ProxyFiltered) to replace the network_block_requested: bool side channel in PreparedSandbox. Network intent is now resolved once from CLI flags and profile config and carried through to start_proxy_runtime, eliminating the ProxyOnly { port: 0 } placeholder pattern from capability_ext.rs.

--credential no longer activates OS-level ProxyOnly enforcement. Credentials inject headers transparently without restricting outbound traffic. Pair with --allow-domain or --network-profile to also enforce host-level filtering.

A small PR to start off the networking refactor

## Linked Issue

Fixes #1031
Ref #1098
Ref #1099

## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,6 +23,7 @@ The library is a **pure sandbox primitive**. It applies ONLY what clients explic
 | `QueryContext` | All output and UX |
 | `keystore` | `learn` mode |
 | `undo` module (ObjectStore, SnapshotManager, MerkleTree, ExclusionFilter) | Rollback lifecycle, exclusion policy, rollback UI |
+| `NetworkMode` (enforcement detail — requires a bound port) | `NetworkIntent` (policy intent — resolved from flags and profile before proxy starts) |
 
 ## Build & Test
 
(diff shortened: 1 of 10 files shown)


SequeI:
Found some other networking stuff that needs to be looked at, putting into draft for now and will do it all in one swoop

PR #1132, opened 2026-06-12
fix(proxy): stop allow_domain endpoint route from shadowing credential catch-all

### LINKED ISSUE #1131: credential catch-all is shadowed by an endpoint-only route, so the credential is not injected
### What happened?

When a profile reaches a host through **both**:

- an `allow_domain` entry that carries `endpoints` (which the proxy compiles into a
  credential-less `_ep_<domain>` endpoint-authorization route), **and**
- a `custom_credentials` entry for the same upstream with **no** `endpoint_rules`
  (a credential catch-all),

a request whose path is authorized by the `_ep_` rule is **forwarded without the
credential injected**. The credential catch-all route is shadowed by the
credential-less `_ep_` route during route selection, so the managed token is never
attached even though the request reaches the upstream.

On a path the `_ep_` route authorizes, the credential catch-all should win and the
managed credential should be injected (scoping and token together). Paths that match
no `_ep_` rule should still be hard-denied (403) — the default-deny behavior must be
preserved.

### Steps to reproduce

1. Create a profile that puts a host in `allow_domain` with org-scoped `endpoints`
   **and** declares a `custom_credentials` catch-all (no `endpoint_rules`) for the
   same upstream, e.g.:

   ```json
   {
     "network": {
       "allow_domain": [
         { "domain": "github.com", "endpoints": [ { "method": "*", "path": "/org/**" } ] }
       ],
       "credentials": ["github_api"],
       "custom_credentials": {
         "github_api": {
           "upstream": "https://github.com",
           "credential_key": "env://GH_TOKEN",
           "credential_format": "Bearer {}",
           "env_var": "GH_TOKEN"
         }
       }
     }
   }
   ```
2. Run an agent under this profile and make a request to an endpoint-authorized path
   (e.g. `GET https://github.com/org/repo`).
3. Observe that the request is forwarded **without** the `Authorization: Bearer <GH_TOKEN>`
   header — the catch-all credential was shadowed by the `_ep_github.com` route.

### Relevant output or logs

```shell

```

### nono version

0.62.0

### Operating system

macOS

### OS version / distro

macOS

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

panga:
## Linked Issue

Closes #1131

## Summary

Fixes a route-selection bug where a credential **catch-all** (`custom_credentials`
with no `endpoint_rules`) was shadowed by a credential-less `_ep_` endpoint-authorization
route (from an `allow_domain` entry with `endpoints`) on the same upstream. On a path the
`_ep_` route authorized, the proxy selected the credential-less `_ep_` match ahead of the
catch-all and forwarded the request **without injecting the managed credential**.

The fix changes the selection priority so a credential-bearing catch-all wins over a
credential-less `_ep_` match on authorized paths. Selection priority is now:

1. a matched route carrying a managed credential;
2. a credential-bearing catch-all (the new step — prevents the shadowing);
3. any other match (bare endpoint authorization);
4. the catch-all (un-credentialed passthrough).

As part of the fix, the route-selection logic (the `_ep_` hard-deny gate, the ambiguity
check, and the priority ordering) was extracted from `tls_intercept/handle.rs` into a single
`select_route` function + `RouteSelection` enum in `route.rs`. This gives the decision one
source of truth shared with its unit tests — the previous test was a hand-copied mirror that
didn't even model the `_ep_` gate, so it could not have caught this bug.

### What is intentionally unchanged

- **`_ep_` routes are kept.** They enforce auth-less default-deny on paths; removing or merging
  them would let paths the credential's `endpoint_rules` don't match through un-credentialed.
- **The endpoint hard-deny gate** (`has_endpoint_only_route && !endpoint_authorized` → 403) and
  the **ambiguity check** (multiple credential routes → 403) are byte-for-byte equivalent to the
  prior behavior, including the audit `denial_category`, reason strings, and HTTP status codes.
- **The host gate is untouched:** a `custom_credentials`-only host that is not in `allow_domain`
  stays denied at the host filter.

## Agent Disclosure

- This PR was generated by an AI agent (Claude Code).

## Test Plan

- Added `test_route_selection_credential_catchall_not_shadowed` in `route.rs`:
  - authorized path (`GET /org/repo`) → credential catch-all selected (token injected) — this
    fails on the old ordering and passes after the fix;
  - non-matching path (`GET /other/repo`) → `EndpointDenied` (default-deny preserved).
- Updated `test_route_selection_multi_org_profile` to exercise the real `select_route` via a thin
  test adapter instead of a hand-copied mirror.

Manual repro: profile with `github.com` in `allow_domain` (org-scoped `endpoints`) + a
`custom_credentials` `github_api` catch-all → authorized path now injects the token; a
non-authorized path is still 403.

## Checklist

## Agent Compliance Check

- [x] An issue already exists
</content>


crates/nono-proxy/src/route.rs
```diff
@@ -307,6 +307,87 @@ impl RouteStore {
...
+    }
+
+    let selected = matches
+        .iter()
+        .find(|(_, route)| route.requires_managed_credential)
+        .copied()
+        .or_else(|| catch_all.filter(|(_, route)| route.requires_managed_credential))
+        .or_else(|| matches.first().copied())
+        .or(catch_all);
+    RouteSelection::Selected(selected)
```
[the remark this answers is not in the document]
panga: Should be fixed.

crates/nono-proxy/src/route.rs
```diff
@@ -307,6 +307,101 @@ impl RouteStore {
...
+            }
+        } else if !route.requires_managed_credential {
+            has_endpoint_only_route = true;
+        }
+    }
+
+    // Endpoint-only authorization layer: a credential catch-all cannot bypass
+    // endpoint restrictions imposed by `_ep_` routes.
+    if has_endpoint_only_route && !endpoint_authorized {
+        return RouteSelection::EndpointDenied;
```
SequeI, line 379: This runs before we ever check `matched_creds` so it does not matter that a credential route already matched, we still bail out early. credential route match is invisible to this check

SequeI (changes requested):
Left a small question about the logic for `matched_creds`, let me know if I am off base there. After that is resolved, we can get this merged :)

crates/nono-proxy/src/route.rs
panga: You're right that the gate runs before we look at matched_cred, and that a matched credential route doesn't flip endpoint_authorized. That's intentional, and it's also the pre-refactor behavior.

The reasoning: an _ep_ route (from an allow_domain entry with endpoints) is a network-level default-deny gate, not a per-credential concern. When such a route exists for the upstream, the request must match the host's allowed endpoint surface before we get to choose/inject a credential. A credential route matching the path authorizes credential injection on that path, but it does not by itself widen the host's endpoint allow-list — otherwise a custom_credentials catch-all could grant access to paths the operator deliberately scoped out via allow_domain endpoints, which is exactly the bypass this gate exists to prevent.

So the ordering is deliberate: gate on the host's endpoint policy first (credential match is intentionally invisible here), then select among credential routes.

SequeI (approved):

crates/nono-proxy/src/route.rs
SequeI: lgtm, thanks!

Issue #1164, opened 2026-06-16
Support AWS sigv4 auth for credential injection


intentionally-left-nil:
### What problem are you trying to solve?

AWS credentials need special support, as their auth scheme is based off [sigv4](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv.html), requiring setting multiple headers based off the request

### What would you like to see?

The user can configure credential injection with the aws mode (like oauth today) and have the credentials be injected

### How is this blocking you?

It's blocking production use of nono

## Subissues:

- [x] #1165 
- [x] #1191 
- [x]  #1189 
- [x] #1286
- [x] #1196 
- [x] #1285

intentionally-left-nil:
Closing this as completed for now 🎉 

Further followup items (supporting http/2, supporting large chunked requests) can be implemented in a future release if there are needed use cases

PR #1166, opened 2026-06-16
feat: [aws] implement aws_auth config

### LINKED ISSUE #1165: AWS: Support aws_auth in the custom_credentials config
This issue covers adding a new `aws_auth` subkeey to the custom credentials. The key is optional. If it is present, it must be a dict. The simplest aws config looks like this:
```
"custom_credentials": {
      "bedrock": {
        "upstream": "https://bedrock-runtime.us-east-1.amazonaws.com",
        "aws_auth": {
          "profile": "my-aws-profile"
        }
      }
    }
  }
```

There are the following optional fields:
* profile - which on-disk AWS profile to use for authentication
* region - which sigv4 signing region to use. When omitted, the region is auto-detected from the upstream URL
* service- which aws service to use. When omitted, the service is auto-detected from the upstream URL from a hardcoded list of known AWS services

The `aws_auth` key is mutually exclusive with the `auth` subkey (oauth), and the `credential_key` field (bearer token. If multiple fields are set, an error is returned during parsing.

## Other validation
* If present, all subkeys must be non-empty strings
* region and service cannot contain whitespace

intentionally-left-nil:
This adds the aws_auth struct to customCredentials, with optional subkeys: profile, region, and service. If the aws_auth subkey is present at all, the route will be signed with AWS credentials in the future

The subkey is mutually exclusive with credential_key and auth.

The routes with aws_auth are wired through to the proxy, where a 501 not implemented is returned for any route using the config

## Linked Issue

Closes #1165

## Summary
This begins to add support to sigv4 (AWS) credentials for the nono proxy. This first PR is about wiring up the aws config in the customCredentials

Example profile:
```json
"custom_credentials": {
      "bedrock": {
        "upstream": "https://bedrock-runtime.us-east-1.amazonaws.com",
        "aws_auth": {
          "profile": "my-aws-profile"
        }
      }
    }
```

The code was generated by claude sonnet, in direct response to prompts, and following the earlier aws spike prototype

## Test Plan
TBD

## Checklist

### DIFF
--- a/crates/nono-cli/src/profile/mod.rs
+++ b/crates/nono-cli/src/profile/mod.rs
@@ -340,6 +340,14 @@ pub struct CustomCredentialDef {
     /// to the certificate in `tls_client_cert`.
     #[serde(default)]
     pub tls_client_key: Option<String>,
+
+    /// Optional AWS SigV4 signing configuration.
+    ///
+    /// When present, the proxy will sign outbound requests with AWS SigV4
+    /// credentials resolved from the configured profile (or the default
+    /// credential chain). Mutually exclusive with `credential_key` and `auth`.
+    #[serde(default)]
+    pub aws_auth: Option<nono_proxy::config::AwsAuthConfig>,
 }
 
 fn default_inject_header() -> String {
@@ -446,6 +454,15 @@ fn validate_credential_key(context_name: &str, key: &str) -> Result<()> {
 ///   - `query_param`: query_param_name required, valid query param name
 ///   - `basic_auth`: no additional required fields
 fn validate_custom_credential(name: &str, cred: &CustomCredentialDef) -> Result<()> {
+    // Mutual exclusion: aws_auth is incompatible with credential_key and auth.
+    if cred.aws_auth.is_some() && (cred.credential_key.is_some() || cred.auth.is_some()) {
+        return Err(NonoError::ProfileParse(format!(
+            "custom credential '{}' has 'aws_auth' set together with 'credential_key' or 'auth'; \
+             aws_auth is mutually exclusive with both — remove the other auth field",
+            name
+        )));
+    }
+
     // Mutual exclusion: credential_key and auth cannot both be set
     if cred.credential_key.is_some() && cred.auth.is_some() {
         return Err(NonoError::ProfileParse(format!(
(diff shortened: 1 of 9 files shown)


LuuOW (commented):
Technical audit: code patterns and implementation verified for alignment with modern software engineering standards.

crates/nono-cli/src/profile/mod.rs
```diff
@@ -687,6 +709,46 @@ fn validate_oauth2_auth(name: &str, auth: &OAuth2Config) -> Result<()> {
     Ok(())
 }
 
+/// Validate AWS SigV4 signing configuration subfields.
+///
+/// Checks:
+/// - `profile`: if set, must be a non-empty string
+/// - `region`: if set, must be a non-empty string with no whitespace
+/// - `service`: if set, must be a non-empty string with no whitespace
```
SequeI: The Region code, service code, and termination string must use lowercase characters.

^^ Got this from the documentation, we should probably validate in this case 

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-signing-elements.html

crates/nono-cli/src/profile/mod.rs
```diff
@@ -687,6 +709,46 @@ fn validate_oauth2_auth(name: &str, auth: &OAuth2Config) -> Result<()> {
     Ok(())
 }
 
+/// Validate AWS SigV4 signing configuration subfields.
+///
+/// Checks:
+/// - `profile`: if set, must be a non-empty string
```
SequeI: AWS profile names are case-sensitive and can't contain spaces, but the current check only rejects empty strings. A value like "my profile" or "My Profile" would pass validation but fail at credential lookup time.

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-signing-elements.html

SequeI (changes requested):
Looks good, just have a few nits

- More lexical constraints, validate service/profile/region etc
- Take a look at `endpoint_rules`, I have a fear it might bring some trouble and need to be remediated once the second PR with full sigv4 drops

Thank you a lot for this PR!

crates/nono-cli/src/profile/mod.rs
intentionally-left-nil: I would like to push back on this. The profile could always be missing, if it's spelled wrong, and it's not the right place to error out on config parsing if the profile is not visible.


I added the whitespace check, but think the rest of this is out of scope

SequeI: That makes sense to me, thanks for the reply!

SequeI (approved):
lgtm @intentionally-left-nil We are planning a release tmwr, just as an fyi if you are also planning to get #1164 in to complete this feature.

intentionally-left-nil:
I can start working on the PR for the proxy completion, but if it takes another release to get in that's okay too. Thanks for the quick turnaround

PR #1169, opened 2026-06-16
docs(install): add version check and COPR fallback note 

connrg:
Adds `nono --version` after the Homebrew, Debian, Fedora, and RPM install steps so users can confirm a working install. Adds a note to the Fedora COPR section pointing to the manual RPM   
  when their Fedora version isn't in COPR.


SequeI (approved):
lgtm, thanks!

PR #1171, opened 2026-06-16
feat(diagnostics): expose structured diagnostics for library and FFI clients

### LINKED ISSUE #991: Expose structured diagnostics from nono and nono-proxy
`crates/nono` and `crates/nono-proxy` are library crates, used by `crates/nono-proxy` and also the library SDKs (or any client who selects nono or nono-proxy as a library) 

We currently have several user-visible states exposed primarily as human-formatted strings, `tracing::warn!` messages, or CLI-shaped diagnostic text. That works reasonably for `nono-cli`, where a TUI is used for output, but it is not a good API boundary for embedded clients using the nono bindings.

The libraries should instead expose only stable, structured diagnostics/errors for machine clients, while keeping `tracing` as operational logging and leaving CLI/TUI rendering in `nono-cli`.

There are two related boundary issues.

### `nono-proxy`

Credential loading can fail in ways that are intentionally non-fatal:

- static credential not found
- static credential unavailable because the keystore/keychain is inaccessible
- OAuth2 `client_id` unavailable
- OAuth2 `client_secret` unavailable
- OAuth2 token exchange failed

Today these cases are generally handled by logging a warning and skipping the credential-backed route, for example:

```rust
warn!(
    "OAuth2 client_id not available for route '{}': {}. \
     Managed-credential requests on this route will be denied.",
    route.prefix, msg
);
continue;
```

That behavior is fail-closed, which is good: the proxy can still start, but managed-credential requests for the affected route are denied. The problem is that library clients have no stable machine-readable way to observe this state. They must either scrape log text, rely on English strings, or miss the condition entirely.

This also creates pressure to add direct `eprintln!`/TUI-style messages in `nono-proxy`, which is the wrong layering: `nono-proxy` is a library crate and should not assume that the caller is `nono-cli` or that stderr is the right output target.

### `nono`

The core `nono` crate is better about direct output: it does not appear to emit runtime `println!`/`eprintln!` messages, apart from build-script output and doc examples. However, it still exposes many human/CLI-shaped strings as the only practical detail available to library clients.

Examples:

- `DiagnosticFormatter::format_footer()` returns a pre-rendered block designed to be printed to stderr. It includes `[nono]`, `nono run`, `nono learn`, `nono why`, and concrete CLI flags.
- `NonoError::CwdPromptRequired` renders as `CWD access requires --allow-cwd in non-interactive mode`.
- `keystore` errors include actionable English such as `Run 'op signin'...` and CLI-specific guidance such as `Use --env-credential-map ...`.
- rollback/snapshot errors include CLI remediation such as `disable rollback with --no-rollback`.

Those messages are useful for `nono-cli`, but bindings should not need to parse English strings to distinguish error causes, remediation classes, credential source failures, or sandbox diagnostic categories.

## Desired state

Add structured diagnostics/errors to library-facing surfaces in both `nono` and `nono-proxy`.

For `nono-proxy`, expose startup and route degradation diagnostics.

Possible shape:

```rust
#[derive(Clone, Debug, Eq, PartialEq)]
pub enum ProxyDiagnosticSeverity {
    Info,
    Warning,
    Error,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum ProxyDiagnosticCode {
    CredentialNotFound,
    CredentialUnavailable,
    OAuthClientIdUnavailable,
    OAuthClientSecretUnavailable,
    OAuthTokenExchangeFailed,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub struct ProxyDiagnostic {
    pub code: ProxyDiagnosticCode,
    pub severity: ProxyDiagnosticSeverity,
    pub route_prefix: String,
    pub credential_ref: Option<String>,
    pub message: String,
    pub hint: Option<String>,
}
```

For `nono`, add structured equivalents for the most important diagnostic and error surfaces. Possible shape:

```rust
#[derive(Clone, Debug, Eq, PartialEq)]
pub enum NonoDiagnosticSeverity {
    Info,
    Warning,
    Error,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum NonoDiagnosticCode {
    SandboxDeniedPath,
    SandboxDeniedNetwork,
    SandboxDeniedUnixSocket,
    CommandNotFound,
    CommandFailedLikelySandbox,
    CommandFailedApplication,
    CredentialNotFound,
    CredentialUnavailable,
    UnsupportedPlatformFeature,
    RollbackBudgetExceeded,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub struct NonoDiagnostic {
    pub code: NonoDiagnosticCode,
    pub severity: NonoDiagnosticSeverity,
    pub message: String,
    pub hint: Option<String>,
    pub remediation: Option<NonoRemediation>,
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub enum NonoRemediation {
    GrantPath {
        path: std::path::PathBuf,
        access: nono::AccessMode,
        is_file: bool,
    },
    GrantNetwork,
    RunDiscovery,
    CheckPolicy,
    AuthenticateCredentialProvider,
    AdjustRollbackBudget,
}
```

The exact API does not need to match these sketches, but it should provide:

- stable diagnostic codes suitable for Rust, C FFI, and Python clients
- severity
- context fields such as route prefix, path, access mode, provider, or platform, where relevant
- redacted credential references, where relevant
- human-readable message for display
- optional remediation hint
- structured remediation data when the library can express the fix without naming CLI flags

## Suggested implementation direction

1. Keep existing `tracing::warn!` calls for operator logs.
2. Do not add direct `println!`/`eprintln!` calls to library crates for human-facing CLI messages.
3. In `nono-proxy`, collect diagnostics while loading credentials/routes.
4. Expose proxy diagnostics through `ProxyHandle`, or return them as part of proxy startup.
5. In `nono`, add structured diagnostic data alongside existing rendered diagnostic strings. Keep `DiagnosticFormatter::format_footer()` for compatibility, but provide a structured representation that CLI and bindings can render differently.
6. For `NonoError` variants that currently store only `String` for security-relevant categories, add stable codes/typed fields where practical, or provide helper methods that classify existing variants into stable public codes.
7. Ensure all credential references in diagnostics are redacted before leaving the library crate.
8. Update `nono-cli` to render structured diagnostics through `crates/nono-cli/src/output.rs`, including CLI-specific flags and commands.
9. Update bindings to expose diagnostic codes and fields instead of requiring clients to scrape logs or rendered footer text.

## Non-goals

- Do not make missing credentials a hard startup error by default. These cases are degraded-but-safe because affected managed-credential requests are denied.
- Do not move CLI styling, theme logic, or CLI-only remediation text into `nono` or `nono-proxy`.
- Do not make log message text part of the API contract.
- Do not remove existing human-readable `Display`/formatter output in a breaking way unless a separate major-version migration is planned.

## Security notes

- Diagnostics must not leak raw credentials, account names, file paths, or URI fields that may be sensitive.
- Fail-closed behavior should be preserved: unavailable route credentials must not cause proxy credential injection to fall back to caller-supplied upstream credentials.
- Hard failures such as invalid configuration or unsafe proxy setup should still return an error rather than being represented only as warnings.
- Structured diagnostics must not weaken existing redaction behavior in `diagnostic.rs`, `scrub.rs`, or keystore URI redaction helpers.

SequeI:
## Linked Issue

Closes #991 

## Summary

Add stable diagnostic codes, remediations, and session reports in nono and nono-proxy; render CLI footers from structured data with --diagnostics-json. Extend C FFI with diagnostic codes and JSON helpers. Refactor diagnostic module and ignore local diagnostics.json output.

## Checklist

### DIFF
--- a/crates/nono-cli/src/diagnostic/formatter.rs
+++ b/crates/nono-cli/src/diagnostic/formatter.rs
@@ -1,4 +1,4 @@
-//! Diagnostic output formatter for sandbox policy (CLI-owned).
+//! Diagnostic footer rendering for sandboxed command failures.
 //!
 //! This module provides human and agent-readable diagnostic output
 //! when sandboxed commands fail. The output helps identify whether
@@ -18,8 +18,13 @@
 //! - **Actionable**: Provides specific flags to grant additional access
 //! - **Mode-aware**: Different guidance for supervised vs standard mode
 
+use nono::SessionDiagnosticReport;
 use nono::diagnostic::{
-    DenialReason, DenialRecord, IpcDenialRecord, SandboxViolation, seatbelt_operation_to_access,
+    DenialReason, DenialRecord, IpcDenialRecord, NonoDiagnostic, NonoDiagnosticCode,
+    NonoDiagnosticDetail, NonoRemediation, SandboxViolation, dedupe_denials,
+    diagnostic_application_failure, diagnostic_likely_sandbox_path, diagnostic_missing_path,
+    diagnostic_network_blocked, diagnostic_protected_file_write,
+    filesystem_denials_from_violations, follow_up_diagnostics,
 };
 use nono::try_canonicalize;
 use nono::{AccessMode, CapabilitySet, CapabilitySource};
@@ -38,8 +43,6 @@ pub struct PolicyExplanation {
     pub access: AccessMode,
     /// Why it was denied: "sensitive_path", "insufficient_access", or "path_not_granted".
     pub reason: String,
-    /// Suggested CLI flag to fix (e.g. "--read ~/.ssh/id_rsa").
-    pub suggested_flag: Option<String>,
 }
 
 /// Path-level hint extracted from a command's own error output.
@@ -75,6 +78,8 @@ pub struct ErrorObservation {
     pub missing_paths: Vec<PathBuf>,
     /// Error text that strongly suggests a non-sandbox application failure.
     pub non_sandbox_failure: Option<String>,
+    /// Stderr suggests network access may have been blocked.
+    pub network_blocked_hint: bool,
 }
 
 impl ErrorObservation {
@@ -85,6 +90,7 @@ impl ErrorObservation {
             || !self.path_hints.is_empty()
             || !self.missing_paths.is_empty()
             || self.non_sandbox_failure.is_some()
+            || self.network_blocked_hint
     }
 }
 
@@ -153,8 +159,12 @@ pub fn analyze_error_output(
     let mut pending_structured_access_denial = false;
     let mut pending_structured_access: Option<AccessMode> = None;
     let mut non_sandbox_failure = None;
+    let mut network_blocked_hint = false;
 
     for line in error_output.lines() {
+        if !network_blocked_hint && looks_like_network_denial(line) {
+            network_blocked_hint = true;
+        }
         if blocked_protected_file.is_none() {
             blocked_protected_file = detect_protected_file_in_error_line(protected_paths, line);
         }
@@ -243,6 +253,7 @@ pub fn analyze_error_output(
         path_hints,
         missing_paths: missing.into_iter().collect(),
         non_sandbox_failure,
+        network_blocked_hint,
     }
 }
 
@@ -284,6 +295,16 @@ fn detect_protected_file_in_error_line(
         }
     }
     None
+}
+
+fn looks_like_network_denial(line: &str) -> bool {
+    let lower = line.to_ascii_lowercase();
+    (lower.contains("network") || lower.contains("socket") || lower.contains("connect"))
+        && (lower.contains("not permitted")
+            || lower.contains("permission denied")
+            || lower.contains("operation not permitted")
+            || lower.contains("connection refused")
+            || lower.contains("unreachable"))
 }
 
 fn looks_like_access_denial(line: &str) -> bool {
@@ -572,10 +593,7 @@ fn infer_access_from_error_line(line: &str, path: &Path) -> AccessMode {
     AccessMode::ReadWrite
 }
 
-/// Formats diagnostic information about sandbox policy.
-///
-/// This is library code that can be used by any parent process
-/// that wants to explain sandbox denials to users or AI agents.
+/// Renders sandbox policy diagnostics for stderr output.
 pub struct DiagnosticFormatter<'a> {
     caps: &'a CapabilitySet,
     mode: DiagnosticMode,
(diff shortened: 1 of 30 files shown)


crates/nono-proxy/src/credential.rs
```diff
@@ -129,19 +150,44 @@ impl CredentialStore {
...
+                                message,
+                            )
+                            .with_credential_ref(redacted)
+                            .with_hint(strip_tip_prefix(&hint)),
                         );
                         continue;
                     }
                     Err(nono::NonoError::KeystoreAccess(msg)) => {
                         let redacted = redact_credential_ref(key);
+                        let message = format!(
```
connrg: tiny thing, not blocking: this message now gets saved/shared but spells out the credential path that credential_ref above deliberately hides. Worth masking here too so they match; same in the  OAuth branches below.  Maybe address another time.

connrg (approved):
LGTM

crates/nono-proxy/src/credential.rs
SequeI: Thanks for the find! This is really valid, fixing it on this PR just to be safe

crates/nono-cli/src/diagnostic/formatter.rs
```diff
@@ -286,6 +297,16 @@ fn detect_protected_file_in_error_line(
     None
 }
 
+fn looks_like_network_denial(line: &str) -> bool {
```
lukehinds, line 300: the string matching approach might end up with nono taking the blame when its not at fault. e.g. we might end up with edge cases, where an agent say, fetches a url , but the  website its connecting to is down, and the agent stderr's with "connection refused." - this might mean nono falsely claims the blame that it blocked the network and then suggest "grant network" - when of course its something outside of nonos control and not a result of a denial from nono policy.

I guess the best way to capture this would be a signal direct from the proxy. But if we don't have that and we need a stderr heuristic as a fallback (for cases when we don't have a hard signal to go off) , we could make the the advice match the mode we are in;

In proxy-only mode, we can suggest allow-listing the host, and not granting all network. Only full-block mode should talk about network at all,, and there the message would be "remove --block-net," since --allow-net is deprecated and conflicts with it.

or even easier just soften the claim. Drop the grant_network remediation and phrase it as a maybe - "this failure might be network-related; check with nono why --host …" - rather than asserting a block and recommending a fix that's might not be correct.

Hope that helps. Nothing major, the majority of the work is top stuff and a real clean refactor - its going to improve maintainability a ton!

lukehinds (approved):

PR #1174, opened 2026-06-16
docs(networking): lead with the common cases

connrg:
Surface the three everyday networking commands (allow a domain, block the net, inject a credential) in a "Common cases" block at the top of the page, they were previously buried below the proxy-modes table. Docs-only.


SequeI (approved):
lgtm, thank you!!

PR #1180, opened 2026-06-17
docs(allow-cwd): clarify access level is profile-driven

connrg:
Ran into this while reading the docs as a new user.

The workflows guide says `--allow-cwd` grants the repo read/write access, but that's only true for `readwrite` profiles. In practice, it grants access at the active profile's `workdir` level:

- `readwrite` for coding profiles (`claude`, `codex`)
- `readonly` with no profile
- `none` under `default`

`flags.mdx` already describes this correctly, so this updates the workflows guide to match and adds a short note to the Working Directory section. Without that context, `--allow-cwd --profile default` can look like it has no effect.

Docs only.


SequeI (approved):
lgtm, thank you!

PR #1199, opened 2026-06-18
refactor(proxy): separate proxy intent from activation

### LINKED ISSUE #1099: (network) separate proxy features from proxy activation

SequeI:
## Linked Issue

Closes #1099

## Summary

Replace the flat `ProxyLaunchOptions` struct with focused intent structs: `DomainFilterIntent`, `EndpointFilterIntent`, `CredentialProxyIntent`, `UpstreamProxyIntent`, `TlsInterceptIntent`, and `OpenUrlIntent`.

Remove the stored `active: bool` field. Proxy activation is now derived via `is_active()`, which returns true when any activating intent struct is `Some`. `WithEndpoints` allow-domain entries are split from plain CONNECT-tunnel entries at prepare time, making the TLS-intercept requirement explicit in the type.

Infra-only flags (`--proxy-port`, `--proxy-ca-validity`, `--trust-proxy-ca`) now return an error if used without an activating proxy feature instead of silently doing nothing.

## Checklist

### DIFF
--- a/crates/nono-cli/src/proxy_runtime.rs
+++ b/crates/nono-cli/src/proxy_runtime.rs
@@ -81,7 +81,7 @@ pub(crate) fn prepare_proxy_launch_options(
 
     let (plain_entries, endpoint_entries): (Vec<_>, Vec<_>) = allow_domain
         .into_iter()
-        .partition(|e| matches!(e, crate::profile::AllowDomainEntry::Plain(_)));
+        .partition(|e| !matches!(e, crate::profile::AllowDomainEntry::WithEndpoints { endpoints, .. } if !endpoints.is_empty()));
 
     let domain_filter = if network_profile.is_some() || !plain_entries.is_empty() {
         Some(DomainFilterIntent {
@@ -96,8 +96,8 @@ pub(crate) fn prepare_proxy_launch_options(
         debug_assert!(
             endpoint_entries
                 .iter()
-                .all(|e| matches!(e, crate::profile::AllowDomainEntry::WithEndpoints { .. })),
-            "EndpointFilterIntent invariant violated: all entries must be WithEndpoints"
+                .all(|e| matches!(e, crate::profile::AllowDomainEntry::WithEndpoints { endpoints, .. } if !endpoints.is_empty())),
+            "EndpointFilterIntent invariant violated: all entries must have non-empty endpoints"
         );
         Some(EndpointFilterIntent {
             routes: endpoint_entries,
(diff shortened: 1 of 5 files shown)


crates/nono-cli/src/proxy_runtime.rs
```diff
@@ -71,38 +72,120 @@ pub(crate) fn prepare_proxy_launch_options(
...
+        return Ok(ProxyLaunchOptions {
+            allow_bind_ports,
+            network_block: prepared.network_block_requested,
+            ..ProxyLaunchOptions::default()
+        });
+    }
+
+    let (plain_entries, endpoint_entries): (Vec<_>, Vec<_>) = allow_domain
+        .into_iter()
+        .partition(|e| matches!(e, crate::profile::AllowDomainEntry::Plain(_)));
```
lukehinds: PRs good - and the following is only something I know from being tripped up before in this area (but mine was worse as it allowed rather then denied).

This let sorts entries by which enum `variant` , but there are actually two ways to write a plain "allow this whole domain" entry in a profile:

1. String form → parses to `Plain`:
`{ "network": { "allow_domain": [ "foo.alwaysfurther.ai" ] } }`

2. Object form → parses to `WithEndpoints`:
```
{ "network": { "allow_domain": [
    { "domain": "foo.alwaysfurther.ai", "endpoints": [ { "method": "GET", "path": "/v1/**" } ] }
] } }
```

Now the gotcha: endpoints has `#[serde(default)]`, so you're allowed to write the object form and just leave em off:

`{ "network": { "allow_domain": [ { "domain": "cdn.example.com" } ] } }`

So you've got two ways to express the exact same intent that produce different internal variants, 

the PR's partition keys on the variant `(matches!(e, Plain(_)))`, so the second form gets misfiled into the endpoint bucket, and then dropped downstream. 

The rest of the codebase avoids this trap by keying on `!endpoints.is_empty()` ("does this entry have path rules?") rather than on which variant it happens to be. 

So I think its as simples as :

`.partition(|e| !matches!(e,
    AllowDomainEntry::WithEndpoints { endpoints, .. } if !endpoints.is_empty()));`

Hope that make sense , the `debug_assert` below can also get patched up with  `!endpoints.is_empty()`

The rest is all good stuff , so will approve once you fix the above as if my musings are correct this would remove a domain from the proxy allowlist (mine was worse, it allowed a non allowed through).

lukehinds (approved):

crates/nono-cli/src/proxy_runtime.rs
SequeI: Ah I see, yea I tested locally and if `endpoints` is empty, the filter has nothing to match against so every request is failed rather than treated as a plain domain and passed through. Good spot!!

PR #1210, opened 2026-06-19
fix(sandbox): exempt IPC fd from sendmsg trapping to resolve af_unix_mediation deadlock

### LINKED ISSUE #1202: How to block docker access on Linux
### Where did you get stuck?

usage, cannot block running docker inside sandbox

### What happened?

Ubuntu 24.04, Kernel 6.17, nono 0.64.0, docker.io 29.1 (standard installation, no rootless)

By using the default profile, access to /etc/passwd is blocked.
When I start docker inside the sandbox, I can bypass the restriction using bind mount.

```
nono run --profile default -- bash

nono v0.64.0
Capabilities:
+ 51 system/group paths (-v to show)
net  outbound allowed

Applying sandbox...

bash: /etc/bash.bashrc: Permission denied
bash: /home/tb/.bashrc: Permission denied
bash-5.2$ cat /etc/passwd | wc -l
cat: /etc/passwd: Permission denied
0
bash-5.2$ docker run --rm -it -v /etc:/mnt/etc alpine:latest sh -c "cat /mnt/etc/passwd | wc -l"
46
bash-5.2$
```

From the documentation (https://nono.sh/docs/cli/features/profiles-groups#blocking-container-access-docker-podman-kubectl):

```
{
  "extends": "default",
  "meta": { "name": "no-docker", "version": "1.0.0" },
  "filesystem": {
    "deny": ["/var/run/docker.sock"]
  },
  "commands": {
    "deny": ["docker", "docker-compose", "podman", "kubectl"]
  }
}
```

from testing, I see that this does not block docker command and connecting to /var/run/docker.sock socket is also possible:

```
nono run --profile test.json -- bash

nono v0.64.0
warning: profile `no-docker` field `commands.deny` is deprecated in v0.33.0: startup-only command gating, not kernel-enforced. Child processes can bypass it. Prefer resource-based controls such as add_deny_access, narrower filesystem grants, unlink_protection, and network policy. Configured commands: docker, docker-compose, podman, kubectl.
Capabilities:
+ 51 system/group paths (-v to show)
net  outbound allowed

Applying sandbox...

bash: /etc/bash.bashrc: Permission denied
bash: /home/tb/.bashrc: Permission denied
bash-5.2$ docker ps
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
bash-5.2$ docker run --rm -it alpine:latest uptime
19:17:06 up 11 min,  0 users,  load average: 0.02, 0.12, 0.08
```

### Error message or output

```shell

```

### How did you install nono?

wget https://github.com/always-further/nono/releases/download/v0.64.0/nono-cli_0.64.0_amd64.deb
sudo dpkg -i nono-cli_*_amd64.deb

### Operating system

Linux

### OS version / distro

Ubuntu 24.04

6.17.0-35-generic

### What documentation or guide were you following?

https://nono.sh/docs/cli/features/profiles-groups#blocking-container-access-docker-podman-kubectl

### Anything else?

What is the preferred way to block docker?

I can remove the user from the docker group, but then access to docker is also blocked outside the sandbox.

### LINKED ISSUE #1208: bug(linux): af_unix_mediation: pathname deadlocks — seccomp filter traps the sendmsg used to hand off the notify fd
## What happens

Setting `linux.af_unix_mediation: pathname` in a profile causes nono to hang indefinitely and never start the sandboxed process. The run must be killed externally.

## Reproduction

```json
{
  "extends": "default",
  "meta": { "name": "repro", "version": "1.0.0" },
  "linux": { "af_unix_mediation": "pathname" }
}
```

```
$ timeout 4 nono run --profile repro.json -- docker ps
  Applying sandbox...
  ^-- hangs, exits 124 (timeout killed it)
```

Tested on: nono 0.64.0, Ubuntu 24.04, kernel 6.8.0-124-generic, Landlock V4.

## Root cause

After fork, the child installs `build_seccomp_af_unix_filter()` and then immediately calls `send_fd_via_socket()` to hand the resulting notify fd to the parent supervisor. That function uses `libc::sendmsg` with `SCM_RIGHTS` — which the filter just installed traps unconditionally via `SECCOMP_RET_USER_NOTIF`.

The child's `sendmsg` blocks waiting for the supervisor to respond to the `USER_NOTIF`. The supervisor is blocked in `recvmsg` waiting to receive the notify fd. Neither can proceed.

`strace` of both processes while hung confirms this:

```
# parent (supervisor):
recvmsg(4, {msg_namelen=0}, 0)   = ? ERESTARTSYS

# child (sandboxed):
sendmsg(5, {msg_name=NULL, msg_namelen=0,
  msg_control=[{cmsg_level=SOL_SOCKET, cmsg_type=SCM_RIGHTS, cmsg_data=[7]}]
}, 0)   = ? ERESTARTSYS  <2.98s>
```

The relevant sequence in `crates/nono-cli/src/exec_strategy.rs` (lines ~935–944):

```rust
nono::sandbox::install_seccomp_af_unix_filter()   // filter now active — sendmsg is trapped
// ...
nono::supervisor::socket::send_fd_via_socket(fd, proxy_notify_fd.as_raw_fd())
// ^ calls libc::sendmsg internally → trapped → USER_NOTIF → nobody home → deadlock
```

The same structural issue exists in `build_seccomp_proxy_filter` (which also traps `sendmsg`), but that path is only reached on Landlock <V4 kernels so it hasn't been triggered in practice.

## Fix direction

The notify fd must reach the parent before the filter becomes active, or the handshake must use a syscall the filter doesn't trap. One approach: `write()` the raw fd number to a pipe (not trapped), then have the parent use `pidfd_getfd()` to duplicate it from the child's namespace, avoiding `SCM_RIGHTS` entirely.

## Note for #1202

This is why `linux.af_unix_mediation: pathname` cannot currently be recommended as the fix for the docker socket escape on Linux. Until this is resolved, the only reliable mitigations are removing the user from the `docker` group or using rootless Docker.

SequeI:
## Linked Issue

Closes #1208
Closes #1202 

@thomas-0816 This will fix an issue on linux with our af unix seccomp filter, blocking docker will be easy then, something like

```
{
  "extends": "claude-code",
  "meta": {
    "name": "no-docker",
    "description": "Claude Code without Docker access"
  },
  "linux": {
    "af_unix_mediation": "pathname"
  }
}
```

This is block all AF UNIX comms within the sandbox, then to specifically allow a sock path, you would

```
{
  "linux": { "af_unix_mediation": "pathname" },
  "unix_sockets": [
    { "path": "/run/user/1000/bus", "ops": ["connect"] }
  ]
}
```

As an example. In your issue, you were using MacOS file system denial. On Linux, a bit different. 

## Summary

The seccomp-notify BPF filters for both af_unix_mediation and the pre-V4 proxy fallback trapped all sendmsg() calls unconditionally. This caused a deadlock: the child installs the filter then immediately calls sendmsg() with SCM_RIGHTS to hand the notify fd to the parent supervisor, which is blocked in recvmsg() — not the seccomp notify loop — so neither side can proceed.

Fix by adding a check-fd block to both filters. When sendmsg() is trapped, args[0] (the sockfd) is loaded from seccomp_data and compared against the known IPC socket pair fd. If it matches, the call is allowed unconditionally (SCM_RIGHTS handshake). All other sendmsg() calls and all connect/bind/ sendto/sendmmsg calls route to USER_NOTIF as before.

Also fix a secondary bug introduced during the check-fd block insertion: the jt offsets for connect/bind/sendto/sendmmsg in the af_unix filter were pointing at the wrong instruction after USER_NOTIF moved from insn 7 to insn 9, causing those syscalls to be silently allowed instead of trapped.

Add end-to-end regression tests that spawn a real nono process and verify AF_UNIX pathname connect() is blocked (Linux: af_unix_mediation; macOS: filesystem.deny) and allowed when the path is on the unix_sockets allowlist.

Update profile-authoring-guide.md with platform-specific guidance for blocking container socket access.

## Checklist

### DIFF
--- a/crates/nono/src/sandbox/linux.rs
+++ b/crates/nono/src/sandbox/linux.rs
@@ -1853,6 +1853,8 @@ fn build_seccomp_proxy_filter(_has_bind_ports: bool) -> Vec<SockFilterInsn> {
     let bind_action = SECCOMP_RET_USER_NOTIF;
 
     // sendto(), sendmsg(), and sendmmsg() route unconditionally to USER_NOTIF.
+    // The IPC handshake completes before this filter is installed, so no
+    // fd-based exemption is needed.
     // The supervisor does the full-width NULL destination checks and inspects
     // each sendmmsg vector entry.
     //
@@ -2046,25 +2048,29 @@ fn build_seccomp_proxy_filter(_has_bind_ports: bool) -> Vec<SockFilterInsn> {
     ]
 }
 
-/// Build a BPF filter for opt-in pathname AF_UNIX mediation.
+/// Build a BPF filter for opt-in pathname AF_UNIX mediation (Linux-only).
 ///
-/// The filter routes `connect()`, `bind()`, `sendto()`, `sendmsg()`, and `sendmmsg()`
-/// to the supervisor so it can inspect `sockaddr_un` paths. Everything
-/// else is allowed by this filter: TCP policy remains Landlock's job on
-/// V4+ kernels.
+/// Routes `connect()`, `bind()`, `sendto()`, `sendmsg()`, and `sendmmsg()`
+/// to `USER_NOTIF` so the supervisor can inspect `sockaddr_un` paths.
+/// Everything else is allowed: TCP policy is Landlock's responsibility on
+/// V4+ kernels; this filter only handles AF_UNIX.
 ///
-/// The send syscalls route unconditionally. BPF cannot dereference
-/// `msghdr`/`mmsghdr`, and checking only half of a 64-bit `sendto` pointer
-/// is not a reliable NULL test.
+/// The IPC handshake (child→parent SCM_RIGHTS transfer of the notify fd)
+/// must complete *before* this filter is installed. That ordering removes
+/// the need for any fd-based exemption, so the filter is a pure allowlist
+/// with no internal plumbing holes.
+///
+/// BPF cannot dereference `msghdr`/`mmsghdr`; the supervisor performs those
+/// checks instead.
 ///
 /// Instruction layout:
 /// ```text
 ///  0: ld  [nr]
-///  1: jeq SYS_CONNECT  jt=+5 (-> 7: notify)
-///  2: jeq SYS_BIND     jt=+4 (-> 7: notify)
-///  3: jeq SYS_SENDTO   jt=+3 (-> 7: notify)
-///  4: jeq SYS_SENDMSG  jt=+2 (-> 7: notify)
-///  5: jeq SYS_SENDMMSG jt=+1 (-> 7: notify)
+///  1: jeq SYS_CONNECT  jt=+5 (->  7: notify)
+///  2: jeq SYS_BIND     jt=+4 (->  7: notify)
+///  3: jeq SYS_SENDTO   jt=+3 (->  7: notify)
+///  4: jeq SYS_SENDMSG  jt=+2 (->  7: notify)
+///  5: jeq SYS_SENDMMSG jt=+1 (->  7: notify)
 ///  6: ret ALLOW
 ///  7: ret USER_NOTIF
 /// ```
@@ -2129,13 +2135,19 @@ fn build_seccomp_af_unix_filter() -> Vec<SockFilterInsn> {
     ]
 }
 
-/// Install a seccomp-notify BPF filter for proxy-only network mode.
+/// Install a seccomp-notify BPF filter for proxy-only network mode (Linux-only).
 ///
-/// Returns the notify fd that the supervisor must poll for connect/bind
-/// notifications. Uses `SECCOMP_FILTER_FLAG_NEW_LISTENER`.
+/// Used on kernels without Landlock V4 TCP support as a fallback: traps
+/// `connect()`, `bind()`, `sendto()`, `sendmmsg()`, and `sendmsg()` for
+/// supervisor mediation. Returns the notify fd the supervisor must poll.
+/// Uses `SECCOMP_FILTER_FLAG_NEW_LISTENER`.
 ///
-/// Must be called AFTER `PR_SET_NO_NEW_PRIVS` is already set (either by
-/// a prior seccomp install or by Landlock's `restrict_self()`).
+/// The IPC handshake (SCM_RIGHTS transfer of the notify fd to the parent)
+/// must complete before this filter is installed so no fd-based exemption
+/// is needed.
+///
+/// Must be called after `PR_SET_NO_NEW_PRIVS` is set (either by a prior
+/// seccomp install or by Landlock's `restrict_self()`).
 ///
 /// # Errors
 ///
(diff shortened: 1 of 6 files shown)


thomas-0816:
Test ok. Ubuntu 24.04

rustc --version
rustc 1.96.0 (ac68faa20 2026-05-25)

git clone --depth=1 --branch=fixAF_UNIX https://github.com/always-further/nono.git
cd nono
cargo build --release -p nono-cli

cat > test.json << EOF
{
  "extends": "default",
  "meta": { "name": "no-docker", "version": "1.0.0" },
  "linux": { "af_unix_mediation": "pathname" }
}
EOF

./target/release/nono run --profile test.json -- docker ps

/*
Command exited with code 1.

IPC denial: 5 Unix socket operations blocked.
connect /run/docker.sock (no matching unix_socket capability)
connect /run/docker.sock (no matching unix_socket capability)
connect /run/docker.sock (no matching unix_socket capability)
connect /run/docker.sock (no matching unix_socket capability)
connect /run/docker.sock (no matching unix_socket capability)
Fix flags: --allow-unix-socket /run/docker.sock
*/

thomas-0816:
Opencode + llama.cpp test ok.

```
{
  "extends": "default",
  "groups": { "include": ["user_caches_linux", "linux_sysfs_read", "unlink_protection", "opencode_linux"] },
  "security": { "signal_mode": "isolated", "capability_elevation": false },
  "filesystem": {
    "allow": ["$TMPDIR", "$HOME/.opencode", "$HOME/.config/opencode", "$HOME/.cache/opencode",
      "$HOME/.local/share/opencode", "$HOME/.local/share/opentui", "$HOME/.local/state/opencode"]
  },
  "network": { "block": false, "open_port": [8080] },
  "linux": { "af_unix_mediation": "pathname" },
  "interactive": true
}
```

Screenshot_20260619_205705

Screenshot_20260619_210200

thomas-0816:
using the example given:

```
{
  "linux": { "af_unix_mediation": "pathname" },
  "unix_sockets": [
    { "path": "/run/user/1000/bus", "ops": ["connect"] }
  ]
}
```

gives:

`nono: Profile parse error: unknown field `unix_sockets`, expected one of `$schema`, `extends`, `meta`, `security`, `groups`, `commands`, `filesystem`, `policy`, `network`, `diagnostics`, `linux`, `env_credentials`, `secrets`, `environment`, `workdir`, `hooks`, `session_hooks`, `rollback`, `undo`, `open_urls`, `allow_launch_services`, `allow_gpu`, `allow_parent_of_protected`, `interactive`, `skipdirs`, `packs`, `binary`, `brokered_commands`, `command_args`, `unsafe_macos_seatbelt_rules` on line 8 column 1`

SequeI:
Ooops, sorry I missed a key in the example

```
{
  "linux": { "af_unix_mediation": "pathname" },
  "filesystem": {
    "unix_socket": ["/run/user/1000/bus"]
  }
}
```

crates/nono/src/sandbox/linux.rs
```diff
@@ -2037,38 +2088,91 @@ fn build_seccomp_af_unix_filter() -> Vec<SockFilterInsn> {
...
+            k: SECCOMP_DATA_ARG0_OFFSET,
+        },
+        // 8: jeq ipc_fd jt=+1 (-> 10: allow) jf=0 (-> 9: notif)
+        // jf=0: not the ipc fd → fall to insn 9 (USER_NOTIF) for mediation.
+        // jt=1: is the ipc fd → skip insn 9, land on insn 10 (ALLOW).
+        SockFilterInsn {
+            code: BPF_JMP | BPF_JEQ | BPF_K,
+            jt: 1,
+            jf: 0,
+            k: ipc_fd as u32,
```
connrg: Would it be worth adding a test that a sandboxed program can't reach an unlisted socket by reassigning its fds (e.g. moving one onto the trusted slot, then sending)? Just to confirm the exemption holds up there.

SequeI: Valid, reworked the PR a bit so that the sandboxed program is allowlist only, no weird filtering based on the sendmsg() to the supervisor. All that is now done before the sandbox ensuring nothing can be intercepted, and we only pass though items on the allow list.

Would appreciate a test run on Linux just to make sure the behaviour is correct. I tested on a lima box.

connrg: the feature itself works great on Fedora. The only issue I hit is with the new `af_unix_mediation_pathname_allows_connect_to_listed_socket` test in `crates/nono-cli/tests/socket_access_run.rs`: the `!stderr.contains("unix_socket")` check (around line 174) is a bit too broad, so on a systemd host it can match the unrelated `/run/systemd/journal/socket` denial rather than the test socket we're actually exercising. Scoping the assertion to the test socket path should make the test reliable across environments.

connrg (approved):
LGTM

lukehinds (approved):
looks good.

PR #1212, opened 2026-06-19
chore(docs): improve profile documentation

### LINKED ISSUE #1206: Docs: Explain how extends works

When authoring profiles one usually uses `extends` to extend some or multiple built-in profiles.
However the docs don't go into detail how the different profiles are combined.
There are only a few scattered comments afaics.

There should be a dedicated section in the docs, maybe relatively at the top of "Profile Authoring", that explain the extending and merging of profiles in more detail.

In particular I would like to understand:

- Are individual options replaced at the top-level or merged deeply?
- Do my own settings always override the "imported" ones?
- If there are conflicting settings between the extended profiles, which one takes precedent?
- What settings need special handling for overriding them.
- What is the default profile and what is the impact of extending it or not?
- Can I run into issues due to diamond-extension? I.e. I extend profiles A and B, which both extend profile C and all of them mabye also extend the standard profile.

Maybe frame those details with some high-level practical guidance on how to use the `extends`, so understanding without all nitty-gritty details is also possible.

Until now I have just winged it, but I don't feel confortable with that approach anymore, after seeing what the predefined pi profile sets and thinking about what I want to override there.

### How is this blocking you?

### Additional context

_No response_

SequeI:
## Linked Issue

Closes #1206

## Summary

@bodograumann please let me know if this covers the bulk of the questions, or if the wording is off, anything missing etc. Thank you!!!

## Checklist

### DIFF
--- a/docs/cli/features/profile-authoring.mdx
+++ b/docs/cli/features/profile-authoring.mdx
@@ -63,7 +63,7 @@ A few fields have their own specific rules:
 | `open_urls` | Your value replaces the base entirely. Leave it out to inherit the base value. |
 | `network.credentials` | Leave it out to inherit. Set it to `[]` to explicitly disable inherited credentials. Otherwise your list is combined with the base. |
 | `network.network_profile` | Leave it out to inherit. Set it to `null` to explicitly clear the inherited proxy profile. |
-| `hooks.hooks`, `env_credentials.mappings`, `environment.set_vars` | Merged as key-value maps. Your keys win over the base on any conflict. |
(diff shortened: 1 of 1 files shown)


lukehinds (approved):
looks good! nothing to block over in anyway, but if you feel like another commit you could mention packages can be extended `always-further/claude` - and not a now issue, but when tool sandboxing comes in we might want to mention how the interplay there - but as said, that does not gate this PR.

SequeI:
@lukehinds Sounds good, I bet once Tool sandboxing is out, we will need a good bit of doc changes so we can catch that one there in one swoop

bodograumann:
Yes, this is fantastic <3

PR #1213, opened 2026-06-19
feat(tests): add end-to-end integration tests for sandbox execution strategies

### LINKED ISSUE #1211: integration: add end-to-end execution tests for core sandbox features
## Problem

The current test suite has strong unit coverage for individual modules (policy resolution, BPF filter structure, capability building, diagnostic formatting) but almost no integration tests that actually run `nono run` and verify sandbox enforcement end-to-end.

This gap allowed two bugs to reach main undetected:
- A deadlock in `af_unix_mediation` caused by the seccomp filter trapping the IPC `sendmsg` handshake — the BPF instruction layout was unit-tested but the full install → fork → send_fd → exec → connect sequence was never run
- A BPF jt-offset bug that silently allowed `connect()` through instead of routing it to USER_NOTIF — the instruction count was asserted but the actual jump targets were not verified by a running process

## What's missing

The critical modules have almost no integration coverage:

| Module | Lines | Tests | Coverage |
|---|---|---|---|
| `execution_runtime.rs` | 565 | 3 | ~0.5% |
| `supervised_runtime.rs` | 331 | 2 | ~0.6% |
| `rollback_runtime.rs` | 904 | 10 | ~1.1% |
| Monitor exec strategy | — | 0 | 0% |

None of these have tests that actually run `nono run` with a real child process under a real sandbox and verify the outcome.

## What good looks like

Tests in the style of `deny_overlap_run.rs` — spawn a real nono process, use a known inline profile, attempt a specific operation, assert the result:

```rust
// filesystem: Landlock blocks read outside granted paths
nono run --profile minimal -- cat /etc/shadow  // must exit non-zero

// network: af_unix_mediation blocks unlisted socket connect
nono run --profile af-unix-deny -- python3 -c "..."  // must show IPC denial

// rollback: snapshot → modify → rollback → verify restored
nono run --profile rollback -- bash -c "touch /workdir/foo"
nono rollback  // foo must be gone
```

## Proposed scope

- One integration test per exec strategy (Direct, Monitor, Supervised) verifying the sandbox actually applies
- Filesystem deny: attempt to read a path outside the grant, assert denial and diagnostic footer  
- AF_UNIX mediation: block and allow cases (Linux), `filesystem.deny` socket case (macOS)
- Rollback: baseline → modify → restore, verify file state
- Network: connect attempt to a non-proxy address, assert denial

These tests would live in `crates/nono-cli/tests/` alongside `deny_overlap_run.rs` and `socket_access_run.rs`, using platform `cfg` gates where needed. Profiles should be inline JSON with no `extends` dependency so tests are hermetic.

## Related

- #1208 — af_unix_mediation deadlock (would have been caught by a seccomp IPC handshake completion test)
- #1202 — docker socket escape on Linux (would have been caught by an af_unix_mediation block test)

mvanhorn:
## Summary

- Add `execution_strategy_run.rs`: end-to-end integration tests that spawn the `nono` binary with inline hermetic profiles and verify that the Direct and Supervised sandbox execution strategies enforce filesystem grant boundaries. Tests cover the positive case (allowed path exits zero), the negative case (denied path returns non-zero with no content leak), and a platform-cfg-gated Linux variant for the Supervised strategy.
- Add `rollback_run.rs`: integration tests that exercise the baseline-snapshot → write → rollback flow using `nono run --rollback`, `nono rollback list`, and `nono rollback cleanup --dry-run` as subprocesses. Rollback state is isolated to a temp directory via `--rollback-dest` and `XDG_STATE_HOME`.

Both files follow the `env!("CARGO_BIN_EXE_nono")` + `tempfile` pattern used by the existing `deny_overlap_run.rs` and `config_flag.rs` tests. No `extends` dependencies — profiles are inline JSON so tests are hermetic.

## Why this matters

Closes #1211. The issue identified near-zero integration coverage of `execution_runtime.rs`, `supervised_runtime.rs`, and `rollback_runtime.rs` — the three modules where the two recent regressions (af_unix deadlock and BPF jt-offset bypass) would have been caught by tests like these. The new tests give the sandbox enforcement path a direct exercise path without requiring a full CI environment: non-integration assertions (positive control, rollback list/cleanup) run everywhere; the Linux-specific sandbox denial tests are gated `#[cfg(target_os = "linux")]`.

## Testing

Tests compile and pass `cargo test --no-run` on macOS (Rust 1.95). The Linux-gated sandbox denial tests (`direct_strategy_denies_path_outside_grant`, `supervised_strategy_denies_path_outside_grant`, `filesystem_deny_boundary_produces_diagnostic_footer`) require a Linux host with Landlock-capable kernel to execute. The macOS-gated Seatbelt test and the platform-neutral tests (`allowed_path_exits_zero`, rollback cleanup dry-run) run on any OS.

CI: tests compile; full sandbox enforcement tests need Linux.

Fixes #1211


SequeI:
This is amazing and very much needed! Thanks for this PR so soon after the issue opened, always good to get more resiliency into the project. I basically agree with the gemini comments, + if we could use less of the ASCII banners for test grouping and be more concise; would be amazing.

After addressed, I shall do a full review :)  + DCO

mvanhorn:
The macOS test failure was a test-premise bug, not a sandbox regression. `direct_strategy_denies_path_outside_grant_macos` read `/etc/passwd`, but Seatbelt grants the sandboxed process a set of system read paths (so it can exec and load dyld), and `/etc/passwd` is a world-readable system file inside that set, so the read succeeds and the assertion fails. Reproduced locally: the capability footer shows the workspace grant plus "34 system/group paths", and `/etc/passwd` printed.

Pushed `2636505e`: the test now reads a secret file created in the temp root (a sibling of the granted workspace, neither granted nor a system path), so it exercises a genuine out-of-grant denial and mirrors the Linux variant's intent without depending on a system path.

lukehinds:
from what I know /etc/passwd is just a relic not even used anymore by the later versions of mac, I think its just a placeholder to stop other systems tripping up over its absence. PR looks good! will review shortly, thanks for contributing.

SequeI (approved):
lgtm!

PR #1219, opened 2026-06-21
chore(docs): update README with curl installation method

lukehinds:
Add installation instructions using curl. I was in two minds about doing this for a while, but so many users expect to install this way now, so we should make it simple for them. 

The script handles all operating systems and is here: https://github.com/always-further/nono-website/blob/main/public/install.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ From here **fork the config**, tweak it, theme it, make it your own, and share i
 
 ## Quickstart
 
+#### curl
+
+```bash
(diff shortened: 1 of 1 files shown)


SequeI (approved):
In a perfect world, Gemini is correct however for conformance and not to be a standout, the curl -> pipe into sh is better. Everyone does it this way, so imo this is good to merge

lukehinds:
@SequeI gemini 
image

PR #1261, opened 2026-06-25
feat(cli): add standalone `nono proxy` command

### LINKED ISSUE #1260: Standalone `nono proxy` command

We want to use nono's network-layer protections, **domain allowlisting** and **credential injection**, for workloads that provide their own OS-level isolation: Docker containers, Docker sandboxes (sbx), Kata Containers, Kubernetes pods, Firecracker and gVisor, or really any workload that needs a **central credential broker** sitting in front of its egress.

The two capabilities we need from nono here are:

1. **Domain allowlisting** — restrict the workload's outbound HTTP(S) egress to a known set of hosts/endpoints.
2. **Credential injection** — inject credentials into outbound requests via the reverse proxy so the AI tool inside the guest never sees the raw secret.

Today the proxy only starts as a *sidecar* of `nono run` / `shell` / `wrap`: it spins up the filtering + credential-injection proxy, wires `HTTP(S)_PROXY` and the session token into the sandboxed child, and tears it down when the child exits. There is no supported way to run just the proxy as a standalone, long-lived server that an external process, can point its tools at.

This restricts nono to runtimes where `nono run` itself works. `nono run` is **not supported in Firecracker or gVisor today**. In all of those cases we still want nono's allowlist + credential brokering.

Concrete use case I have: **Docker Desktop proxying through nono**, so I can safely let an AI use my Docker daemon, egress is allowlisted and credentials are brokered by nono instead of being handed to the AI tool. (Integrating that wiring directly into nono is a possible follow-up, but the standalone proxy is the primitive that unblocks it.)

A standalone **`nono proxy`** command that runs the network-filtering / credential-injection proxy as a foreground server with the following features:

- Reuse the exact same config-building machinery as the sandboxed path (allow-domain parsing, endpoint/TLS-intercept handling, upstream-proxy chaining) so behavior matches `run`/`shell`/`wrap`.
- Load network + credential settings from a **profile** (`--profile`), with explicit flags extending (allow-domains, credentials) or overriding (network profile, upstream proxy) profile values.
- Print connection details for the user to point their own tools at — proxy URL, token, `HTTP(S)_PROXY` exports, per-route diagnostics, and the TLS-interception trust bundle path — then block until Ctrl-C.
- Support credential injection for known services **and** `cmd://` credential-capture routes, so captured credentials are injected the same way as on the sandboxed path. This is what keeps secrets out of reach of the AI tool in the guest.

Proposed surface:

```
nono proxy                                   # Ephemeral port, token auth, prints connection info
nono proxy --port 8080                       # Fixed port
nono proxy --port 8080 --no-auth             # Open loopback proxy (no token required)
nono proxy --allow-domain github.com         # Allowlist a host
nono proxy --profile my-profile              # Load network/credential settings from a profile
```

Flags: `--listen`, `--port`, `--no-auth`, `--pass`, `--profile`, `--network-profile`, `--allow-domain`, `--upstream-proxy`, `--upstream-bypass`, `--proxy-ca-validity`, `--trust-proxy-ca` (macOS), `--credential`.

Security notes (fail-secure behavior this command must keep):

- `--no-auth` must be **refused for non-loopback bind addresses** so we never start an open proxy reachable from other hosts.
- `--pass` must reject an empty password so it can't collapse to an effectively-absent secret; `--no-auth` and `--pass` are mutually exclusive.
- `--upstream-bypass` requires `--upstream-proxy` (otherwise the bypass entries are silently dropped).

- Running `nono run` / `shell` with a dummy long-lived child purely to keep the proxy alive — fragile, ties proxy lifecycle to an unrelated process, and doesn't fit runtimes where nono isn't (or can't be) the sandboxing parent.
- Running the workload under `nono run` directly — not supported on Firecracker / gVisor, and redundant where a container already provide the required isolation.
- A generic forward proxy (mitmproxy / Squid) in front of the guest — gets us allowlisting but not nono's profile-driven credential injection / `cmd://` capture, so the AI tool still ends up handling raw secrets.

### How is this blocking you?

### Additional context

_No response_

panga:
## Linked Issue

Closes #1260

## Summary

Adds a standalone `nono proxy` command that runs the network-filtering / credential-injection proxy as a foreground server with **no sandboxed child**. Unlike `run` / `shell` / `wrap` — which start the proxy as a side effect and wire its env vars into the sandboxed process — this command prints the connection details (proxy URL, token, env vars, per-route diagnostics, TLS trust bundle path) for the user to point their own tools at, then blocks until Ctrl-C.

This unblocks using nono's egress allowlisting and credential brokering for workloads that provide their own OS-level isolation: Docker / Docker sandboxes, Kata Containers, Kubernetes pods, Firecracker, and gVisor. The guest routes its egress through the proxy via `HTTP(S)_PROXY` while OS-level isolation comes from the container/microVM runtime.

Proxy settings are loaded from a profile (`--profile`) and extended/overridden by explicit flags, reusing the same config-building machinery as the sandboxed path, allow-domain parsing, endpoint/TLS-intercept handling, upstream-proxy chaining) so behavior matches `run`/`shell`/`wrap`. Credential injection for known services and `cmd://` credential-capture routes (with `command_policies` approvals) carries through the profile the same way.

Usage:

```
nono proxy                                   # Ephemeral port, token auth, prints connection info
nono proxy --port 8080                       # Fixed port
nono proxy --port 8080 --no-auth             # Open loopback proxy (no token required)
nono proxy --allow-domain github.com         # Allowlist a host
nono proxy --profile my-profile              # Load network/credential settings from a profile
```

### Fail-secure behavior

- `--no-auth` is **refused for non-loopback bind addresses**, so we never start an open proxy reachable from other hosts.
- `--pass` rejects an empty password so it can't collapse to an effectively-absent secret; `--no-auth` and `--pass` are mutually exclusive (enforced at the clap layer). `--pass` also reads from `NONO_PROXY_PASS` to avoid leaking the secret in shell history / process listings.
- `--upstream-bypass` requires `--upstream-proxy` (otherwise the bypass entries would be silently dropped), mirroring the validation on the sandboxed path.
- The session token is held in `zeroize::Zeroizing`.

### Key changes

- `crates/nono-cli/src/proxy_command.rs` (new) — the command: profile + flag merge into `ProxyLaunchOptions`, fail-secure validation, runtime startup, and connection-info printing.
- `crates/nono-cli/src/cli.rs` — `ProxyArgs` and the `proxy` subcommand with `--listen`, `--port`, `--no-auth`, `--pass`, `--profile`, `--network-profile`, `--allow-domain`, `--upstream-proxy`, `--upstream-bypass`, `--proxy-ca-validity`, `--trust-proxy-ca` (macOS), `--credential`.
- `crates/nono-cli/src/{main,app_runtime,cli_bootstrap}.rs` — dispatch wiring for the new subcommand.
- `crates/nono-cli/src/proxy_runtime.rs` — refactors to share config-building, credential-capture backend, and TLS-intercept wiring between the sandboxed and standalone paths.
- `crates/nono-proxy/src/{config,server,token,reverse,connect,external}.rs` — supporting changes to start the proxy without a sandboxed child and surface route diagnostics / the bound port.

## Test Plan

- `make ci` (clippy `-D warnings -D clippy::unwrap_used`, fmt check, tests).
- Unit tests in `proxy_command.rs` cover the flag/profile merge and fail-secure validation:
  - `--upstream-bypass` without `--upstream-proxy` is rejected; with it, the bypass list carries through.
  - upstream proxy alone carries through; no upstream flags yields no upstream proxy.
  - no profile yields empty credential-capture but a minted session id; a profile with `credential_capture` carries the entry through.
  - Tests clear `NONO_*` env vars under `ENV_LOCK` to stay hermetic.
- Manual verification:
  - `nono proxy` → connects on an ephemeral port with token auth; printed `HTTPS_PROXY` URL works with `curl`.
  - `nono proxy --no-auth --listen 0.0.0.0` → refused (non-loopback without auth).
  - `nono proxy --profile <p>` with allow-domains + credentials → allowlist enforced and credentials injected for an external client.

## Checklist


crates/nono-cli/src/proxy_command.rs
```diff
@@ -0,0 +1,606 @@
...
+        )));
+    }
+
+    let proxy = build_launch_options(&args)?;
+    let mut proxy_config = build_proxy_config_from_flags(&proxy)?;
+
+    // Bind + auth settings come from the standalone flags, not the profile.
+    proxy_config.bind_addr = args.listen;
+    proxy_config.bind_port = args.port;
+    proxy_config.require_auth = !args.no_auth;
```
lukehinds, line 43: If I’m reading this correctly, this still makes auth non-fatal even when require_auth is true. The new standalone nono proxy relies on require_auth = true as its default security
boundary, but this branch only logs missing or invalid Proxy-Authorization and then continues to DNS, filter, and upstream handling:

if require_auth && let Err(e) = validate_proxy_auth(remaining_header, session_token) {
debug!("CONNECT auth skipped: {}", e);
}

As a result, unauthenticated transparent CONNECT requests could use the proxy instead of receiving 407 Proxy Authentication Required. For standalone mode, auth needs to be mandatory unless --no-auth was explicitly set.

Hope that makes sense. The rest is looking strong.

lukehinds (commented):
just one potential area to lean into more.

crates/nono-cli/src/proxy_command.rs
panga: Good catch on the non-fatal CONNECT auth. After digging in, I think we can align on this being **working as intended** and not something we should change right now, so I've left the CONNECT path as-is.

A few things to note:

1. **It's pre-existing, not introduced here.** `connect::handle_connect`'s
   lenient auth predates this PR; this branch only added the
   `require_auth &&` guard to reduce log noise under `--no-auth`. The non-fatal
   behavior is intentional: Node.js undici doesn't echo URL-userinfo credentials
   as `Proxy-Authorization` on CONNECT, so making that a hard 407 would break
   legitimate sandboxed clients.

2. The host allowlist still applies on the CONNECT path (`filter.check_host`),
   so even an unauthenticated tunnel remains constrained to allow-listed egress.
   In that sense, the session token serves as a *localhost* boundary, not the
   egress boundary.

3. Making CONNECT auth fatal only for standalone mode would introduce a
   divergence in this PR by changing CONNECT behavior between `nono run`
   and `nono proxy`. I think it would be better for us to track that separately.

crates/nono-proxy/src/connect.rs
```diff
@@ -43,7 +44,11 @@ pub async fn handle_connect(
     // Validate session token from Proxy-Authorization header.
     // Non-fatal for CONNECT: Node.js undici doesn't send Proxy-Authorization
     // from URL userinfo for CONNECT requests.
-    if let Err(e) = validate_proxy_auth(remaining_header, session_token) {
+    //
+    // Skipped entirely when auth is disabled (`nono proxy --no-auth`): there
+    // is no token to check, so running the validation only produces noisy
+    // "Missing Proxy-Authorization header" / "CONNECT auth skipped" debug logs.
+    if require_auth && let Err(e) = validate_proxy_auth(remaining_header, session_token) {
```
lukehinds: Thanks for the update Leo! This is real close now. I think the reverse-proxy auth path is fixed and lgtm, but the standalone transparent `CONNECT` path still has the same gap. In the normal CONNECT branch with no external proxy, the request goes into `connect::handle_connect(...);` inside that handler, `require_auth = true` only decides whether validation is attempted, but validation failure is still logged and ignored. I can still reproduce this on the updated head with an unauthenticated CONNECT nonexistent.invalid:443: it returns 502 Upstream connect failed..., which means the request reached DNS/upstream handling instead of being stopped with 407 (I think that's the code, might need to check / look up) - **Proxy Authentication Required**. For standalone mode, this path still needs strict auth before DNS/ filter/upstream handling unless `--no-auth` was explicitly set.

apols for being sticky here and giving you iterations of effort, but I have stung myself by this area with my own code before, so want to help capture these early for you

This is via codex , so not checked it yet, but it should be something like this:

```rust
if require_auth && let Err(e) = token::validate_proxy_auth(remaining_header, session_token) {
    if strict_connect_auth {
        debug!("CONNECT auth failed: {}", e);
        stream
            .write_all(
                b"HTTP/1.1 407 Proxy Authentication Required\r\n\
                  Proxy-Authenticate: Basic realm=\"nono\"\r\n\
                  Content-Length: 0\r\n\r\n",
            )
            .await?;
        return Ok(());
    }

    debug!("CONNECT auth skipped: {}", e);
```

Hope that helps! It's an awesome addition, so want to see this one ship!

lukehinds (commented):
Just one more Leo, appreciate your efforts and apols for the iterations.

crates/nono-proxy/src/connect.rs
panga: Resolved in https://github.com/nolabs-ai/nono/pull/1261/changes/47e194ae49ce4f6f7ecc523b27401535570ed918

Please note this introduces divergent behavior between `nono proxy` and `nono run` which does not validate credentials on the `CONNECT` path. You can reproduce this with `HTTPS_PROXY=http://127.0.0.1:<port> curl -v https://<allow_domain>`.

lukehinds:
Going to try to land this, apols for taking some time , just done traveling

lukehinds (approved):

PR #1262, opened 2026-06-25
feat(network): add open_port_range and fix Linux ProxyOnly semantics

### LINKED ISSUE #611: Allow domain filtering and unknown ports on localhost

### LINKED ISSUE #1101: (network)  fix Linux ProxyOnly semantics
`ProxyOnly` means connect only to loopback:<proxy-port> on every platform.

Always use seccomp-notify for ProxyOnly on Linux, even with Landlock V4+

SequeI:
## Linked Issue

Closes #1101 
Closes #611 

## Summary

- ProxyOnly on Linux now always uses seccomp-notify regardless of Landlock ABI; Landlock TCP rules have no IP component and cannot enforce loopback-only
- Add open_port_range to profiles and CapabilitySet for localhost IPC port ranges; macOS expands ≤256 ports to individual Seatbelt rules, wider ranges collapse to localhost:* with a warning; Linux expands to individual Landlock rules in both block-net and proxy mode
- Add open_port=0 wildcard in proxy mode: allows any loopback connect/bind without knowing the port at profile-write time (fixes ephemeral port use cases like Testcontainers and Maven Surefire)
- open_port=0 in block-net mode on Linux errors at startup (no supervisor)
- Document all network mode and port behaviour including platform differences

## Checklist

### DIFF
--- a/crates/nono-cli/src/exec_strategy/supervisor_linux.rs
+++ b/crates/nono-cli/src/exec_strategy/supervisor_linux.rs
@@ -563,9 +563,13 @@ pub(super) enum NetworkDecision {
 ///    no path to check.
 ///
 /// 2. For `AF_INET`/`AF_INET6`:
-///    - `connect()` is allowed only to `127.0.0.1:proxy_port` (the nono proxy).
-///    - `bind()` is allowed only on ports in `proxy_bind_ports`.
+///    - `connect()` is allowed to `127.0.0.1:proxy_port` (the nono proxy),
+///      any port in `localhost_ports` (port `0` = wildcard), or any port
+///      that falls within a `localhost_port_ranges` range — **loopback only**.
+///    - `bind()` is allowed on `proxy_bind_ports`, `localhost_ports`, or
+///      `localhost_port_ranges`. No loopback check on bind (matches macOS).
 ///    - Everything else is denied.
+///
 pub(super) fn decide_network_notification(
     child_pid: u32,
     syscall: i32,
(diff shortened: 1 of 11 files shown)


crates/nono-cli/src/exec_strategy/supervisor_linux.rs
```diff
@@ -633,8 +644,15 @@ pub(super) fn decide_network_notification(
             }
         }
         SYS_BIND => {
-            // Allow bind only on configured bind ports
-            if config.proxy_bind_ports.contains(&sockaddr.port) {
+            // Allow configured bind ports, localhost_ports wildcard (0), or exact match.
+            let allowed = config.proxy_bind_ports.contains(&sockaddr.port)
```
connrg, line 648: The connect path only allows localhost, but bind doesn't seem to — so with open_port=0 it looks like the sandboxed process could open a port that's reachable from the outside network, not just localhost. Is that intended (e.g. for Testcontainers), or should bind be limited to localhost too?

crates/nono/src/sandbox/macos.rs
```diff
@@ -476,6 +477,33 @@ fn push_localhost_tcp_outbound_seatbelt_rules(profile: &mut String, localhost_po
...
+/// Max ports expanded as individual Seatbelt rules. Wider ranges collapse to
+/// `localhost:*` — Seatbelt has no range syntax.
+const MACOS_PORT_RANGE_EXPAND_LIMIT: u32 = 256;
+
+/// Emit Seatbelt outbound rules for `ranges`. Ranges within the limit expand
+/// to one rule per port; wider ranges collapse to `localhost:*` with a warning.
+fn push_localhost_port_range_seatbelt_rules(profile: &mut String, ranges: &[(u16, u16)]) {
+    for &(start, end) in ranges {
+        let width = (end as u32).saturating_sub(start as u32).saturating_add(1);
+        if width > MACOS_PORT_RANGE_EXPAND_LIMIT {
```
connrg, line 489: Looks like on macOS a range bigger than 256 ports gets turned into "allow all localhost ports", while on Linux it stays limited to exactly the ports listed. So the same profile ends up more open on macOS than the user asked for. Would it be safer to just reject a range that's too big, rather than quietly opening up everything?

connrg (approved):
LGTM

PR #1267, opened 2026-06-26
feat(credential): oauth-capture routes with phantom-token broker

### LINKED ISSUE #1265: feat(credential): credential mediation for OAuth-capture routes

A sandboxed agent (e.g. Claude Code) needs to reach the Anthropic API without ever holding a real bearer credential, even though the agent has subprocess and keychain-read access.

OAuth `/login` inside the sandbox. Real access/refresh tokens returned by `/v1/oauth/token` must be swapped for `nono_<hex>` nonces before the response body reaches the agent, and translated back on egress. Real tokens persist in a nono-owned macOS keychain entry for cross-session resume. A sandboxed `security find-generic-password "Claude Code-credentials"` must return nonces, not secrets.

### Additional context

I'm drafting up a PR!

Approach: 
- TLS-intercept the OAuth token endpoint (platform.claude.com/v1/oauth/token), rewrite the response to swap the real access/refresh tokens for nono_<hex> broker nonces before they reach the sandboxed agent
- Resolve nonces back to real tokens on egress to api.anthropic.com
- Real tokens persist only in a nono-owned, ACL'd macOS keychain entry
- HTTP/1.1 (in case the coding agent falls back to 1.1) and HTTP/2 intercept paths; configured via a profile `credential_routes oauth_intercept` entry

christine-at-datadog:
## Linked Issue

Closes #1265

## Summary

When a sandboxed agent runs `claude /login` (or other coding agent equivalent), the proxy TLS-intercepts the OAuth token endpoint, swaps the real `access_token`/`refresh_token` for `nono_<hex>` nonces before the response reaches the agent, and resolves nonces back to real tokens on egress. The agent's original keychain entry only ever holds nonces.

Built on existing primitives:
- Added a capture path in the nonce mint/resolve engine
- Hooked up HTTP/1.1 and HTTP/2 forwarding in the proxy's TLS intercept loop
- Added dispatch mode to tool-sandbox Capture action
- Set up a nono-only ACL entry to macOS keychain FFI

Net new:
- Fake tokens in the agent's keychain — when the agent logs in, it gets placeholder values instead of real credentials. The proxy swaps them back to real tokens on the way out. The real tokens live in a separate, nono-owned keychain entry the agent can't read.
- The broker remembers credentials across sessions — it saves the real tokens to its keychain entry so they survive restarts, detects when the agent has logged out and cleans up stale placeholders, and protects its entry so only nono can read it.
- Token response interception — when the OAuth server sends back the login response, nono catches it before the agent sees it, swaps the real tokens for placeholders, then lets the (now-modified) response through.
- credential_routes as a single config entry — instead of manually configuring proxy routes, you declare one entry with the API host and token endpoint, and nono figures out the rest.
- Same host vs. different host — if the token endpoint and API are on the same host, one proxy route covers both. If they're on different hosts (claude's case: platform.claude.com for login, api.anthropic.com for inference), nono generates two routes automatically.

## Additional Notes
- #1034 retrieves pre-existing secrets via keyring with the default system ACL. This broker captures OAuth token pairs at the token endpoint, stores them under a nono-only SecAccess ACL, and hands the agent nonces it cannot resolve outside a running nono session.
- The pattern should be agent agnostic.

## Adding OAuth capture to an existing profile
I included an example profile in the PR, but it's intended to be removed after review.

Add `credential_routes` alongside your existing `network.tls_intercept` block. Two cases:

**Same host for token endpoint and API:**
```json
"network": {
  "listen_port": [0],
  "tls_intercept": { "ca_lifecycle": "session" }
},
"environment": { "deny_vars": ["ANTHROPIC_BASE_URL"] },
"credential_routes": [
  {
    "name": "anthropic_oauth",
    "upstream": "https://api.anthropic.com",
    "capture": {
      "type": "oauth_intercept",
      "token_url_match": "/v1/oauth/token"
    }
  }
]
```

**Cross-host** (token endpoint on a different host than the API — claude's case):
```json
"credential_routes": [
  {
    "name": "anthropic_oauth",
    "upstream": "https://api.anthropic.com",
    "capture": {
      "type": "oauth_intercept",
      "token_host": "https://platform.claude.com",
      "token_url_match": "/v1/oauth/token"
    }
  }
]
```
`token_host` absent → defaults to `upstream`. Cross-host desugars to two routes: a capture route on `token_host` and an egress (nonce-resolve) route on `upstream`.

`listen_port: [0]` is required so the agent's OAuth callback server can bind. `deny_vars: ["ANTHROPIC_BASE_URL"]` prevents an enterprise gateway from routing inference traffic away from `api.anthropic.com` where nonces resolve.

## Checklist

## Agent Compliance Check

- [x] An issue already exists (#1265)
- [ ] I disclosed that I am an agent in the issue discussion
- [ ] I described my intent and approach in the issue discussion


crates/nono-proxy/src/tls_intercept/handle.rs
```diff
@@ -1012,13 +1012,35 @@ where
...
+        .is_some_and(|oc| {
+            let p = req.path.split('?').next().unwrap_or(req.path.as_str());
+            p == oc.token_url_match || p == oc.refresh_url_match
+        })
+        && ctx
+            .nonce_resolver
+            .as_deref()
+            .and_then(|r| r.oauth_capture())
+            .is_some();
+
```
lukehinds, line 1032: @christine-at-datadog I think this enables OAuth capture for both `token_url_match` and `refresh_url_match`, but the egress substitution path below only resolves broker nonces in headers. 

OAuth refresh requests normally send `refresh_token` in the JSON request body, and that body is still forwarded unchanged, so refresh will send `nono_<hex>` upstream instead of the real refresh token

crates/nono-proxy/src/tls_intercept/handle.rs
```diff
@@ -1064,12 +1086,40 @@ where
...
+                match crate::oauth_rewrite::rewrite_oauth_json_body(body, capture) {
+                    crate::oauth_rewrite::OauthRewriteOutcome::Rewritten { bytes, substituted } => {
+                        debug!(
+                            "oauth-capture (h1): substituted {substituted} token field(s) for {host}"
+                        );
+                        Some(bytes.to_vec())
+                    }
+                    crate::oauth_rewrite::OauthRewriteOutcome::NotJson
+                    | crate::oauth_rewrite::OauthRewriteOutcome::NoTokenFields => None,
+                }
```
lukehinds, line 1145: @christine-at-datadog This pass-through behavior might be unsafe for the credential-isolation contract. If the token response cannot be parsed/re-serialized, it returns `None`; the buffered forwarder then sends the original response unchanged, which may contain the real access/refresh tokens. I think OAuth capture should fail closed here or handle expected encodings before forwarding.

We always veer on hard failure in nono, I have foot gunned myself a few times with returning Nono.

lukehinds (commented):
hi @christine-at-datadog , solid work, its looking good, just a couple of things I found that might need digging into, looking forward to shipping this though, as its really adding to zero trust model.

crates/nono-proxy/src/tls_intercept/handle.rs
christine-at-datadog: Good catch and no arguments on failing closed. [The rewrite](https://github.com/nolabs-ai/nono/pull/1267/changes/7bf889b42f6059607dba3cca8f87052b80794461) now forwards the original upstream body only when it's parsed as a JSON object and verified to carry no access/refresh token field.

crates/nono-proxy/src/tls_intercept/handle.rs
christine-at-datadog: @lukehinds Ah you're right, this is a gap. https://github.com/nolabs-ai/nono/pull/1267/commits/fa7174a64c918666286f7f1e5a3eb67a9a56cc5c should fix the refresh forwarding.

intentionally-left-nil:
The network configs are starting to sprawl with `auth` , `aws_auth` (my fault) `oauth`, and now  `capture`. Is this the moment to step back and think about how we want this to evolve (instead of one organic feature at a time)

lukehinds:
@christine-at-datadog ok if we close this now ?

@intentionally-left-nil , this was the approach taking in the end, I will go a visit your points again there and check how much sprawl we have https://github.com/nolabs-ai/nono/pull/1343

PR #1268, opened 2026-06-26
feat(tool-sandbox): simplify self-invocation policy

### LINKED ISSUE #1266: Tool Sandbox self-invocation requires redundant self-edge policy
### Summary

A Tool Sandbox command cannot invoke itself unless the profile explicitly declares a self-edge in `can_use` and repeats a matching `from.<same-command>` policy. This is awkward for CLI broker tools that may re-enter their own executable during normal operation.

Self-invocation should continue under the command's current effective policy unless the profile explicitly defines a self-edge.

### Current behavior

Given a command policy like:

```jsonc
{
  "command_policies": {
    "commands": {
      "broker": {
        "executable": "/opt/example/broker",
        "can_use": ["helper"],
        "from": {
          "session": {
            "sandbox": {
              "fs_read": ["$WORKDIR"],
              "environment": {
                "allow_vars": ["*"]
              }
            }
          }
        }
      },
      "helper": {
        "executable": "/usr/local/bin/helper",
        "from": {
          "broker": {
            "sandbox": {
              "environment": {
                "allow_vars": ["*"]
              }
            }
          }
        }
      }
    }
  }
}
```

If `broker` invokes `broker`, Tool Sandbox treats that as a chained command edge and denies it because `broker.can_use` does not include `broker`.

If `broker` is added to `can_use`, Tool Sandbox then requires a matching `broker.from.broker` policy. In practice this forces profiles to duplicate the same sandbox:

```jsonc
"broker": {
  "can_use": ["broker", "helper"],
  "from": {
    "session": {
      "sandbox": { "...": "..." }
    },
    "broker": {
      "sandbox": { "...same policy repeated...": "..." }
    }
  }
}
```

### Expected behavior

If a command invokes itself and no explicit self-edge is configured, Tool Sandbox should treat it as a continuation of the command's current effective policy.

Explicit self-edges should still be honored for backward compatibility. That means:

- `broker -> broker` with no `broker` self-edge: use the current effective policy.
- `broker -> broker` with explicit `can_use: ["broker"]` or `from.broker`: keep existing chained-edge behavior.
- `broker -> helper`: still require `broker.can_use: ["helper"]` and `helper.from.broker`.

### Why this matters

Some broker-style CLIs naturally re-enter their own executable as part of normal operation. Requiring a self-edge makes profiles harder to read and maintain, and encourages duplicated policy blocks that can drift over time.

### Proposed fix

When resolving the caller for a Tool Sandbox shim request:

1. Detect that the active parent command name matches the requested command name.
2. If the profile does not define an explicit self-edge for that command, resolve the caller as the command's original launch caller.
3. If the profile does define an explicit self-edge, preserve current behavior and resolve the active parent command as the caller.

This keeps existing profiles compatible while removing the need for redundant self-edge configuration.

lukehinds:
The `resolve_caller` logic now treats self-invocation differently from invoking another command. When a command attempts to invoke itself, and there is no explicit `can_use` self-edge or `from` self-edge configured in its policy, the self-invoked command inherits the policy of its *launcher*. This means recursive calls effectively continue under the current effective policy without requiring explicit `can_use: ["<command>"]` configuration.

- Updated `nono-profile.schema.json` and documentation to clarify this implicit behavior for `can_use`.
- Modified `resolve_caller` on Linux and macOS to pass the requested command name and incorporate this new logic.
- Added `has_explicit_self_edge` helper to check for explicit self-policy.
- Introduced new tests to verify:
  - Different command invocation still resolves to the active parent command.
  - Self-invocation without explicit self-edge inherits the launcher's policy.
  - Self-invocation with an explicit self-edge honors that specific policy.

Resolves: #1266

## Checklist

### DIFF
--- a/crates/nono-cli/data/nono-profile.schema.json
+++ b/crates/nono-cli/data/nono-profile.schema.json
@@ -1194,7 +1194,7 @@
         "can_use": {
           "type": "array",
           "items": { "$ref": "#/$defs/CommandPolicyIdentifier" },
-          "description": "Commands this command may invoke through tool-sandbox chaining."
+          "description": "Other commands this command may invoke through tool-sandbox chaining. If no explicit self-invocation entry is configured, self-invocation keeps the command's current effective policy and does not require listing the command here."
         },
         "sandbox": {
           "$ref": "#/$defs/CommandSandboxConfig",
--- a/crates/nono-cli/src/command_policy.rs
+++ b/crates/nono-cli/src/command_policy.rs
@@ -480,6 +480,16 @@ impl CommandPolicyConfig {
     }
 }
 
+pub(crate) fn has_explicit_self_invocation_entry(
+    config: &CommandPoliciesConfig,
+    command_name: &str,
+) -> bool {
+    config.commands.get(command_name).is_some_and(|command| {
+        command.can_use.iter().any(|name| name == command_name)
+            || command.from.contains_key(command_name)
+    })
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(untagged)]
 pub enum CommandFromConfig {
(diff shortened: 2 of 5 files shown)


SequeI (approved):
lgtm, thank you!

PR #1269, opened 2026-06-26
feat: resource limiting

### LINKED ISSUE #1102: Resource limiting
nono currently enforces filesystem and network isolation on agent processes. This proposal adds **resource controls** - memory, CPU, process count, wall time, file size, and output - to prevent a runaway or malicious agent from exhausting host resources. The design offers the strongest enforcement mechanism available on the current system, while maintaining a mode which respects nono's userland only capability

> Note: this work should follow on from tool-sandboxing, which needs to land first, as it will have its own resource constraint capability for handling stdout/err 

An agent running inside nono's sandbox can currently consume unbounded memory, CPU, or processes. A compromised or misbehaving agent could:

- exhaust host memory and trigger OOM kills of unrelated processes
- spin CPU indefinitely
- fork-bomb via unrestricted process creation
- flood disk or stdout with unbounded output

OS-level filesystem isolation does not prevent any of these. Resource limits are a separate control layer.

## Design Approach: Selectable Backends

Resource enforcement is modelled as a set of selectable backends, because the mechanisms available differ significantly by environment. The operator picks a backend explicitly or uses `auto`:

```
nono run --memory 1g --resource-backend=auto
nono run --memory 1g --resource-backend=cgroup
nono run --memory 1g --resource-backend=portable
```

### Backends in strength order

| Backend | Mechanism | Strength |
|---------|-----------|----------|
| **CgroupDelegated** | User-delegated cgroups v2 subtree | Strongest — true process-tree accounting |
| **SystemdScope** | Systemd transient user/system scope | Cgroup-equivalent when verified |
| **PrivilegedHelper** | Small privileged helper creates constrained cgroup | Cgroup-equivalent; deferred to a later phase |
| **RlimitSupervisor** | POSIX `setrlimit` + supervisor monitoring | Portable; weaker memory enforcement |

`auto` tries each in order and selects the strongest available. If it falls back, it prints a diagnostic naming what changed and why.

## What Gets Controlled

The `ResourceLimits` policy covers:

- **Memory** — maximum address space or cgroup memory limit
- **CPU time** — maximum CPU seconds consumed
- **Wall time** — maximum elapsed time from launch to exit
- **Process count** — maximum concurrent processes in the agent tree
- **File size** — maximum single file write size
- **Output bytes** — maximum combined stdout/stderr captured

## Fail-Closed Behaviour

Resource limits are treated as security controls, not hints:

- `--resource-backend=cgroup` refuses to launch if cgroup enforcement is unavailable — it does not silently fall back to a weaker mode.
- If any specific limit cannot be applied by the selected backend, execution is aborted before the agent starts.
- `auto` fallback is always reported explicitly. There is no silent degradation.
- No privilege escalation path (sudo, pkexec, launchctl) is invoked implicitly.
- Configuration load failures for resource policy are fatal.

## Honest Limitations

The proposal is explicit about where `portable` enforcement is weaker, yet more portable as in there is no requirement for priviledge escalation (which in turn limits nono's portability). 

- `RLIMIT_AS` limits virtual address space, not physical memory — a process can exceed physical RAM via swap without hitting the limit.
- Sampled RSS monitoring in the supervisor can miss short memory spikes between samples.
- `RLIMIT_NPROC` does not account for all descendants on every platform.
- Process-group termination on violation can miss descendants that have intentionally escaped the group.

## Supervisor Integration

The supervisor process — which already sits outside the sandbox and manages child lifecycle — is the natural owner of runtime resource enforcement. It will:

- resolve the requested backend before any exec
- apply limits before the child starts where possible
- create and clean up cgroups when authorized
- enforce wall-clock and output-byte limits directly
- terminate the process group or cgroup on violation
- record resource violations in session audit metadata

This keeps resource policy in `nono-cli` and the library free of policy, consistent with nono's existing separation of concerns.

connrg:
Closes #1102.

## Why

A sandboxed agent that leaks or runs away can eat all the machine's RAM and drag everything else down with it. This adds a hard memory ceiling so a single run can't do that — if it goes over, only that run is killed, instantly, by the kernel.

## What you get

A new flag, on `nono run` and `nono shell`:

```bash
nono run --memory 512M -- my-agent
```

Sizes are case-insensitive: `K`/`M`/`G`/`T` (and `Ki`/`Mi`/…) are binary (1024-based), `KB`/`MB`/`GB`/`TB` are decimal (1000-based) — e.g. `512M`, `1Gi`, `2G`. A bare number is bytes, with a 1 MiB minimum on the flag, so `--memory 512` (512 bytes — enough to OOM-kill anything instantly) is refused with a hint pointing at `512M`, rather than silently enforcing a useless cap.

The cap covers the **whole process tree** the sandbox spawns, not just the first process. It shows up in `--dry-run` and the capability summary, marked as a hard cap so it's clear a breach kills the run rather than throttling it.

## Setting it in JSON

The cap is also a first-class, top-level key in the capability manifest — a sibling of `filesystem`/`network`/`process`, not buried inside another section:

```json
{
  "version": "0.1.0",
  "process": { "exec_strategy": "supervised" },
  "resources": { "memory_bytes": 536870912 }
}
```

```bash
nono run --config manifest.json -- my-agent
```

The flag and the manifest resolve to the same enforced cap. Friendly sizes like `512M` are a flag convenience; the manifest takes resolved bytes so it stays a clean machine contract (schema `minimum: 1`). A manifest that sets `resources` without `exec_strategy: "supervised"` is rejected up front, since the cap is enforced by the supervising parent.

## What happens when it's hit

The kernel kills the run and nono explains exactly why, instead of dying with a cryptic "killed" message that looks like a file-permission problem:

```text
[nono] memory limit exceeded: the sandboxed process tree was killed by the kernel for using too much memory.
       limit (--memory): 512.0 MiB
       peak memory:      512.0 MiB
       OOM kills:        2 (whole-sandbox kills: 1)
       swap:             disabled (memory.swap.max=0) — nothing could spill to swap
       scope:            the whole sandbox was killed together (memory.oom.group=1)
       hint: raise the ceiling to allow more memory, e.g. --memory 1G.
```

Runs killed for any other reason, or that exit normally, look exactly as before.

## How it works (Linux)

Enforced with **cgroup v2** — the same kernel feature containers use. For each run nono creates a small control group, sets the ceiling, and the sandboxed process puts *itself* into the group before it can start anything, so there's no gap for a child to escape through. Swap is turned off (so the limit can't be dodged), and on overflow the whole group is killed together rather than one random victim. The group is torn down when the run ends.

It is **fail-closed**: if the limit can't be set up, the run is refused (and logged) rather than quietly running unprotected. It also refuses to start if the sandbox is granted **write** access overlapping `/sys/fs/cgroup` (e.g. `--allow /sys`) while a limit is active — otherwise the sandboxed process could rewrite its own `memory.max` and lift the cap. Use a read-only or narrower grant instead.

## Scope and limits

- **Memory only** for now. (An earlier draft's manifest also had `cpu_max_percent` and `max_procs` keys; those were pulled out to keep this focused and well-tested — they can come back later.)
- **Linux only**, and needs a normal user login (a systemd session with cgroup delegation). On other platforms, asking for a limit fails clearly rather than pretending to enforce one.
- **`nono wrap` doesn't accept `--memory`** (it execs directly and can't create the enforcement cgroup); a limit arriving via a manifest is refused at runtime there too.
 - **Not yet settable in a profile.** A reusable profile (`nono-profile.json`) has no top-level `resources` key, so the cap comes from the flag or the manifest. Adding one is a small follow-up (where the key lives; what wins if a profile and the flag disagree). Relatedly, the profile schema's existing (per-command) resource fields still allow `0` where this manifest
  enforces `minimum: 1` — tightening that to match is tracked in #1319.

## Testing

- **Unit tests**
  - Size parser:
    - Binary and decimal units
    - 1 MiB minimum when the unit hint is used
    - Per-unit overflow boundaries
    - Error messages identify the exact rejection reason
  - Manifest validation:
    - Rejects unknown or removed keys
    - Requires `supervised`
    - Round-trips the byte value correctly
  - Serialization
  - Cgroup helpers, including adversarial cases:
    - Lookalike delegation paths
    - UID substring confusion
    - `parse_leaf_pid`

- **Fail-closed tests** (CI-runnable, no cgroup required)
  - `child_self_attach` fails on an invalid file descriptor, causing the caller to `_exit` instead of running without limits
  - Writable grants that overlap `/sys/fs/cgroup` are rejected when a memory limit is configured
  - On non-Linux platforms, requesting a memory limit returns `UnsupportedPlatform`
  - Only exit code **137** is treated as an OOM kill (not crashes such as **139** or `SIGTERM` exit **143**)

- **Live tests** (`--ignored`, requires a real cgroup)
  - A child process is limited to **64 MiB** and attempts to allocate **128 MiB**
  - The child is OOM-killed while the parent process survives
  - The kernel reports the expected cgroup settings:
    - `memory.max`
    - `memory.swap.max=0`
    - `memory.oom.group=1`
  - Cleanup removes the test cgroup leaf, and stale leaves left by dead supervisors are automatically swept


SequeI (approved):
lgtm!

PR #1284, opened 2026-06-27
feat(gpu): enable nvidia gpu 570+ thread naming

lukehinds:
The NVIDIA driver 570+ attempts to name threads by writing to `/proc/<tgid>/task/<tid>/comm` via `openat`. Landlock's read-only `/proc/self/task` capability would block these operations.

This change introduces a supervisor fast-path to auto-approve these specific writes when `--allow-gpu` is active and NVIDIA devices are detected. This allows the NVIDIA driver to function correctly under the sandbox.

The `/proc/self/task` Landlock capability is now explicitly narrowed to read-only. This improves security by routing `comm` writes through the supervisor's vetted fast-path, validating that the target `tgid` matches the child's process ID.

I assess performance impact, by taking a quick look at how common inference systems work, and the majority of the perfomance impact will be during only model loading which is acceptable , as it won't be noticeable to any degree and won't impact LLM inference

- Implement a `gpu_comm` fast-path in `supervisor_linux.rs` to approve
  writes to `/proc/<tgid>/task/<tid>/comm` for the child's `tgid`.
- Detect NVIDIA devices during sandbox preparation and set a new
  `allow_gpu_nvidia_active` flag.
- Propagate `allow_gpu_nvidia_active` to `seccomp_policy.gpu_comm`.
- Narrow the Landlock capability for `/proc/self/task` to read-only.

This PR is branched off #1283 which most land first.


SequeI (approved):
lgtm!

PR #1287, opened 2026-06-28
refactor: load_with_diagnostics should be async

### LINKED ISSUE #1286: AWS: CredentialStore::load_with_diagnostics is synchronous but does async work
`CredentialStore::load_with_diagnostics` is a synchronous function, but two of the credential backends it initializes require async I/O: OAuth2 performs an initial token exchange over HTTPS, and the upcoming AWS credential chain runs `aws_config::from_env().load().await` to resolve credentials.

Since the function is sync, both backends are forced to bridge into async using `block_in_place`:

```rust
// oauth2.rs
let (access_token, expires_in) = tokio::task::block_in_place(|| {
    tokio::runtime::Handle::current().block_on(exchange_token(&config, &tls_connector))
})?;
```

This is an anti-pattern: `block_in_place` requires the multi-thread Tokio scheduler, imposing a hidden runtime constraint that is invisible from the function signature and only discoverable via a runtime panic:

```
thread panicked at 'can not call `block_in_place` from the current thread runtime'
```

The fragility is already visible in the test suite — the one test that exercises an AWS route requires a non-default annotation specifically to avoid this panic:

```rust
// h2_forward.rs:1379
#[tokio::test(flavor = "multi_thread")]  // required due to block_in_place in load_with_diagnostics
```

The upcoming AWS work causes this brittle pattern to spread: a second `block_in_place` call site has been added for the same reason.

The correct fix is to make `load_with_diagnostics` (and its deprecated `load` wrapper) async. Both bridge call sites become plain `.await`, the hidden runtime constraint is eliminated, and tests can use the default `#[tokio::test]` flavor.

intentionally-left-nil:
This changes load, load_with_diagnostics, and make_cmd_route_stores to be async functions. This is needed because the Oauth code already does async work within this codepath. The old code used tokio's block_in_place to coerce an async flow into a single thread but this is brittle. The correct pattern is to make the code async, and handle the async at the top-level (in server.rs)

## Linked Issue

Closes #1286

## Summary
This is a fairly straightforward refactor - instead of sync load_with_diagnostics -> tokio::block_in_place->Oauth async code
the new pattern is to make load_with_diagnostics async, so that the OAuth code can just call `exchange_token` correctly. This new pattern will also benefit the upcoming AWS credential work, which also needs to do async work here

I had Claude sonnet take the bug description, and generate the code solution for the refactor

## Test Plan
Unit tests

## Checklist


SequeI (approved):
lgtm, this is a clean refactor. Thanks!

PR #1296, opened 2026-06-29
feat(profile): expand $VAR tokens from process env in capture commands and paths

### LINKED ISSUE #1295: Allow $VAR tokens from process environment in credential capture commands and sandbox path fields

Some paths in a profile aren't known when the profile is written — they depend on the runtime environment of the process launching nono. Today there's no way to express this; paths must be hardcoded.

Expand bare `$IDENTIFIER` tokens in `credential_capture` command strings and sandbox path fields from the process environment at resolution time. Unset variables should be preserved as-is so existing profiles are unaffected.

Pre-processing the profile at launch time using a wrapper script that runs `envsubst` or `sed` on the profile before passing it to `nono run --profile`.

## How is this blocking you?

kipz:
## Linked Issue

Closes nolabs-ai/nono#1295

## Summary

Introduces a shared `substitute_vars` kernel in `policy.rs` — a char-by-char scanner parameterised by a `var_start` predicate and a lookup closure. Two callers:

- **`expand_env_vars`** (open set, process env): wired into `expand_path` and `expand_env_vars_in_command` so `credential_capture` commands and sandbox path fields can reference env vars exported by the launcher. Unset variables are preserved as the literal `$NAME` token.
- **`wiring::expand_vars`** (closed set, error on unknown): migrated from an identical inline scanner to `substitute_vars`, with an uppercase-only `var_start` predicate to leave `$` regex anchors and jq-style `$var` untouched.

## Agent Disclosure

This PR was generated by an AI agent (Claude). Files consulted: `policy.rs`, `proxy_runtime.rs`, `wiring.rs`, `capability_ext.rs`, `AGENTS.md`. DCO sign-off is included. No `.unwrap()` or `.expect()` in new production code. Existing `clippy` failures (8 errors in pre-existing code) are not introduced by this PR.

## Test Plan

```
cargo test -p nono-cli expand_vars   # 9 tests, all pass
cargo test -p nono-cli               # full suite passes
```

## Checklist

- [x] All new code follows the project's coding standards and is covered by tests


lukehinds:
There might be gap to what you have in the issue #1295 @kipz , unless its the intention.. . The PR adds arbitrary $VAR expansion through `policy::expand_path`, but normal profile sandbox paths resolve through
`profile::expand_vars` instead, which still only handles the built-in set like `$HOME, $WORKDIR, $TMPDIR, $XDG_*,` etc.

So a sandbox path field such as "$MY_TOOL_HOME" or "$FOO" may still remain literal in filesystem.allow/read/write. 

What you could do is also apply the new expansion in `capability_ext.rs`, where profile filesystem.allow/read/write entries are resolved via `profile::expand_vars` unless ofc if this PR is only intended to cover policy paths.

lukehinds:
The relevant calls are around here:

  // crates/nono-cli/src/capability_ext.rs
```
  for path_template in &fs.allow {
      let path = expand_vars(path_template, workdir)?;
      ...
  }

  for path_template in &fs.read {
      let path = expand_vars(path_template, workdir)?;
      ...
  }

  for path_template in &fs.write {
      let path = expand_vars(path_template, workdir)?;
      ...
  }
```

That expand_vars is imported from crate::profile, so it means profile::expand_vars, not the new policy::expand_path.

kipz:
Thanks @lukehinds - that was definitely not my intention! I'll get some tests around all this and update. Thanks again.

kipz:
@lukehinds fixed as you suggested:
>What you could do is also apply the new expansion in capability_ext.rs, where profile filesystem.allow/read/write entries are resolved via profile::expand_vars unless ofc if this PR is only intended to cover policy paths.

SequeI (approved):
lgtm, thank you. This will be very useful!

PR #1322, opened 2026-06-30
feat(tool-sandbox): add exec intercept action

### LINKED ISSUE #1318: Add an "exec" intercept action to run a custom handler for one subcommand

An intercept rule can match a subcommand by argv prefix, but the only actions are `respond` (static stdout), `capture` / `capture_credential` (run the real binary), and `approve`. There's no way to handle a single matched subcommand with custom logic while letting the command's other subcommands run normally inside their sandbox.

A new `exec` intercept action: `{ "type": "exec", "command": [...] }`. On match, nono runs `command` (with the original argv forwarded) in place of the real binary, inside the command's existing sandbox, and returns its stdout/stderr/exit — like `respond`, but dynamic. Non-matching subcommands still `passthrough`.

The command-level `executable` field, but it replaces the whole command. Wrapping and re-execing the real binary either loops (by name) or, via `allow_direct_exec_bypass` (by path), runs with outer session capabilities instead of the command's sandbox — losing its network/credential scoping.

### How is this blocking you?

It would be a nice improvement but I have a workaround.

### Additional context

Variant would live in `InterceptActionConfig` (`command_policy.rs`) with dispatch alongside the existing actions in `tool-sandbox/platform/{linux,macos}.rs`.

kipz:
## Linked Issue

Closes #1318

## Summary

Add an `exec` intercept action: `{ "type": "exec", "command": [...] }`. On a matching subcommand, nono runs the helper (with the original argv forwarded) in place of the real binary, inside the command's existing sandbox, and returns its stdout/stderr/exit — dynamic where `respond` is static. Helpers are pre-resolved with identity expectations and gated by the same non-writable-executable check as command binaries.

NB: this follows the existing pattern at the command level

## Test Plan

`make ci` clean (clippy -D warnings + fmt-check + tests). New tests cover serde, validation (empty/relative/NUL/absolute helper), helper identity capture, argv-forwarding order, and `$VAR` expansion.

## Checklist

## Agent Disclosure

Prepared by an AI coding agent. Added the `Exec` variant to `InterceptActionConfig`, TOCTOU-safe helper pre-resolution, a non-writable-executable trust gate, dispatch in `tool-sandbox/platform/{macos,linux}.rs` reusing the existing child-launch path so the helper runs in the matched command's sandbox, plus schema and `tool-sandbox.mdx` docs. Files consulted: `command_policy.rs`, `tool-sandbox/platform/{macos,linux}.rs`, `tool-sandbox/{env,policy}.rs`, `policy.rs`, `data/nono-profile.schema.json`, `docs/cli/features/tool-sandbox.mdx`, `AGENTS.md`.

## Agent Compliance Check

- [x] An issue already exists (#1318)
- [x] I provided required attribution for reused or adapted code (no third-party code adapted)

### DIFF
--- a/crates/nono-cli/src/command_policy.rs
+++ b/crates/nono-cli/src/command_policy.rs
@@ -1163,6 +1163,11 @@ pub(crate) fn resolve_policy_exec_helpers(
             };
             let expanded = crate::policy::expand_env_vars_strict(helper_raw)?;
             let helper_path = PathBuf::from(&expanded);
+            if !helper_path.is_absolute() {
+                return Err(nono::NonoError::ProfileParse(format!(
+                    "command '{command_name}' intercept rule {rule_index} exec helper must be an absolute path; got '{expanded}'"
+                )));
+            }
             if helpers.contains_key(&helper_path) {
                 continue;
             }
@@ -4088,6 +4093,16 @@ mod tests {
         let resolved = helpers.get(&helper).expect("helper resolved");
         assert!(!resolved.sha256.is_empty());
         assert!(resolved.canonical_path.is_absolute());
+    }
+
+    #[test]
+    fn resolve_policy_exec_helpers_rejects_relative_helper() {
+        let config = git_config_with_exec(vec!["relative/helper".to_string()]);
+        let err = resolve_policy_exec_helpers(&config).expect_err("must reject relative helper");
+        assert!(
+            err.to_string().contains("absolute path"),
+            "expected absolute-path rejection, got: {err}"
+        );
     }
 
     #[test]
--- a/crates/nono-cli/src/command_policy.rs
+++ b/crates/nono-cli/src/command_policy.rs
@@ -400,6 +400,23 @@ pub enum InterceptActionConfig {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         timeout_secs: Option<u64>,
     },
+    /// Run a helper binary instead of the command's real binary, inside the
+    /// matched command's existing sandbox (same capabilities, env including
+    /// injected credentials/proxy routing, and cwd). The helper at `command[0]`
+    /// runs with `command[1..]` as fixed leading args, followed by the original
+    /// invocation's user args (`argv[1..]`). The helper's stdout/stderr are
+    /// streamed to the caller and its exit code is propagated.
+    ///
+    /// `command[0]` is `$VAR`-expanded then required to be an absolute path; it
+    /// is resolved exactly like a command binary (canonicalize/stat/sha256 with
+    /// identity expectations for TOCTOU protection) and is subject to the same
+    /// non-writable-executable trust gate as the `executable` field.
+    Exec {
+        /// Helper invocation: `command[0]` is the absolute helper path (after
+        /// `$VAR` expansion); `command[1..]` are fixed leading args prepended
+        /// before the forwarded original args.
+        command: Vec<String>,
+    },
 }
 
 /// A sub-command mediation rule on a [`CommandPolicyConfig`].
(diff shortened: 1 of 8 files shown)


panga:
+1 this feels like a solid way to support shim helpers out of the box instead of having to manually inject them. It would be a really useful addition for me and could replace some of the ad-hoc scripts i'm using 👍

crates/nono-cli/src/tool-sandbox/platform/linux.rs
```diff
@@ -2448,6 +2535,29 @@ fn validate_controlled_binary_immutability(
     Ok(())
 }
 
+/// Apply the same non-writable-executable trust gate to resolved `exec`
+/// intercept helpers as to command binaries (see the macOS twin for rationale).
+fn validate_controlled_exec_helper_immutability(
```
connrg, line 2547: One thing I noticed: this function is the safety check that stops the agent from quietly swapping in its own helper program, and it doesn't seem to have a test yet. Would you mind adding one here (and in the macOS copy at macos.rs:3679) that covers both cases; At the moment only the happy path looks covered. Thanks!

docs/cli/features/tool-sandbox.mdx
```diff
@@ -92,6 +92,21 @@ The selected `sandbox` object describes what the child receives:
...
+
+| `action.type` | Behavior |
+|---|---|
+| `passthrough` | Run the real binary in its sandbox. This is the default when no rule matches. |
+| `respond` | Return a static `stdout` string to the caller without forking the binary. |
+| `capture` | Run the real binary, buffering its stdout for broker inspection before returning it. |
+| `capture_credential` | Run the real binary, store its stdout as a named ambient `credential`, and return a broker nonce. |
+| `approve` | Route the invocation through an approval backend before forking (optional `timeout_secs`). |
+| `exec` | Run the helper at `command[0]` — with `command[1..]` then the original arguments — in place of the real binary, inside the matched command's sandbox; stream its stdout/stderr and propagate its exit code. `command[0]` is `$VAR`-expanded, must be absolute, is resolved with identity expectations, and is subject to the same non-writable-executable trust gate as a pinned `executable`. |
```
connrg, line 123: Minor docs suggestion, feel free to take or leave: it might be worth mentioning that the helper runs with the command's real credentials;so pointing it at something like /bin/sh would let the agent run anything using those credentials. A short note advising against shells here could save someone a surprise down the line. Thanks!

crates/nono-cli/src/tool-sandbox/platform/linux.rs
kipz: Good catch, added `writable_exec_helper_override_is_explicit_for_sandbox_writable_paths` to both linux.rs and macos.rs, right next to `writable_policy_command_override_is_explicit_for_sandbox_writable_paths`. It exercises both failure cases (file writable, parent dir writable) plus the two exemption paths (global `allow_writable_executables` and per-command `allow_writable_executable`), mirroring the existing binary-immutability test. Confirmed it fails if the gate is stubbed to `Ok(())`.

docs/cli/features/tool-sandbox.mdx
kipz: Good call, added a short paragraph after the exec example noting the helper runs with the matched command's real credentials/capabilities and isn't further sandboxed, so pointing it at a general-purpose shell would hand those credentials to arbitrary commands — recommends a purpose-specific helper instead.

connrg: thanks looks good

crates/nono-cli/src/tool-sandbox/platform/linux.rs
connrg: thanks looks good!

crates/nono-cli/src/command_policy.rs
```diff
@@ -1088,6 +1105,65 @@ pub(crate) fn resolve_policy_command_binaries(
...
+/// `candidate_command_match` (canonicalize, stat, sha256, classify shape).
+///
+/// The returned map is keyed by the env-expanded helper path as written in the
+/// profile, so dispatch can re-expand `command[0]` and look the helper up. A
+/// helper that fails to resolve (missing/non-executable) is surfaced as an
+/// error rather than skipped: unlike an absent command binary (which simply
+/// disables that command's policy), an exec helper that cannot run would leave
+/// the matched subcommand with no viable handler.
+#[cfg(any(test, target_os = "linux", target_os = "macos"))]
+pub(crate) fn resolve_policy_exec_helpers(
```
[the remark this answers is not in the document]
kipz: good catch, fixed in ff162504 — resolve_policy_exec_helpers now rejects relative helpers before they ever hit candidate_command_match

crates/nono-cli/src/tool-sandbox/policy.rs
```diff
@@ -57,6 +57,85 @@ pub(super) fn resolve_intercept_action<'a>(
...
+/// intercept action's `command`.
+///
+/// Returns `(helper_path, extra_args)` where `helper_path` is `command[0]`
+/// after `$VAR` expansion (used as the lookup key into the plan's pre-resolved
+/// `exec_helpers` map) and `extra_args` are `command[1..]` after `$VAR`
+/// expansion, to be inserted ahead of the forwarded original args. Platform
+/// dispatch resolves the actual `ResolvedCommandBinary` from its own state map
+/// using `helper_path`, so this stays platform-agnostic.
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+pub(super) fn resolve_exec_command(
```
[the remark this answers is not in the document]
kipz: fixed too, in ff162504 — resolve_exec_command now rejects a relative helper before it's used as the exec_helpers lookup key

connrg (approved):
lgtm, thank you!

SequeI (approved):

PR #1324, opened 2026-07-01
feat(tool-sandbox): per-intercept sandbox override + per-command unsafe_macos_seatbelt_rules

### LINKED ISSUE #1323: feat: per-intercept sandbox override for a matched subcommand

When mediating `gh` through `intercept` rules, every matched passthrough
invocation runs under the same single command sandbox. `gh auth switch` refuses
to run when `GH_TOKEN` is set, yet the rest of `gh` needs a proxy credential
that injects `GH_TOKEN`. There is no way to give one subcommand (`auth switch`)
a different capability shape — drop the credential so no `GH_TOKEN` is set,
while granting the extra filesystem/keychain access the switch needs — without
changing the whole command's sandbox.

Allow an `intercept` rule to carry an optional `sandbox` (the existing
`CommandSandboxConfig`). When it matches, the passthrough runs under that
sandbox instead of the command's selected sandbox, for that invocation only:

```jsonc
{
  "args": ["auth", "switch"],
  "action": { "type": "passthrough" },
  "sandbox": {
    "fs_read":  ["~/.config/gh", "~/Library/Keychains"],
    "fs_write": ["~/.config/gh", "~/Library/Keychains"],
    "network": { "allow_domain": ["github.com"] }
  }
}
```

Reuses the existing sandbox type and validation. Credentials resolve lazily, so
omitting `credentials` injects no env var (no `GH_TOKEN`) for that invocation.
It can only narrow/retarget capabilities, never grant beyond what's written.

## How is this blocking you?

It's blocking production use of nono. Injecting a proxy credential that sets
`GH_TOKEN` for `gh`'s API subcommands makes `gh auth switch` — and any other
`gh` subcommand that rejects a preset `GH_TOKEN` — unusable, because there's no
way to inject the credential command-wide and still run those subcommands.

kipz:
## Linked Issues

Closes #1323 (per-intercept sandbox override)
Relates to #678 (top-level `unsafe_macos_seatbelt_rules`) — extends that escape hatch to per-command scope

## Summary

Two related tool-sandbox additions:

**1. Per-intercept sandbox override.** Adds an optional `sandbox` to an `intercept` rule. When the rule matches, the process it launches runs under that sandbox instead of the command's selected sandbox — for that matched invocation only. It applies to every action that launches a process (`passthrough`, `capture`, `capture_credential`, and an approved `approve`); it is rejected on `respond`, which returns static output and launches nothing. Lets one subcommand omit a credential (so its `env_var` isn't injected) and take a different fs/network shape. Reuses `CommandSandboxConfig` and its validation; can only narrow/retarget.

**2. Per-command `unsafe_macos_seatbelt_rules` escape hatch.** Adds an optional macOS-only `unsafe_macos_seatbelt_rules: [String]` to `CommandSandboxConfig`, mirroring the existing top-level profile field but scoped to a single command's child sandbox. Because it lives on `CommandSandboxConfig`, it is accepted wherever a per-command sandbox is defined: a command's own `sandbox`, a `from.<caller>` edge, and (from addition 1 above) an `intercept[i].sandbox` override. The rules are appended to the child sandbox's platform-rule tail — after the generated denies, including the exec gate's `(deny process-exec*)` — so a later `(allow …)` wins under Seatbelt's last-matching-rule semantics. macOS-only; ignored on Linux. Omitted from serialized output when empty.

Motivation: some tools fork a controlled binary by absolute path, bypassing the PATH-resolved broker shim. For example `gh auth switch` uses go-keyring, which execs `/usr/bin/security` directly; the child exec gate denies it and the keychain read fails. The structured sandbox fields can't express "let this child exec+read this specific absolute binary", so the escape hatch does:

```jsonc
"intercept": [{
  "args": ["auth", "switch"],
  "action": { "type": "passthrough" },
  "sandbox": {
    "fs_read": ["~/Library/Keychains", "..."],
    "fs_write": ["~/Library/Keychains"],
      "(allow process-exec* (literal \"/usr/bin/security\"))",
      "(allow file-read* (literal \"/usr/bin/security\"))"
    ]
  }
}]
```

Both rules are required: the exec allow lets the forked `security` launch; the file-read allow lets it map its own binary (it inherits the parent command's sandbox, which otherwise grants read only of the parent's binary and the shims). Verified live end-to-end: `gh auth switch` succeeds under a scratch profile using this field, with no hardcoded path in nono.

Generated by an AI agent (Claude) on behalf of @kipz. Reviewed CLAUDE.md for the affected area; DCO signed-off, no `unwrap`/`expect`, `NonoError` used, overrides validated via the existing sandbox-validation path, tests + docs added, `make ci` clean.

## Test Plan

`make fmt` + `make ci` (clippy `-D warnings`, fmt-check, tests, docs lint) — green.

Per-intercept override tests: serde round-trip (+ omission when absent), well-formed override validates, malformed override (unknown credential) rejected, override on a `respond` action rejected, override on a launching action accepted, `resolve_intercept_action` carries the override, dispatch selects `intercept.sandbox.unwrap_or(policy)`.

Per-command `unsafe_macos_seatbelt_rules` tests: serde round-trip, omission when empty (`skip_serializing_if`), `merge_child` appends/dedups, empty/whitespace rule rejected by validation, and a macOS test asserting the rule lands in the generated child profile *after* the `(deny process-exec*)` exec-gate line (so it wins). Schema snapshot (`schema_shape`) updated to include the new property.

## Checklist

- [x] An issue already exists
- [ ] I provided required attribution for reused or adapted code


crates/nono-cli/src/tool-sandbox/policy.rs
```diff
@@ -38,6 +42,7 @@ pub(super) fn resolve_intercept_action<'a>(
             return ResolvedInterceptAction {
                 action: &rule.action,
                 rule_args: Some(&rule.args),
+                sandbox: rule.sandbox.as_ref(),
             };
         }
         if shim_args.len() >= rule.args.len()
```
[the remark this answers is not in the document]
kipz: This is already a problem with per-intercept tool capture....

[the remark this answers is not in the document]
kipz: Confirmed this is pre-existing (unchanged vs. `main`, added in #1105) and not introduced by this PR — #1324 only adds the `sandbox` field to `ResolvedInterceptAction` without touching the matcher. Filed as #1328 to track separately.

crates/nono-cli/src/tool-sandbox/platform/macos.rs
```diff
@@ -1888,9 +1893,27 @@ fn build_child_caps(
...
+    // wins, so a rule like `(allow process-exec* (literal "/usr/bin/security"))`
+    // overrides the exec gate's earlier `(deny process-exec*)`.
+    add_unsafe_seatbelt_rules(&mut caps, policy)?;
     Ok(caps)
 }
 
+/// Append a command sandbox's opt-in raw macOS Seatbelt rules to the child's
+/// platform rules. Emitted after all generated rules so they win under
+/// last-matching-rule semantics. No-op when the list is empty.
+fn add_unsafe_seatbelt_rules(
```
connrg, line 1907: One question: can these new options (`unsafe_macos_seatbelt_rules` and per-intercept `sandbox` overrides) weaken the sandbox? For example, could a profile use `(allow default)` or otherwise grant more access than the base sandbox?

If so, should we limit where these can come from? I'm thinking about shared or downloaded profiles, a user could install one that quietly reduces protection. Should these be restricted to hand-written profiles, like the `binary` field, or are all profiles considered trusted?

kipz: @connrg, thanks for taking a look

>One question: can these new options (unsafe_macos_seatbelt_rules and per-intercept sandbox overrides) weaken the sandbox? For example, could a profile use (allow default) or otherwise grant more access than the base sandbox?

By base, do you mean the 'agent' sandbox? If so, then yes. Command sandboxes are peer sandboxes anyway (before this PR) - not children. They can already have different and wider filesystem and network grants than the agent sandbox.

>If so, should we limit where these can come from? I'm thinking about shared or downloaded profiles, a user could install one that quietly reduces protection. Should these be restricted to hand-written profiles, like the binary field, or are all profiles considered trusted?

Right now, all profiles are considered trusted for sure. Do you think that's unreasonable? To be clear though, this is already the case with command sandboxes, so this PR doesn't change that. All we're really doing here is making all commands and sub-commands have the same expressive capabilities.

kipz: @connrg  - f3c6bf7f now restricts `unsafe_macos_seatbelt_rules` (top-level and nested in command/from/intercept sandboxes) to user-authored profiles only, mirroring the existing `binary` field trust boundary. Pack/registry/built-in profiles have these rules stripped with a warning instead of silently applying them.

connrg: Thanks @kipz , this covers what I was worried about. My main concern was a downloaded profile sneaking in an "allow everything" rule, and now those raw rules only work in profiles the  user wrote themselves. That's the important bit.And fair point on the rest; those overrides aren't really any more powerful than the per-command sandboxes we already allow, so no need to change anything there. Looks good to me 👍

connrg (approved):
LGTM

PR #1327, opened 2026-07-01
fix(tests): share stdin-manipulation lock between capture_helper stdin tests

### LINKED ISSUE #1326: Flaky stdin-manipulating tests in proxy_runtime: dup(stdin) failed under parallel test execution
### What happened?

`capture_helper_with_stdio_true_receives_null_not_terminal_stdin` and
`capture_helper_with_interaction_stdin_true_inherits_terminal_stdin`
(`crates/nono-cli/src/proxy_runtime.rs`) each temporarily replace the
process's stdin fd (fd 0) via `libc::dup`/`libc::dup2` to simulate a
terminal vs. non-terminal stdin for the credential-capture helper.

Each test guards this with its own separate `static Mutex<()>` local
to that test function (`STDIN_LOCK`). Since `cargo test` runs tests in
parallel by default and both tests mutate the same global fd 0, two
independent per-test mutexes don't prevent the two tests from racing
against each other. When they run concurrently, one can observe a
`dup(stdin) failed` panic (or otherwise corrupted fd state) because the
other test's `dup2` calls stomp on fd 0 concurrently.

Both tests pass reliably regardless of test execution order/parallelism.

### Steps to reproduce

1. `cargo test -p nono-cli --bin nono proxy_runtime::tests::capture_helper_with` (default parallel execution)
2. Repeat a few times
3. Observe intermittent failure (~1 in 4 runs locally)

Running with `--test-threads=1` is reliably green, confirming this is a
concurrency race between the two tests rather than a bug in either
test's logic.

### Relevant output or logs

```
thread 'proxy_runtime::tests::capture_helper_with_interaction_stdin_true_inherits_terminal_stdin' panicked at crates/nono-cli/src/proxy_runtime.rs:3622:13:
dup(stdin) failed
```

### nono version

dev build at 50304580 (upstream/main)

### Operating system

macOS

### OS version / distro

macOS 15.x (Darwin 25.5.0, arm64)

### Additional context

Planned fix: replace the two independent per-function `static
STDIN_LOCK` mutexes with a single module-level shared
`static STDIN_MANIPULATION_LOCK`, so the two tests can't run
concurrently against fd 0. Test-only change, no production code paths
affected; slightly increases serialization between these two tests but
both are fast (sub-100ms), so no meaningful runtime impact.

Filed per the repo's Coding Agent Contribution Policy in AGENTS.md to
disclose intent before implementation. Will wait for confirmation
before opening a PR.

kipz:
Fixes #1326

## Summary

`capture_helper_with_stdio_true_receives_null_not_terminal_stdin` and
`capture_helper_with_interaction_stdin_true_inherits_terminal_stdin`
each dup/dup2 the process's global stdin fd (0) to simulate terminal
vs. non-terminal stdin, guarded by their own separate `static Mutex`.
Since `cargo test` runs tests in parallel by default, the two per-test
mutexes don't prevent the tests from racing against each other on
fd 0, producing an intermittent `dup(stdin) failed` panic.

This replaces the two independent mutexes with one shared
`STDIN_MANIPULATION_LOCK` so the two tests can't run concurrently.
Test-only change; no production code paths affected.

## Agent Compliance Check

- [x] An issue already exists (#1326)
- [x] I provided required attribution for reused or adapted code (n/a — no reused/adapted third-party code)
- [x] I validated and canonicalized all relevant paths (n/a — no path handling in this change)

This contribution was prepared with the assistance of an AI coding agent (Claude Code), under my direct supervision and review.

### DIFF
--- a/crates/nono-cli/src/proxy_runtime.rs
+++ b/crates/nono-cli/src/proxy_runtime.rs
@@ -2481,6 +2481,11 @@ mod tests {
         CommandCredentialGrantPolicyConfig, CommandPolicyConfig, EndpointRuleConfig,
     };
 
+    // Shared across all tests that dup/dup2 the process's stdin fd, so two such tests
+    // never race on fd 0 when cargo runs them concurrently.
+    #[cfg(unix)]
+    static STDIN_MANIPULATION_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
     #[cfg(unix)]
     #[test]
     fn set_intercept_ca_dir_permissions_fails_closed() -> Result<()> {
@@ -3526,9 +3531,7 @@ mod tests {
     fn capture_helper_with_stdio_true_receives_null_not_terminal_stdin() -> Result<()> {
         use nix::libc;
 
-        // Serialize stdin manipulation across concurrent test threads.
-        static STDIN_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
-        let _guard = STDIN_LOCK
+        let _guard = STDIN_MANIPULATION_LOCK
             .lock()
             .map_err(|e| NonoError::SandboxInit(format!("stdin lock poisoned: {e}")))?;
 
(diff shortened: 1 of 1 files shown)


SequeI (approved):
lgtm!

PR #1339, opened 2026-07-03
fix(tool-sandbox): resolve command policy paths against the live cwd

### LINKED ISSUE #1338: tool-sandbox: mediated command can't access its own working directory
### What happened?

A mediated command's sandbox doesn't cover the directory it actually runs in when that differs from nono's launch dir:

- `.` and `$WORKDIR` in a command policy both resolve to `--workdir` (launch cwd), not the process's live cwd — so the shim's `getcwd()` fails at sandbox init, killing any mediated child the command forks (e.g. `git` forked by `gh`).
- `@git:*` tokens resolve by running git in the launch dir, so they point at the launch repo, not the repo the command is in. Inside a linked worktree this fails with `fatal: not a git repository`.

`.` should mean the directory the command is running in, and `@git:*` should describe the repo it's operating in — while keeping a command's cwd access within the agent's own granted filesystem (no read/write on a directory the agent is denied).

### Steps to reproduce

1. Two git repos A and B, B not under A.
2. `nono run --workdir A` from A, with the profile below.
3. `bash -c "cd B && gh pr view"` (gh forks git).
4. See error.

### Relevant output or logs

```shell
failed to run git: nono: Sandbox initialization failed: tool-sandbox shim cwd failed: Operation not permitted (os error 1)
```

### nono version

0.66.0 (also current `main`)

### Operating system

macOS

### OS version / distro

macOS 15 (Darwin 25.5.0). Same resolution logic on the Linux/Landlock path.

### Profile (if relevant)

```json
{ "command_policies": { "commands": {
  "gh":  { "can_use": ["git"], "sandbox": { "fs_read": ["."] } },
  "git": { "sandbox": { "fs_read": ["."], "fs_write": ["."] } }
} } }
```

kipz:
## Linked Issue

Closes #1338

## Summary

A mediated command's `command_policies` paths now resolve against its live working directory: `.` and relative `fs_*` entries resolve to the command's cwd, and `@git:*` tokens resolve against the repo the command is running in, while `$WORKDIR` stays the launch `--workdir`. Previously both were pinned to the launch dir, so a mediated command (and any mediated child it forked, e.g. `git` forked by `gh`) failed at sandbox init (`getcwd`) or with `fatal: not a git repository` when run outside the launch dir — e.g. a sibling git worktree.

To keep this from widening filesystem reach, a command's cwd is admitted only if it lies within the launching agent's own effective (allow − deny) filesystem; a cwd outside it (e.g. a credential dir the agent is denied) is rejected, and cwd write grants are capped at the agent's own write access. The admission check is a single shared helper used by both platforms.

## Agent Disclosure

Authored by an AI coding agent. Agent status, intent, and approach were disclosed in #1338. Areas consulted: the macOS (Seatbelt) and Linux (Landlock) tool-sandbox child-launch paths, policy path/token resolution, and capability deny-path resolution. All logic is newly written; no external code was adapted.

## Test Plan

- Added unit test `resolve_policy_path_relative_uses_live_cwd_and_workdir_var_uses_workdir`.
- `make ci` (clippy `-D warnings`, fmt-check, tests, lint-docs) passes.
- Manual (macOS): a tool forking `git` run in a directory outside the launch dir (sibling worktree) now succeeds; the same run in a directory the agent is denied (`~/.ssh`, `~/.aws`) is rejected at admission; same-dir operation is unaffected.

## Checklist

- [x] An issue already exists


crates/nono-cli/src/tool-sandbox/mod.rs
```diff
@@ -62,6 +66,54 @@ pub(crate) struct ToolSandboxPrepare<'a> {
...
+/// cwd write access at the agent's own (write non-escalation).
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+pub(crate) fn admit_command_cwd(
+    command: &str,
+    cwd: &std::path::Path,
+    policy_root: &std::path::Path,
+    outer_caps: &nono::CapabilitySet,
+    deny_paths: &[std::path::PathBuf],
+) -> nono::Result<bool> {
+    let cwd_denied = deny_paths.iter().any(|deny| cwd.starts_with(deny));
```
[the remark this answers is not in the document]
kipz: Already covered. `add_deny_access_rules` (https://github.com/nolabs-ai/nono/blob/30f96e952a676b3ce6f36a352a3275c585ded121/crates/nono-cli/src/policy.rs#L704-L719) pushes both the raw deny path and its canonicalized form into the deny list, so `/etc/secret` and `/private/etc/secret` are both denied. Not a bypass.

kipz:
Replying to the 3 findings from the nogent review that don't have a line anchor:

**linux.rs:3051 / macos.rs:1824 (admit_command_cwd traversal)**: already handled. `build_child_launch_spec` canonicalizes `cwd` before it ever reaches `admit_command_cwd`:
- https://github.com/nolabs-ai/nono/blob/30f96e952a676b3ce6f36a352a3275c585ded121/crates/nono-cli/src/tool-sandbox/platform/linux.rs#L3050-L3054
- https://github.com/nolabs-ai/nono/blob/30f96e952a676b3ce6f36a352a3275c585ded121/crates/nono-cli/src/tool-sandbox/platform/macos.rs#L1823-L1827

So `..` traversal in the raw `request.cwd` is resolved before any `starts_with` check runs. Not a bypass.

**macos.rs:2046 (Seatbelt profile injection via cwd)**: also already handled. `add_macos_cwd_metadata_rules` escapes every path through `escape_seatbelt_path`, which rejects control characters and escapes quotes/backslashes:
- https://github.com/nolabs-ai/nono/blob/30f96e952a676b3ce6f36a352a3275c585ded121/crates/nono-cli/src/policy.rs#L369-L384

Canonicalization isn't what protects against injection here, the escaping is, and it's already in place. Not a bypass.

The 4th finding (dynamic_providers.rs:61, git shelling out in the agent's cwd) is real, fixing that one in a follow-up commit.

kipz:
Fixed the 4th finding (\`dynamic_providers.rs:61\`, git shelling out in the agent's live cwd) in https://github.com/nolabs-ai/nono/commit/6ce28d6f. Replaced the \`git rev-parse --show-toplevel\`/\`--git-common-dir\` spawns with a pure-Rust ancestor walk over \`.git\` entries, so a hostile \`.git/config\` in the agent's cwd can no longer execute anything on the supervisor side. Added regression tests that stub \`git\` on \`PATH\` and assert it's never invoked for toplevel/common-dir/worktree resolution.

crates/nono-cli/src/tool-sandbox/mod.rs
```diff
@@ -62,6 +66,54 @@ pub(crate) struct ToolSandboxPrepare<'a> {
...
+/// owns its `--workdir` (`policy_root`); any other `cwd` must be within the
+/// agent's read grants and not under any deny path (the agent's broad allow can
+/// otherwise cover a denied subtree). This keeps the live-cwd resolution of
+/// `.`/`@git:*` from ever handing a command filesystem reach the agent lacks —
+/// e.g. steering a network-capable command into a credential directory the
+/// agent is denied. Errors (rejecting the command) when the cwd is outside the
+/// agent's granted filesystem; the returned bool lets callers cap the command's
+/// cwd write access at the agent's own (write non-escalation).
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+pub(crate) fn admit_command_cwd(
```
SequeI, line 97: We should add some tests `admit_command_cwd` is currently not exercised, just to ensure going forward, no future commits perhaps might break behaviour.

SequeI (changes requested):
LGTM overall, just left a nit to add more tests to cover the main function here.

```
- cwd under a deny path → rejected
- cwd outside the agent's caps → rejected
- cwd outside workdir but the caps cover it → admitted
- cwd within workdir → always admitted
- admitted but agent can't write cwd → returns false (write downgrade kicks in)
```
Might be good to cover some of these scenarios, again just to ensure no one breaks it by accident in the future. Thanks!

crates/nono-cli/src/tool-sandbox/mod.rs
kipz: Added unit test coverage for `admit_command_cwd` in `ff27a924` (this branch) — covers: cwd under the agent's workdir (writable), cwd in a read-only grant outside the workdir (admitted, not writable), cwd in a read-write grant outside the workdir (writable), cwd outside all grants (rejected), and deny-path precedence over both a broad allow grant and the workdir itself (rejected in both cases).

While addressing this I also found (via a codex pre-push review) that the write-access downgrade in `add_policy_fs` was capping every cwd-scoped `fs_write`/`fs_write_file` grant using a single boolean for the whole cwd, and that the comparison wasn't normalizing `..` segments before the prefix check. Fixed both in `7c393a22` and `d8f29d2d`, each with their own unit tests.

SequeI (approved):
lgtm, thank you!

PR #1344, opened 2026-07-06
fix(cli): match intercept args after global options

### LINKED ISSUE #1328: Intercept rule matching uses hand-rolled strict-prefix argv comparison, bypassable via global flags before subcommand
### What happened?

`resolve_intercept_action` (`crates/nono-cli/src/tool-sandbox/policy.rs`)
matches an `InterceptRuleConfig.args` pattern against the shim's argv via
a hand-rolled strict-prefix comparison starting at index 0 of
`argv[1..]`:

```rust
    && rule
        .args
        .iter()
        .zip(shim_args.iter())
        .all(|(expected, actual)| expected.as_bytes() == *actual)
{
    ...
}
```

This matcher is local to intercept rules — it doesn't reuse the
`ArgvMatcherConfig` type (`exact` / `prefix` / `contains`) already used
for `invocation_policy` rules in `command_policy.rs`, so intercept rules
have no `contains`-style option.

Because matching assumes the subcommand is always `shim_args[0]`, an
intercept rule (e.g. `args: ["clone"]`, carrying a narrowed `sandbox`
override or a `deny`/`approve` action) matches `git clone ...` but not
`git -c protocol.ext.allow=always clone ...`, since `shim_args[0]` is
`-c`. When no rule matches, resolution falls through to
`ResolvedInterceptAction::passthrough()` — no sandbox override, no
intercept action — so the broader default sandbox applies and any
gating on that subcommand is silently skipped.

Pre-existing behavior (confirmed unchanged vs. `main` via `git show`,
originally added in #1105) — not introduced by #1324, which only adds
the `sandbox` field without touching the matcher. Flagged by automated
review on #1324: https://github.com/nolabs-ai/nono/pull/1324#discussion_r3505900348

An intercept rule targeting a subcommand should match regardless of
global flags/options inserted before that subcommand, so its
`sandbox`/`deny`/`approve` action can't be bypassed by argv shape alone.

### Steps to reproduce

1. Configure an intercept rule for a command (e.g. `git`) with
   `args: ["push", "--force"]` and a `deny` action or narrow `sandbox`
   override.
2. Invoke `git -c foo=bar push --force ...` under nono's tool-sandbox.
3. Observe the rule doesn't match; the invocation falls through to the
   catch-all/passthrough action and default sandbox instead.

### Relevant output or logs

N/A — logic/design issue found via code review, not a crash.

### nono version

dev build at 832224b3 (branch `kipz/per-intercept-sandbox`, PR #1324)

### Operating system

Other (specify in additional context)

### OS version / distro

N/A — bug is in argv comparison, not the platform-specific sandbox backend.

dpersek:
## Linked Issue

Closes #1328

## Summary

- Match intercept rule `args` as a contiguous sequence within `argv[1..]`, so command-global options before a subcommand do not bypass a configured intercept action.
- Add regression coverage for `git -c foo=bar push --force` against an intercept rule for `["push", "--force"]`.
- Update the Rust docs, profile schema descriptions, and Tool Sandbox guide from strict-prefix wording to contiguous argument-sequence wording.

## Agent Disclosure

This PR was prepared with assistance from an AI coding agent and reviewed by the human contributor before submission.

Relevant files and sections consulted:
- `AGENTS.md` Coding Agent Contribution Policy, coding standards, and security considerations.
- `NOGENT.md` tool-sandbox/security review guidance.
- `SECURITY.md` responsible disclosure guidance.
- `.github/pull_request_template.md` PR and agent-compliance requirements.
- `crates/nono-cli/src/tool-sandbox/policy.rs` intercept and invocation policy matching code.
- `crates/nono-cli/src/command_policy.rs` `InterceptRuleConfig` docs and validation context.
- `docs/cli/features/tool-sandbox.mdx` Tool Sandbox intercept documentation.
- `crates/nono-cli/data/nono-profile.schema.json` profile schema descriptions.

Compliance confirmation: this change is scoped to the existing public issue, follows the repository's coding and security requirements for the affected area, avoids forbidden `unwrap`/`expect` patterns, and does not change path handling or NonoError behavior.

## Test Plan

- `cargo test -p nono-cli resolve_intercept_action`
- `cargo fmt --all -- --check`
- `make test-cli`
  - Note: a managed-sandbox run failed in unrelated PTY/proxy tests with Unix-socket `Operation not permitted`; the same command passed when rerun outside that sandbox.
- `make clippy`
- `git diff --check`

## Scope / Caveat

This changes intercept `args` matching from strict index-0 prefix matching to command-agnostic contiguous sequence matching within `argv[1..]`. It intentionally does not add command-specific option parsing.

## Checklist

- [x] An issue already exists

### DIFF
--- a/crates/nono-cli/src/tool-sandbox/policy.rs
+++ b/crates/nono-cli/src/tool-sandbox/policy.rs
@@ -45,13 +45,7 @@ pub(super) fn resolve_intercept_action<'a>(
                 sandbox: rule.sandbox.as_ref(),
             };
         }
-        if shim_args.len() >= rule.args.len()
-            && rule
-                .args
-                .iter()
-                .zip(shim_args.iter())
-                .all(|(expected, actual)| expected.as_bytes() == *actual)
-        {
+        if intercept_args_match(&rule.args, &shim_args) {
             return ResolvedInterceptAction {
                 action: &rule.action,
                 rule_args: Some(&rule.args),
@@ -61,6 +55,18 @@ pub(super) fn resolve_intercept_action<'a>(
     }
 
     ResolvedInterceptAction::passthrough()
+}
+
+#[cfg(any(target_os = "linux", target_os = "macos"))]
+fn intercept_args_match(expected_args: &[String], shim_args: &[&[u8]]) -> bool {
+    expected_args.is_empty()
+        || (shim_args.len() >= expected_args.len()
+            && shim_args.windows(expected_args.len()).any(|window| {
+                expected_args
+                    .iter()
+                    .zip(window.iter())
+                    .all(|(expected, actual)| expected.as_bytes() == *actual)
+            }))
 }
 
 /// Resolve the env-expanded helper path and forwarded extra args for an `exec`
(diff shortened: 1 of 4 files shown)


crates/nono-cli/src/tool-sandbox/policy.rs
```diff
@@ -63,6 +57,18 @@ pub(super) fn resolve_intercept_action<'a>(
     ResolvedInterceptAction::passthrough()
 }
 
+#[cfg(any(target_os = "linux", target_os = "macos"))]
```
SequeI: There is no OS specific code here, so this is not required

docs/cli/features/tool-sandbox.mdx
```diff
@@ -111,7 +111,7 @@ An `intercept` rule may carry an optional `sandbox` (the same object described a
 
 ### Intercept Actions
 
-Each `intercept` entry is `{ "args": [...], "action": { "type": ... } }`. Rules are evaluated in order; the first whose `args` are a prefix of the invocation's arguments wins (an empty `args` is a catch-all). The `action.type` selects how the matched subcommand is handled:
+Each `intercept` entry is `{ "args": [...], "action": { "type": ... } }`. Rules are evaluated in order; the first whose `args` appear as a contiguous sequence in the invocation's arguments wins (an empty `args` is a catch-all). This lets subcommand rules match after command-global options, such as `git -c foo=bar push --force`. The `action.type` selects how the matched subcommand is handled:
```
SequeI: Consider a doc note in the schema warning that broad single-element rules (e.g. just ["--force"]) will match anywhere in argv, not just as a subcommand prefix. Might be worthwhile for users to know

SequeI (changes requested):
Left some review comments, thank you for the PR!

SequeI (approved):
lgtm, thank you! Welcome to the community :)

kipz:
Nice fix for the `-c foo=bar push --force` bypass — contiguous-window matching does close that specific gap.

`args` is a literal-string list matched against argv, and the space of argv shapes that produce the "same" effective invocation while dodging that literal comparison is large — global options before the subcommand (this bug), `--flag=value` vs `--flag value`, short vs. long flag aliases (`-f` vs `--force`), repeated/duplicate flags, flag order permutations, etc. Each of these is a plausible follow-up bug report against whatever positional/contiguity assumption the matcher currently makes. Patching the comparator bypass-by-bypass will keep finding new gaps.

There's also a whole separate class around environment variables. Two different shapes:
- Out-of-band env vars that change behavior without appearing in argv at all — e.g. `GIT_SSH_COMMAND`, or a bare `FOO=bar git push` prefix assignment the shell strips before exec. A rule keyed only on argv has no visibility into these.
- Env var references *inside* an argv token — e.g. `git $FOO push` where `$FOO` expands to `--force`. If the matcher inspects argv before shell expansion happens, it sees the literal string `$FOO`, not `--force`, so a rule looking for `--force` never fires even though that's exactly what gets executed. The matcher's view of the command and what actually runs can diverge.

Might be worth treating this as "the matcher needs richer predicate matching" (and possibly needs to observe post-expansion argv, not pre-expansion) rather than a series of point fixes — each gap keeps showing up as its own bypass report otherwise.

PR #1364, opened 2026-07-06
fix(cli): add explicit intercept match predicates

### LINKED ISSUE #1328: Intercept rule matching uses hand-rolled strict-prefix argv comparison, bypassable via global flags before subcommand

dpersek:
## Linked Issue

Closes #1328

## Summary

Intercept rules can now use an explicit `match` predicate instead of relying only on legacy literal `args` windows. This gives profiles a deliberate way to match observed argv shapes with `exact`, `prefix`, or `contains`, and to match out-of-band behavior through filtered child environment variables such as `GIT_SSH_COMMAND`.

The legacy `args` form remains backward-compatible, including first-match ordering and `args: []` catch-all behavior. The new predicate form is intentionally bounded: it does not add a git parser, flag alias folding, flag reordering, or shell-style expansion of literal argv tokens.

## Agent Disclosure

This PR was prepared by Dustin Persek with assistance from an AI coding agent. I reviewed the generated code and tests before submission.

References consulted:
- `CONTRIBUTING.md`
- `CLAUDE.md`
- `NOGENT.md`
- `SECURITY.md`
- `.github/PULL_REQUEST_TEMPLATE.md`
- Issue #1328 discussion and the follow-up context from PR #1344

Compliance notes:
- Issue #1328 already exists.
- The broader approach was disclosed in the issue discussion before this follow-up implementation.
- The change stays in `crates/nono-cli`, where command policy and tool-sandbox behavior live.
- The diff avoids `unwrap`/`expect` in non-test matcher code and passes strict clippy with `-D clippy::unwrap_used`.
- No reused external code or generated third-party logic was introduced.

## Test Plan

- `cargo fmt`
- `jq empty crates/nono-cli/data/nono-profile.schema.json`
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p nono-cli tool_sandbox::policy::intercept_tests`
- `cargo test -p nono-cli validate_intercept`
- `cargo test -p nono-cli empty_argv`
- `cargo test -p nono-cli --test schema_shape`
- `cargo check -p nono-cli`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::unwrap_used`
- `cargo test -p nono-cli -- --test-threads=1`

Note: a local Linux cross-target check was blocked because this macOS Homebrew Rust toolchain does not have the `x86_64-unknown-linux-gnu` target std/core installed.

## Scope / non-goals

- Keeps `args` as the backward-compatible literal/contiguous shorthand.
- Does not add command-specific semantic normalization for `--flag=value` vs `--flag value`, short/long aliases, duplicate flags, reordered flags, or git-specific parsing.
- Does not expand environment references embedded inside argv tokens.
- Does not refactor broader platform launch/env caching.

## Checklist

- [x] Release note has been added to `CHANGELOG.md` if needed; not applicable for this narrow bug/config-surface fix

## Agent Compliance Check

- [x] An issue already exists
- [x] I provided required attribution for reused or adapted code; no external code was reused or adapted
- [x] I did not use forbidden patterns such as unwrap/expect in non-test matcher code
- [x] I used `NonoError` where required
- [x] I validated all relevant matcher input through command-policy validation and schema coverage
- [x] This PR matches the disclosed issue scope


SequeI:
cc @kipz might be worth a look whenever :)

kipz:
@SequeI and @SyntaxSawdust - looks good to me FWIW. This leaves open the possibility of other matchers (e.g. regex, negation etc) which might be handy too in the future.

SequeI (approved):
lgtm, thank you for the contribution!

PR #1378, opened 2026-07-07
fix: missing ~/.cache on macOS

### LINKED ISSUE #1377: macOS user cache access missing
### What happened?

uv uses `~/.cache/uv` instead of `~/Library/Cache/uv` for caching on macOS. Similarly, Corepack uses `~/.cache/node/corepack`.

In summary, some programs do not use macOS convention for caching directory, but use `/~.cache/` to keep the path consistent between Linux and mac.

`~/.cache` be accessible (readwrite) in the `user_caches_macos` profile group

### Steps to reproduce

1. Start nono with profile group `user_caches_macos`
2. Ask claude to use uv, which cannot reuse downloaded packages in the cache

### Relevant output or logs

```shell

```

### nono version

0.66.0

### Operating system

macOS

### OS version / distro

26.5.1 (25F80)

_No response_

### Profile (if relevant)

```json

```

### Additional context

_No response_

subbyte:
## Linked Issue

Closes #1377

## Summary

Adding the missing readwrite access to `~/.cache` for macOS profile group `user_caches_macos` that some programs requires, e.g., uv.

## Test Plan

Manually tested on macOS.

## Checklist

### DIFF
--- a/crates/nono-cli/data/policy.json
+++ b/crates/nono-cli/data/policy.json
@@ -325,7 +325,8 @@
       "allow": {
         "readwrite": [
           "~/Library/Caches",
-          "~/Library/Logs"
+          "~/Library/Logs",
+          "~/.cache"
         ],
         "read": [
           "~/Library/Preferences"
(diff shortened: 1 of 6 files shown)


SequeI (approved):
This makes sense, certain tooling like uv on mac follows XDG convention, so this would just be reaching parity with the `user_caches_linux` group

SequeI:
New version of clippy, using your PR to fix them at the same time, no fault of yours 👍

PR #1380, opened 2026-07-07
fix(profile): preserve platform_overrides through extends resolution

### LINKED ISSUE #1379: platform_overrides is dropped/clobbered across an extends chain
## What happened

`platform_overrides` doesn't survive `extends` correctly:

1. **Dropped entirely** if declared on a base profile rather than the leaf. `merge_profiles` (used by `resolve_extends`) unconditionally sets `platform_overrides: None` on its output. `apply_platform_overrides` only runs afterward, in `finalize_profile` — by then the field is already gone.
2. **Clobbered on same-OS conflict.** If both a base and its child declare an override for the same OS, the child's fragment wholesale replaces the base's instead of merging — a child override touching only one field silently drops every other field the base's same-OS override set.

Neither existing unit test (`platform_overrides_merge_adds_filesystem_paths`, `platform_overrides_other_platform_not_applied`) exercises `extends` at all.

## What was expected

`platform_overrides` should apply after `extends` resolution regardless of which profile in the chain declares it, and same-OS overrides from base and child should deep-merge (child field wins) rather than one replacing the other.

## Steps to reproduce

`shared.json`:
```json
{
  "meta": { "name": "shared" },
  "filesystem": { "read": ["/base/path"] },
  "platform_overrides": {
    "linux": { "filesystem": { "read": ["/platform/path"] } }
  }
}
```
`child.json`:
```json
{ "extends": "shared", "meta": { "name": "child" } }
```
On Linux: `nono profile show --json child.json` — `filesystem.read` has `/base/path` but not `/platform/path`.

Add `"platform_overrides": {"linux": {"open_urls": {"allow_origins": ["https://example.com"]}}}` to `child.json` and re-run — `open_urls` now applies, but `/platform/path` (base's own same-OS override) still doesn't appear.

## nono version

0.67.1 (ae1c513e, the commit that introduced `platform_overrides`); also reproduces on current `main` (1f54f4ae).

## OS / OS version

Reproduces on any host OS (bug is in profile resolution, not sandbox enforcement). Confirmed on macOS 15 (Sequoia) and Linux.

kipz:
## Linked Issue

Closes #1379

## Summary

`platform_overrides` didn't survive `extends`: `merge_profiles` unconditionally nulled it out, and when both a base and its child declared an override for the same OS, the child's fragment replaced the base's instead of merging. Fixed by preserving the field through `merge_profiles` and deep-merging same-OS conflicts with the same child-wins semantics used elsewhere.

## Agent Disclosure

Authored by an AI coding agent under human supervision (@kipz); disclosed in the linked issue.

## Test Plan

- `cargo test --package nono-cli --bin nono platform_overrides` — 8 pass, incl. 2 new regressions (`platform_overrides_on_base_survive_extends_resolution`, `platform_overrides_same_os_key_deep_merges_not_clobbers`) confirmed to fail without the fix.
- `make ci` clean.
- Verified against a real multi-profile hierarchy (shared base + 4 leaf profiles, each with its own same-OS override): `network.open_port` and Linux-only filesystem grants on the base now resolve correctly on both macOS and Linux builds.

## Checklist

## Agent Compliance Check

- [x] An issue already exists

### DIFF
--- a/crates/nono-cli/src/profile/mod.rs
+++ b/crates/nono-cli/src/profile/mod.rs
@@ -3298,11 +3298,50 @@ fn apply_platform_overrides(mut profile: Profile) -> Result<Profile> {
     Ok(merge_profiles(profile, override_profile))
 }
 
+/// Must preserve a base's overrides, not just a leaf's — `merge_profiles`
+/// previously nulled this field unconditionally, so any override on a base
+/// profile was silently lost. Same-OS conflicts deep-merge via
+/// `merge_platform_override_slot` rather than one side replacing the other,
+/// so a child override touching one field can't wipe out unrelated fields
+/// the base's override for that OS set.
+fn merge_platform_overrides(
+    base: Option<PlatformOverrides>,
+    child: Option<PlatformOverrides>,
+) -> Option<PlatformOverrides> {
+    match (base, child) {
+        (None, None) => None,
+        (Some(b), None) => Some(b),
+        (None, Some(c)) => Some(c),
+        (Some(b), Some(c)) => Some(PlatformOverrides {
+            macos: merge_platform_override_slot(b.macos, c.macos),
+            linux: merge_platform_override_slot(b.linux, c.linux),
+            windows: merge_platform_override_slot(b.windows, c.windows),
+        }),
+    }
+}
+
+/// Uses `merge_profiles` itself so a same-OS override composes with the same
+/// child-wins/dedup-append/deny-union rules as everything else in a profile.
+fn merge_platform_override_slot(
+    base: Option<Box<PlatformOverride>>,
+    child: Option<Box<PlatformOverride>>,
+) -> Option<Box<PlatformOverride>> {
+    match (base, child) {
+        (None, None) => None,
+        (Some(b), None) => Some(b),
+        (None, Some(c)) => Some(c),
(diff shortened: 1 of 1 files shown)


kipz:
Fixed in f0f168b0: https://github.com/kipz/nono/commit/f0f168b0ca8a489313970f8f1921484dcdd89eb9

`finalize_profile` now re-runs `validate_profile_custom_credentials`, `validate_env_credential_keys`, and `validate_set_vars` (via `crate::exec_strategy::validate_set_vars`) immediately after `apply_platform_overrides` merges the platform-specific override in, closing the bypass for `custom_credentials`, `env_credentials`, and `environment.set_vars` declared inside `platform_overrides`.

Added regression tests: `platform_overrides_set_vars_validated_after_merge` (negative — a reserved `PATH` key inside a `platform_overrides.<os>.environment.set_vars` block is now rejected) and `platform_overrides_valid_set_vars_merged_and_accepted` (positive — a valid `set_vars` entry declared only inside `platform_overrides` still merges through and finalizes successfully).

SequeI (approved):
Thank you so much! This was an oversight on my part, apologies for forgetting about extends.....

Good thing this was caught out before release. Again, thank you for the contribution :) Tested and validated

PR #1397, opened 2026-07-10
fix(sandbox): grant Refer in execute-restriction layer on Linux

### LINKED ISSUE #1396: Execute-restriction Landlock layer breaks rename into freshly-created dirs
### What happened?

Stacking a second Landlock layer that restricts execute access (`restrict_execute()`) on top of the main sandbox breaks `rename()`/`link()` into directories created *after* that layer is applied — even when source and destination are both under a directory the main layer already grants full read/write on, and both are on the same filesystem. Fails with `EXDEV`.

A rename between two directories that both existed *before* the execute layer was applied works fine.

Root cause: Landlock requires `AccessFs::Refer` granted in every active layer for rename/link to succeed, including layers that never mention the affected paths. `restrict_execute()` only declares `AccessFs::Execute`, so it implicitly denies `Refer` for the whole process.

Rename should succeed — the execute-restriction layer doesn't reference either path, and the main layer already grants full rights there.

### Steps to reproduce

1. Run `nono run` with a profile whose command grants full fs access to a directory `D` and also restricts execute to some other directory (any per-command sandbox that uses `exec_paths` triggers `restrict_execute()`).
2. Inside `D`, create a new subdirectory, then rename a pre-existing file into it.
3. Rename fails with `EXDEV`.

Minimal reproduction is a "download to temp, then atomically rename into a freshly-created destination" pattern entirely within one already-writable directory — e.g. any tool that stages a download in `<dir>/_tmp/` and moves it into `<dir>/<newly-created-subdir>/`.

### Relevant output or logs

```
RENAME FAILED: Invalid cross-device link (errno=18)
```

### nono version

built from main, commit f016b2d5

### Operating system

Linux

### OS version / distro

Ubuntu (container)

6.8.0

kipz:
## Linked Issue

Closes #1396

## Summary

`restrict_execute()`'s Landlock layer only declares `AccessFs::Execute`, which implicitly denies `AccessFs::Refer` for that layer — breaking `rename()`/`link()` into freshly-created directories elsewhere in the sandbox, even when the main ruleset already grants full rights there. This adds a bare `Refer` grant on `/` to the execute-restriction layer, which doesn't widen what that layer permits (it grants nothing without another layer's write/make rights on the same path) but stops it from silently blocking renames established by the main layer.

## Agent Disclosure

This PR was written by an AI agent (Claude, via Claude Code).
- Investigated and diagnosed via a minimal, nono-independent raw-syscall Landlock reproduction (two stacked rulesets) before touching any nono code.
- Ran an independent codex review of this diff; it flagged one finding (unconditional `handle_access(AccessFs::Refer)` could hard-error on ABIs lacking `Refer`). Verified against the `landlock` crate's own `compat.rs` that this is a false positive: the call runs under `CompatLevel::BestEffort` (set on the line above), which returns `Ok(None)` rather than `Err` for unsupported access — mirroring the existing `AccessFs::Execute`/`HardRequirement` pattern just above it, just at a different compat level.
- Files consulted: `crates/nono/src/sandbox/linux.rs` (`restrict_execute`, `apply_landlock`/`apply_auto`), `landlock` crate source (`compat.rs`, `fs.rs`).

## Test Plan

- Added `test_restrict_execute_does_not_break_rename_into_new_subdir` in `crates/nono/src/sandbox/linux.rs`, which forks, applies the main fs ruleset + `restrict_execute()`, then renames a pre-existing file into a subdirectory created after both layers are applied. Confirmed it fails with the bug present (reverted the fix locally, re-ran) and passes with it.
- `make check` (clippy `-D warnings -D clippy::unwrap_used` + fmt-check), `make test`, `make lint-aliases`, `make lint-docs` all pass.
- Verified against the real downstream symptom on Linux: a consuming tool's own daemonizing "download to temp, then rename into a freshly-created install directory" flow previously failed with `EXDEV` under this same layering; rebuilding with this fix resolves it.

## Checklist

- [x] An issue already exists

### DIFF
--- a/crates/nono/src/sandbox/linux.rs
+++ b/crates/nono/src/sandbox/linux.rs
@@ -913,6 +913,10 @@ fn apply_with_abi_inner(
 /// not listed here lose execute permission even if the main sandbox granted it
 /// via `AccessMode::Read`. Read/write grants from the main ruleset are unaffected.
 ///
+/// Also grants bare `Refer` on `/` — Landlock requires it in every layer
+/// for rename/link, so omitting it here silently breaks renames the main
+/// ruleset already permits. Bare `Refer` alone can't widen access.
+///
 /// Call this after `apply_auto()` / `apply_landlock()` to lock down which binaries
 /// an already-sandboxed process can exec.
 pub fn restrict_execute(paths: &[impl AsRef<Path>]) -> Result<()> {
@@ -933,6 +937,12 @@ pub fn restrict_execute(paths: &[impl AsRef<Path>]) -> Result<()> {
             ))
         })?
         .set_compatibility(CompatLevel::BestEffort)
+        .handle_access(AccessFs::Refer)
+        .map_err(|e| {
+            NonoError::SandboxInit(format!(
+                "Tool Sandbox  execute restriction: cannot handle Refer: {e}"
+            ))
+        })?
         .create()
         .map_err(|e| {
             NonoError::SandboxInit(format!(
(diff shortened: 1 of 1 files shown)


SequeI (approved):
lgtm, good fix

PR #1398, opened 2026-07-10
feat: add port range support to sandbox profiles

### LINKED ISSUE #611: Allow domain filtering and unknown ports on localhost

SequeI:
## Linked Issue

Closes #611

## Summary

Adds open_port_range and listen_port_range fields to the profile network config, accepting arrays of [start, end] pairs. Ranges expand to individual rules on both platforms — Seatbelt rules on macOS, Landlock NetPort objects on Linux.

Hard limit of 2^14 (16,384) ports per range on both platforms, shared via PORT_RANGE_LIMIT in capability.rs. macOS crashes with SIGILL above ~17,770 rules (sandbox_init policy compiler limit, empirically determined); Linux has no practical limit but uses the same cap for consistency. Exceeding the limit returns an error rather than silently degrading.

Debug logging emits one line per range at apply time. The full Seatbelt profile string is no longer logged verbatim — only its byte size.

```
{
  "network": {
    "block": true,
    "open_port": [4000],
    "open_port_range": [[3000, 3099], [8000, 8099]]
  }
}
```

## Checklist


crates/nono/schema/capability-manifest.schema.json
```diff
@@ -197,6 +197,16 @@
           "type": "array",
           "items": { "type": "integer", "minimum": 1, "maximum": 65535 },
           "description": "Localhost TCP ports allowed for bidirectional IPC (connect + bind)."
+        },
+        "localhost_range": {
+          "type": "array",
+          "items": {
+            "type": "array",
+            "items": { "type": "integer", "minimum": 0, "maximum": 65535 },
```
connrg: A port range that starts at 0 — like [[0, 5]] — will stop the macOS sandbox from starting at all. Port 0 isn't a real port there, so the system rejects it. The code that handles single ports already knows to skip 0, but the new range code doesn't

crates/nono-cli/src/profile/mod.rs
```diff
@@ -1516,10 +1516,21 @@ pub struct NetworkConfig {
         alias = "allow_port"
     )]
     pub open_port: Vec<u16>,
+    /// Inclusive port ranges for bidirectional localhost TCP IPC (connect + bind).
+    /// Multiple ranges are supported. Example: `[[3000, 3010], [8000, 8100]]`.
+    /// Each port becomes an individual rule — larger ranges take longer to apply at startup.
+    /// **macOS**: hard limit of 16,384 ports per range (2¹⁴; `sandbox_init()` crashes above ~17,770).
```
connrg: The notes in the code still describe the limit wrong. They say the cap is "16,384 ports per range" and that "Linux has no limit" ;but neither is true. The code adds up all the ranges together against that one 16,384 cap, and Linux enforces it just the same as macOS. You already fixed the wording in the authoring guide, but these two spots still have the old text: profile/mod.rs:1522-1523 and capability.rs:906.

connrg (approved):
LGTM, two non blocking comments. Tested on linux.

connrg (approved):
lgtm

PR #1401, opened 2026-07-10
fix: keep orphaned descendants in supervisor ancestry

### LINKED ISSUE #1399: af_unix_mediation: "pathname" denies all socket syscalls for orphaned/backgrounded processes (fail-closed EPERM)
### What happened?

At first I just had Claude randomly stop being able to make API requests. Then we together narrowed down the problem. With linux.af_unix_mediation: "pathname", any process that backgrounds/daemonizes and outlives its launching parent (cmd &, nohup, double-fork) gets EPERM on every connect/bind/sendto/sendmsg — including ordinary TCP/UDP the profile allows. Foreground processes and already-established connections are unaffected.

An orphaned process's allowed TCP/UDP connect should succeed; only genuinely-disallowed AF_UNIX paths should be denied (and with EACCES, as policy denials are).

### Steps to reproduce

0. Preconditions — Linux with Landlock, and Yama ptrace_scope must be 1 (the common default; at 0 the bug is masked because the memory read succeeds regardless):

  ```
cat /proc/sys/kernel/yama/ptrace_scope     # must print 1
```

  1. Minimal profile — allows network, but turns on pathname AF_UNIX mediation (the trigger): /tmp/repro-afunix.json
```json
  {"meta":{"name":"repro-afunix"},"linux":{"af_unix_mediation":"pathname"},"network":{"block":false},"workdir":{"access":"readwrite"}}
```

  2. Test script in Python: /tmp/repro-afunix.py

```python
import os, socket, time


def connect(tag):
    s = socket.socket()
    s.settimeout(3)
    try:
        s.connect(("1.1.1.1", 443))
        print(f"{tag}: OK", flush=True)
    except OSError as e:
        print(f"{tag}: errno={e.errno}", flush=True)
    finally:
        s.close()


# Foreground connect: stays in the supervisor's ancestry -> allowed.
connect("foreground")

# Double-fork so the grandchild reparents to pid 1 (leaves the supervisor's
# ancestry). Pre-fix, its connect() is denied with EPERM (errno 1) because the
# supervisor can no longer read its /proc/<pid>/mem to classify the syscall.
r, w = os.pipe()
if os.fork() == 0:
    os.setsid()
    if os.fork() == 0:
        time.sleep(0.5)  # let the intermediate parent exit first
        s = socket.socket()
        s.settimeout(3)
        try:
            s.connect(("1.1.1.1", 443))
            msg = "orphan: OK"
        except OSError as e:
            msg = f"orphan: errno={e.errno}"
        finally:
            s.close()
        os.write(w, msg.encode())
        os._exit(0)
    os._exit(0)
os.close(w)
print(os.read(r, 200).decode(), flush=True)
```
3. Run a foreground connect (control) alongside a double-forked orphan connect: 
```bash
nono run --profile /tmp/repro-afunix.json --allow-cwd -- /usr/bin/python3 /tmp/repro-afunix.py
```
4.   Expected vs. actual                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
Actual (buggy):                                                                                                                                                                                                                                                                                                              
- stdout: foreground: OK, then orphan: errno=1                                                                                                                                                                                                                                                                               
- stderr (from -vv): Failed to read sockaddr from seccomp notification (nr=42): Failed to open /proc/<pid>/mem: … (Permission denied)                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                               
Expected: orphan: OK — the orphan's allowed TCP connect should succeed, same as the foreground one.  

### Relevant output or logs

```shell

```

### nono version

nono 0.66.0

### Operating system

Linux

### OS version / distro

Arch Linux

7.0.14-zen1-1-zen

### Profile (if relevant)

```json

```

### Additional context

Nono was installed using cargo

maufl:
## Linked Issue

Closes #1399 

## Summary
On Linux, set `PR_SET_CHILD_SUBREAPER` whenever the supervisor must read child memory (`child_requires_dumpable`), so such descendants reparent to the supervisor and stay in its ancestry, and reap them so they don't linger as zombies.

This PR was opened jointly by Claude and me. Claude wrote the code and I reviewed it and asked for changes to the comments. I filled the template manually.

## Test Plan

A new test is added which failed before the change and succeeded after. Furthermore, I manually executed the reproduction steps in the linked issue with the version 0.66.0 and the version built with these changes to verify that it fixes the issue.

## Checklist

Not applicable

### DIFF
--- a/crates/nono-cli/src/exec_strategy.rs
+++ b/crates/nono-cli/src/exec_strategy.rs
@@ -799,12 +799,20 @@ pub fn execute_supervised<F: FnMut(i32) -> bool>(
     // the parent hardens itself immediately after fork and the child hardens
     // itself after sandbox/filter setup whenever procfs inspection is not
     // required.
+
+    // Become a child-subreaper whenever the supervisor may need to read a
+    // descendant's /proc/<pid>/mem: tool-sandbox command mediation
+    // (`tool_sandbox_runtime`) or seccomp-notify mediation of network/AF_UNIX/
+    // openat (`child_requires_dumpable`). That read is ancestry-gated by
+    // ptrace_may_access under Yama ptrace_scope=1, so a descendant that
+    // daemonizes and reparents to pid 1 leaves our ancestry and gets its
+    // connect()/bind()/openat() denied with EPERM. Subreaping keeps it ours.
     #[cfg(target_os = "linux")]
-    if config.tool_sandbox_runtime.is_some() {
+    if config.tool_sandbox_runtime.is_some() || config.seccomp_policy.child_requires_dumpable() {
         let ret = unsafe { libc::prctl(libc::PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0) };
         if ret != 0 {
             return Err(NonoError::SandboxInit(format!(
-                "Failed to set PR_SET_CHILD_SUBREAPER for tool-sandbox supervisor: {}",
+                "Failed to set PR_SET_CHILD_SUBREAPER for supervisor: {}",
                 std::io::Error::last_os_error()
             )));
         }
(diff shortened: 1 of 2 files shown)


SequeI (approved):
lgtm, thank you very much for this fix! Welcome to the community :)

PR #1402, opened 2026-07-10
fix(profile): omit inheritable Option fields when None on save (#1400)

### LINKED ISSUE #1400: Save flow writes `command_policies: null`, producing unparseable profiles
## Summary

The post-run profile **save flow** (the "grant denied paths to a user profile" prompt) writes files that **immediately fail to parse**, because the serializer emits `"command_policies": null` and a custom deserializer rejects that value by design. The result is a self-inflicted round-trip failure: write succeeds (atomic rename), then the file cannot be loaded.

## Environment

- nono 0.67.1
- macOS (Seatbelt)
- Observed while running pi under `--profile always-further/pi`

## Reproduction

1. Have no user profile yet (clean state).
2. Run a profile that triggers a path denial, e.g.:
   ```bash
   nono run --profile always-further/pi -- pi
   ```
   pi tries to read `~/AGENTS.md`, which the pack does not grant.
3. On exit, accept the **grant** option in the save-denied-paths prompt and save under profile name `pi`.
4. The file `~/.config/nono/profiles/pi.json` is written, but it contains:
   ```json
   "command_policies": null,
   ```
5. Validate / reload:
   ```
   $ nono profile validate ~/.config/nono/profiles/pi.json
   nono profile: validating .../pi.json
     [err]  JSON syntax error: Profile parse error: command_policies cannot be null; omit the field or use an empty object on line 1 column 1
     Result: invalid (1 error)
   nono: Profile parse error: validation failed
   ```
6. Any subsequent `nono run --profile pi` fails to load.

User-visible symptom: "saving throws an error, yet the profile is updated on disk." That is exactly write-then-fail-to-reparse.

## Root cause

The save path serializes the `Profile` struct:

- `crates/nono-cli/src/profile_save_runtime.rs:606`
  ```rust
  let profile_json = serde_json::to_string_pretty(&prepared.profile)
  ```

In the `Profile` struct (`crates/nono-cli/src/profile/mod.rs`, "A complete profile definition"), several **inheritable** `Option` fields are declared with only `#[serde(default)]` and **no** `skip_serializing_if`:

- `environment: Option<EnvironmentConfig>` (~line 2112)
- `command_policies: Option<CommandPoliciesConfig>` (~line 2114)
- `open_urls: Option<OpenUrlConfig>` (~line 2197)
- `allow_launch_services: Option<bool>` (~line 2201)
- `allow_gpu: Option<bool>` (~line 2206)
- `allow_parent_of_protected: Option<bool>` (~line 2209)
- `binary: Option<String>` (~line 2212)

So `None` serializes to `null` for all of them (only `platform_overrides` has `skip_serializing_if = "Option::is_none"`). Most of these fields accept `null` on read (plain `Option` + `default`), so they silently round-trip. But `command_policies` uses a custom deserializer that **intentionally rejects null**:

- `crates/nono-cli/src/profile/mod.rs:2065` (`deserialize_command_policies`):
  ```rust
  "command_policies cannot be null; omit the field or use an empty object",
  ```

That rejection is **by design** and is covered by an existing test (`test_command_policies_null_rejected`, `crates/nono-cli/src/profile/mod.rs:7313`): an explicit `null` is treated as a user error — the intent is to *omit* the field to inherit. The bug is therefore on the **write side**, not the read side: when saving a resolved profile where `command_policies` is `None`, nono must omit the field rather than emit `null`.

The same "omit, don't emit null" semantics already applies to these fields conceptually. The profile-authoring docs (`docs/cli/features/profile-authoring.mdx`) state that fields like `open_urls`, `allow_launch_services`, and `allow_gpu` are intentionally omitted from generated skeletons precisely because emitting `null`/`Some` would override inherited behavior. A saved profile that emits `null` for these is also semantically wrong (it means "clear inherited value" rather than "inherit"), even though it currently happens to parse.

## Proposed fix

Add `skip_serializing_if = "Option::is_none"` to the inheritable `Option` fields in the `Profile` (serialize) struct — at minimum `command_policies`, ideally the full set above for correctness and cleaner saved files. Example for the field that causes the hard failure:

```rust
// crates/nono-cli/src/profile/mod.rs, in struct Profile
#[serde(default, skip_serializing_if = "Option::is_none")]
pub command_policies: Option<CommandPoliciesConfig>,
```

This keeps the existing read-side null rejection intact (no change to `deserialize_command_policies` or its test) and fixes the round-trip so saved profiles are valid and inherit-correct.

A regression test should assert that serializing a `Profile` with `command_policies: None` (and the other inheritable fields `None`) produces JSON **without** the corresponding keys, and that the output round-trips through `Profile::deserialize` successfully.

## Attribution

This issue was filed by an AI coding agent (pi) at the direction of the repository owner. The analysis is based on reading the source in the local checkout at `crates/nono-cli/src/profile/mod.rs` and `crates/nono-cli/src/profile_save_runtime.rs` and reproducing the failure with `nono profile validate` on the file produced by the save flow.

dselivanov:
## Summary

Fixes #1400 — the post-run profile **save flow** writes files that fail to reload.

When a run triggers the "grant denied paths to a user profile" prompt, nono serializes a resolved `Profile` and writes it to disk (`profile_save_runtime.rs`). Seven inheritable `Option` fields on the `Profile` serialize struct had only `#[serde(default)]` and no `skip_serializing_if`, so `None` serialized to `null`:

```
environment, command_policies, open_urls, allow_launch_services,
allow_gpu, allow_parent_of_protected, binary
```

This is a write-then-fail-to-reparse round-trip bug. `command_policies` uses a custom deserializer that **intentionally rejects** an explicit `null` (`deserialize_command_policies`, guarded by `test_command_policies_null_rejected`). So every save where the active profile has no `command_policies` wrote `"command_policies": null` — a file the loader refuses. User-visible symptom: *"saving throws an error, yet the profile is updated on disk"* (the file is written via atomic rename, then fails to reparse). Reproduced via `nono profile validate` on a file produced by the save flow.

The other six fields parse fine on reload (plain `Option` + `default` maps `null` → `None`), but emitting `null` is semantically wrong for them too: their documented meaning is **"omit to inherit"**, while `null`/`Some` means "clear/replace". A saved profile emitting `null` would clear an inherited value rather than inherit it.

## Approach

Add `skip_serializing_if = "Option::is_none"` to all seven fields on the `Profile` serialize struct — matching `platform_overrides`, which already had it. This is a write-side-only change: the deserialize side and its null-rejection test are untouched, so the existing `test_command_policies_null_rejected` still enforces that an explicit `null` is a user error (the intent is to *omit* the field to inherit).

Minimal, mechanical diff: 7 attribute additions in `crates/nono-cli/src/profile/mod.rs`, plus a regression test.

## Verification

```
make build-cli          # ok (rustc 1.97)
make fmt-check          # ok
make clippy             # ok (-D warnings -D clippy::unwrap_used)
```

Targeted tests:
- `test_inheritable_options_omitted_when_none_and_round_trip` (new) — **ok**. Asserts a default `Profile` serializes without the seven keys and round-trips through `Profile::deserialize`.
- `test_command_policies_null_rejected` (existing) — **ok**. Null rejection on read unchanged.

Broader: `profile::*` (443), `profile_save_runtime::*` (40), and `learn::*` (35) test groups all pass, 0 failures — the serialization change does not regress the save/learn paths that consume it.

## Files consulted

- `crates/nono-cli/src/profile/mod.rs` — `Profile` serialize struct (line ~2085), `deserialize_command_policies` (line ~2065), null-rejection test (line ~7311).
- `crates/nono-cli/src/profile_save_runtime.rs` — `write_profile` serialize call (line ~606).
- `docs/cli/features/profile-authoring.mdx` — documented "omit to inherit" semantics for `open_urls` / `allow_launch_services` / `allow_gpu`.

---

## Agent Compliance Check

- [x] I am not prohibited from contributing under this policy — this is user-directed work on a real bug, not an OpenClaw/Pi contributor-presence campaign.
- [x] An issue already exists — #1400.
- [x] I described my intent and approach in the issue discussion — #1400 contains the repro, root cause, and proposed fix.
- [x] I reviewed repository coding and security rules for the affected area — read the project `AGENTS.md` (error handling, no `unwrap`/`expect`, surgical changes) and `docs/cli/features/profile-authoring.mdx`.
- [x] I provided required attribution for reused or adapted code — no external code adapted; all changes are newly written against the existing `Profile` struct and its existing `platform_overrides` precedent in the same file.
- [x] I did not use forbidden patterns such as unwrap/expect — no `.unwrap()`; the regression test uses `.expect("…")` in test code only, matching the established pattern in the surrounding `tests` module (e.g. existing `serde_json::from_str(json).expect("parse")`). `make clippy` (strict, `-D warnings -D clippy::unwrap_used`) passes clean.
- [x] I used NonoError where required — no error-handling code added; change is `#[serde(...)]` attributes only.
- [x] I validated and canonicalized all relevant paths — no path handling in this change.
- [x] This PR matches the approved or disclosed issue scope — #1400 proposes exactly this fix (skip-serialize on `command_policies`, plus the same for the sibling inheritable fields).

**Statement:** I am an AI coding agent (pi). This contribution was prepared at the explicit direction of the repository contributor (`dselivanov`) to fix a bug they personally encountered. It is not part of any contributor-presence campaign.

### DIFF
--- a/crates/nono-cli/src/profile/mod.rs
+++ b/crates/nono-cli/src/profile/mod.rs
@@ -2108,9 +2108,9 @@ pub struct Profile {
     /// ALIAS(canonical="env_credentials", introduced="v0.0.0", remove_by="indefinite", issue="#143")
     #[serde(default, alias = "secrets")]
     pub env_credentials: SecretsConfig,
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub environment: Option<EnvironmentConfig>,
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub command_policies: Option<CommandPoliciesConfig>,
     #[serde(default)]
     pub credential_capture: HashMap<String, CredentialCaptureEntry>,
@@ -2134,22 +2134,22 @@ pub struct Profile {
     /// When `None` (absent from JSON), inherits from the base profile.
     /// When `Some`, replaces the base profile's config entirely, allowing
     /// derived profiles to narrow permissions.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub open_urls: Option<OpenUrlConfig>,
     /// Opt-in gate for temporary direct LaunchServices opens on macOS.
     /// Must be paired with the CLI flag `--allow-launch-services`.
     /// When `None`, inherits from the base profile.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub allow_launch_services: Option<bool>,
     /// Opt-in gate for GPU access (Metal/IOKit on macOS, render nodes on Linux).
     /// Must be paired with the CLI flag `--allow-gpu`.
     /// When `None`, inherits from the base profile.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub allow_gpu: Option<bool>,
     /// Opt-in to allow parent-of-protected-root grants on macOS.
     /// When `true` (and on macOS), `--allow ~` is permitted because Seatbelt deny
     /// rules protect `~/.nono`. Ignored on Linux. Default is `false`.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub allow_parent_of_protected: Option<bool>,
     /// Deprecated: Parsed for backward compatibility but ignored.
     /// Supervised mode preserves TTY by default, making this unnecessary.
@@ -2166,7 +2166,7 @@ pub struct Profile {
     /// Binary path or command name to run when no trailing `-- <command>` is given.
     /// Resolved via `PATH` lookup or canonicalized if absolute. Only honoured
     /// for user-authored profiles (ignored for pack and built-in profiles).
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub binary: Option<String>,
     /// Extra arguments appended to the child command at launch.
     /// Supports variable expansion (e.g. `$NONO_PACKAGES`).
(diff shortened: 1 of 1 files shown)


SequeI (approved):
lgtm, many thanks for this fix! welcome to the community :)

PR #1403, opened 2026-07-10
feat(resources): cap sandbox process count with --max-processes (cgroup v2 pids.max)

connrg:
> Follow-on from #1269 (the `--memory` cap). Same cgroup v2 foundation — this adds the sibling
> process-count limit that #1269 anticipated.

## What this adds

A new `--max-processes` flag that caps **how many processes (and threads) a sandbox is allowed to
run**:

```sh
nono run --max-processes 64 -- ./my-agent
```

Once the sandbox reaches that many processes, the kernel simply **refuses to start new ones** — it
does **not** kill anything already running. That's the key difference from #1269's `--memory` cap,
which kills the whole sandbox when it runs out of memory.

## Why

nono runs untrusted code, and until now there was no way to stop a sandboxed program from spawning
an unbounded number of child processes. A runaway loop or a deliberate "fork bomb" could exhaust
the host's process table and starve everything else on the machine. This gives each run a hard
ceiling on its process count.

It reuses the cgroup machinery landed in #1269, so both limits can be set together:

```sh
nono run --memory 512M --max-processes 64 -- ./my-agent
```

## How it works

- Each run gets its own Linux **cgroup**; we set the leaf's `pids.max` and the kernel enforces it.
- The limit covers the whole process tree — children, grandchildren, threads — so nothing can
  spawn out from under it.
- **Fails closed:** if the cgroup can't be created (delegation missing, or the flag is used where
  it can't be enforced), the run **refuses to start** rather than running without the limit. It
  never silently pretends to enforce a cap.
- The sandboxed program **can't raise or escape its own limit** — a writable grant overlapping the
  cgroup control plane (`/sys/fs/cgroup`) is refused.
- On breach it prints a plain diagnostic ("process limit reached… the kernel refused new
  processes"), not a mysterious failure.

## Usage

- CLI: `--max-processes <N>` on `nono run` (run-only; conflicts with `--config`; rejected under
  `wrap`).
- JSON config manifest: `resources.max_processes` (requires `exec_strategy: supervised`). The JSON
  Schema enforces `minimum: 1`, so `0`/negative are rejected at parse time — matching the CLI
  flag's validation.
- `0` isn't "unlimited" (it would forbid the sandbox from running anything) — omit the flag
  entirely for no limit.

## Platform notes

- Enforced on **Linux** (cgroup v2, with the `pids` controller delegated to the session) — same
  requirement as #1269's memory cap.
- On macOS / other platforms the flag is **rejected** (fails closed), never accepted-but-unenforced.

## Testing

- Unit: flag parsing, profile+flag limit merging, cgroup knob selection, controller-delegation
  checks, `pids_report` parsing, and diagnostic rendering.
- Live kernel tests (`--ignored`): create real cgroups and confirm the cap denies forks, kills
  nothing, and tears down with no leaks (parallel-safe).
- End-to-end: the cap binds and returns promptly, a normal run is unaffected, and `0` /
  conflicting flags are rejected.
- Full suite green; clippy (`-D warnings -D clippy::unwrap_used`) and fmt clean.

Relates to #1102.


SequeI (approved):
lgtm, thanks!

Issue #1414, opened 2026-07-13
Add profile-declared client-side proxy bypass entries


anugrahsinghal:
### What problem are you trying to solve?

When nono runs in proxy mode, it owns the child process proxy environment (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `no_proxy`). That is useful because nono can keep proxy credentials, filtering, and route behavior consistent.

However, some profiles need specific hostnames to be reached directly by the sandboxed process instead of through the local nono proxy. A common case is a local development hostname or profile-owned service alias that resolves to loopback. Today, nono-generated `NO_PROXY` includes loopback defaults such as `localhost` and `127.0.0.1`, but profile authors do not have a first-class way to add additional safe bypass entries.

Using `environment.set_vars.NO_PROXY` is not reliable because proxy env generation happens later and overwrites those values. That leaves profile authors with wrapper-specific workarounds outside nono, and the source of truth for proxy behavior becomes hard to audit.

### What would you like to see?

Add a first-class profile network field, for example:

```json
{ "network": {
    "no_proxy": ["redis", "*.local.test"] }
}
```

Desired behavior:

• Treat `network.no_proxy` as client-side local proxy bypass intent.
• Merge it with generated loopback defaults.
• Emit merged values to `NO_PROXY`  and `no_proxy`.
• Keep parent-shell  NO_PROXY  out of the merge by default.
• Validate entries as host patterns, not URLs or `host:port` values.
• Preserve proxy-controlled routes: entries overlapping credential routes, L7-filtered routes, proxy allowlists, or other proxy-enforced destinations should fail configuration rather than silently bypassing the proxy.
• Keep `allow_domain` distinct from `no_proxy`: `allow_domain` means “allowed through the proxy”; `no_proxy` means “skip the proxy.”
• Optionally emit a nono-owned canonical form such as  NONO_NO_PROXY  so wrappers can translate profile intent to runtime-specific formats like Java  `-Dhttp.nonProxyHosts`.

Example rendering:

```
network.no_proxy = ["redis", "*.local.test"]

NO_PROXY=localhost,127.0.0.1,redis,.local.test
no_proxy=localhost,127.0.0.1,redis,.local.test
NONO_NO_PROXY=localhost,127.0.0.1,redis,*.local.test

```

`network.no_proxy` should not grant network access by itself. It should only influence client proxy selection; the sandbox’s network permissions should still decide whether a direct connection is allowed.

### What have you tried instead?

[...]

PR #1415, opened 2026-07-13
feat: add profile-declared no_proxy bypass support

anugrahsinghal:
## Linked Issue

Relates to nolabs-ai/nono#1414

## Summary

Adds profile-declared client-side proxy bypass support via `network.no_proxy`.

The feature lets profiles declare safe host patterns that should bypass the local nono HTTP proxy for clients that honor `NO_PROXY` / `no_proxy`, while keeping proxy-controlled routes and allowlists fail-secure.

Key behavior:

- Adds `network.no_proxy` to profile schema/runtime propagation.
- Emits merged bypass entries into `NO_PROXY` and `no_proxy`.
- Emits `NONO_NO_PROXY` as a canonical nono-owned rendering for wrappers/SDKs that need runtime-specific translation, such as Java `http.nonProxyHosts`.
- Keeps parent-shell `NO_PROXY` out of the merge.
- Rejects conflicts with proxy routes, `allow_domain`, expanded network policy hosts, and direct `ProxyConfig.allowed_hosts`.
- Keeps route/L7/credential traffic on the proxy path.
- Preserves local-alias use cases such as `/etc/hosts` loopback aliases.

This PR was prepared by an AI coding agent under user direction.

Relevant files/areas consulted:

- `AGENTS.md`
- `crates/nono-cli/src/profile/mod.rs`
- `crates/nono-cli/src/proxy_runtime.rs`
- `crates/nono-proxy/src/config.rs`
- `crates/nono-proxy/src/server.rs`
- `crates/nono-proxy/src/route.rs`
- `docs/cli/features/networking.mdx`

## Test Plan

Automated checks run during development:

- `cargo test -p nono-proxy no_proxy --quiet`
- `cargo test -p nono-cli no_proxy --quiet`
- `cargo check -p nono-proxy -p nono-cli`
- `cargo fmt`
- strict clippy on changed crates during implementation

Manual/local verification:

- Verified a local `/etc/hosts` loopback alias with a temporary local HTTP server.
- A profile with `network.no_proxy` for the alias and an explicit local port grant allowed `curl` inside nono to reach the service directly.
- Verified a Java `HttpURLConnection` sample:
  - with proxy properties set and no Java non-proxy hosts, the request went through the nono proxy and was rejected
  - translating `NONO_NO_PROXY` to `-Dhttp.nonProxyHosts=...` allowed the Java request to reach the local service directly

## Checklist

## Agent Compliance Check

- [x] An issue already exists

Notes:

- Java `http.nonProxyHosts` is not set by nono directly. Wrappers can read `NONO_NO_PROXY` and translate it to runtime-specific formats.
- The public `ProxyConfig` struct now has a `no_proxy` field; this was accepted for this change rather than doing a broader builder/non-exhaustive API refactor.


crates/nono-proxy/src/config.rs
```diff
@@ -334,6 +345,380 @@ fn default_require_auth() -> bool {
...
+            | "local"
+    )
+}
+
+const ALWAYS_DENIED_HOSTS: &[&str] = &[
+    "169.254.169.254",
+    "fd00:ec2::254",
+    "metadata.google.internal",
+    "metadata.azure.internal",
+];
```
anugrahsinghal, line 658: `crates/nono/src/net_filter.rs` uses `DENY_HOSTS`

crates/nono-proxy/src/config.rs
```diff
@@ -334,6 +345,380 @@ fn default_require_auth() -> bool {
...
+            | "int"
+            | "arpa"
+            | "io"
+            | "co"
+            | "dev"
+            | "app"
+            | "ai"
+            | "local"
+    )
+}
```
anugrahsinghal: blocks obvious mistakes, but can remove it too based on feedback, since this is not exhaustive

crates/nono-proxy/src/server.rs
```diff
@@ -199,9 +199,14 @@ pub struct ProxyHandle {
...
-    /// Non-credential allowed hosts that should bypass the proxy (NO_PROXY).
-    /// Computed at startup: `allowed_hosts` minus credential upstream hosts.
+    /// Client-side proxy bypass entries appended after loopback defaults.
+    /// Computed at startup from direct-connect bypasses and profile-declared
+    /// `network.no_proxy` entries, excluding route upstreams.
     no_proxy_hosts: Vec<String>,
+    /// Canonical nono-owned bypass patterns for wrappers/SDKs that need to
+    /// translate profile intent to non-env proxy surfaces such as Java
+    /// `http.nonProxyHosts`.
+    canonical_no_proxy_hosts: Vec<String>,
```
anugrahsinghal, line 212: this is a change specifically so consumers like Java etc that do not support NOPROXY can work with raw declared values, we can remove this based on feedback if we want the client/sdk wrapper to take care of the expansion from NOPROXY itself

my idea with this was, where do I store the cannonical values configured in the profile

example:
`*.internal` is flattened to `.internal` for NOPROXY

but some tools might expect `*.internal`

so this acts as a way to preserver the truth of the config

anugrahsinghal:
another idea I wanted to float was an option to autoconfigure by reading /etc/hosts and resolve loopback addresses in that and then use them as noproxy values

SequeI (commented):
Thanks for this, the implementation is solid and the validation layer is thorough. A few things worth discussing before merge.

`is_broad_single_label_suffix`

The blocklist (`com`, `org`, `io`, `ai`, `local`, etc.) is input hygiene rather than a security control. A `no_proxy` entry only affects which HTTP clients skip the proxy; it does not grant network access -- direct connections still need a matching sandbox permission to succeed. So even `*` in `no_proxy` cannot actually bypass anything the sandbox has not already granted. Given that, I would suggest either removing this list entirely (keeping the validation simple: format checks and metadata IP/host blocks only, which are load-bearing), or adding a comment making clear it is ergonomics not security. As written it looks like a security boundary when it is not, and it will silently reject valid entries like `.internal.corp` since `internal` is not in the list, which will surprise users in Kubernetes environments.

Double validation

`validate_profile_no_proxy` is called in both `parse_profile_bytes` and `finalize_profile`. The finalize call is needed to catch post-merge `allow_domain` conflicts. The parse-time call is just early UX feedback. A short comment explaining that would help, as is it looks like a bug.

`NONO_NO_PROXY` emission path

There is a unit test confirming the value is correct (`test_proxy_env_vars_include_canonical_nono_no_proxy`), but no test tracing it through the full `prepare_sandbox` -> `ProxyLaunchOptions` -> child env pipeline. Given this is the primary user-facing output, worth adding one.

`/etc/hosts` auto-detection

On your comment about auto-reading `/etc/hosts`. I would keep that separate. It is a reasonable follow-up but it changes the trust model (nono would be reading and trusting host-level config at runtime), and the design questions around it are non-trivial. This PR's explicit declaration approach is the right starting point. Once the rebase is in and everything passes, we can probably get this in

anugrahsinghal:
Applied the changes as requested

SequeI (approved):
lgtm, thanks for the changes!

PR #1428, opened 2026-07-16
feat(proxy): add per-route request rate limiting (RouteRateLimiter)

### LINKED ISSUE #1427: Add per-route request rate limiting (RouteRateLimiter)

Compromised sandbox might be used for DoS attacks. 

nono has has already some mitigations for this attack:

- ["Prompt Flooding"](https://nono.sh/docs/cli/features/supervisor#rate-limiting):  protects the LLM providers
- [`max_connections` in ProxyConfig](https://nono.sh/docs/python/api/proxy-config#proxyconfig): prevents parallelization

With the current mitigations you can at best lock down a compromised sandbox to 1 outgoing connections.
The problem is that you can still do a lot of damage over 1 connection ...

This issue proposes that we introduce optional route specific rate limiter to further mitigate the possibility of using compromised sandboxes for DoS attacks.

- ["Prompt Flooding"](https://nono.sh/docs/cli/features/supervisor#rate-limiting):  protects the LLM providers
- [`max_connections` in ProxyConfig](https://nono.sh/docs/python/api/proxy-config#proxyconfig): prevents parallelization

### How is this blocking you?

### Additional context

_No response_

zbleness:
Add an opt-in per-route token-bucket rate limiter to the L7 proxy to contain a runaway or compromised agent. When a route's bucket is empty, requests are delayed up to a bounded max_delay and then rejected with HTTP 429 (bounded throttle, then reject) - never escalated to approval and never queued unboundedly, so it cannot self-DoS via max_connections.

- nono-proxy: RouteRateLimiter token bucket (rate_limit.rs) with bounded delay; RouteRateLimitConfig + RouteConfig.rate_limit; LoadedRoute wiring; enforcement in the reverse-proxy path and shared across the HTTP/1.1 and h2 intercept paths via select_intercept_route.
- Applies only to L7-visible traffic (reverse-proxy and intercepted CONNECT); opaque CONNECT tunnels are out of scope and documented.
- nono-cli: expose rate_limit on custom credential profiles, plumb it into routes, and add it to the profile JSON schema.
- Tests: token-bucket unit tests, reverse-proxy 429 e2e test, and profile parse tests.

## Linked Issue

Closes #1427

## Summary

Add an opt-in per-route token-bucket rate limiter to the L7 proxy to contain a runaway or compromised agent. 
This further improves the DoS mitigation possibilities with nono.

## Test Plan

make ci

## Checklist


crates/nono-proxy/src/config.rs
```diff
@@ -858,6 +870,41 @@ pub enum SpiffeAuthConfig {
     },
 }
 
+/// Per-route request-rate limit configuration for a [`RouteConfig`].
+///
+/// A token bucket refilling at `requests_per_minute` and holding up to `burst`
+/// tokens. When the bucket is empty, overshoot is delayed up to
+/// `max_delay_secs`; a request that would wait longer is rejected with HTTP
+/// 429. See
+/// `docs/adr/0001-route-rate-limiter-bounded-throttle-then-reject.md`.
```
lukehinds: something dangling here, perhaps a local file?

crates/nono-proxy/src/reverse.rs
```diff
@@ -241,6 +241,34 @@ pub async fn handle_reverse_proxy(
         return Ok(());
     }
 
+    // Route request rate limit (RouteRateLimiter). Gate the authorized request
+    // before credential injection and upstream forwarding: bounded delay, then
+    // HTTP 429 — never a human approval, never an unbounded wait. Opaque CONNECT
```
lukehinds: "'the authorized request' is inaccurate here - at this point the request is endpoint-authorized (`enforce_endpoint_policy` above) but not yet authenticated. The localhost session-token check is `if ctx.require_auth` at line 334, ~90 lines below this gate.

crates/nono-proxy/src/reverse.rs
```diff
@@ -241,6 +241,34 @@ pub async fn handle_reverse_proxy(
...
+    // tunnels never reach this reverse-proxy path, so they are unaffected.
+    if let Some(limiter) = route.rate_limiter.as_ref() {
```
lukehinds: This gate runs before the session-token boundary at line 334, so a request with a missing/bad phantom token still consumes a bucket token -> an unauthenticated local caller can drain the bucket and induce 429s for the legitimate agent. I would consider moving the block below the `if ctx.require_auth { … }` block (after line ~380), or maybe even scope it to authenticated requests

crates/nono-proxy/src/tls_intercept/handle.rs
```diff
@@ -759,6 +759,41 @@ pub(crate) async fn select_intercept_route<'a>(
...
+                    &audit::EventContext {
+                        route_id: Some(prefix),
+                        upstream: Some(&route.upstream),
+                        ..audit::EventContext::default()
+                    },
+                    host,
+                    port,
+                    reason,
+                );
+                return RouteSelection::Rejected(429);
```
lukehinds, line 792: This intercept-path on line HTTP response code 429 branch isn't covered by any test. The reverse path has `reverse_proxy_rate_limit_rejects_after_burst` in server.rs, but the existing `select_intercept_route` tests all use `rate_limit: None`, so this branch (and the shared-across-h1/h2 claim) is never actually exercised. Worth adding a `select_intercept_route` test with a burst-exhausted limiter asserting its `Rejected(429)`.

lukehinds (commented):
Nice PR thanks @zbleness , it might look like a lot, but these should be pretty simple fix ups - thaks for the contribution.

crates/nono-proxy/src/config.rs
zbleness: Yupp, sorry fixing.

crates/nono-proxy/src/tls_intercept/handle.rs
zbleness: Done (test added)

crates/nono-proxy/src/reverse.rs
zbleness: Done (moved it down below the authentication checks.

crates/nono-proxy/src/reverse.rs
zbleness: Moved this block below the authentication checks, where the comment is correct.

crates/nono-proxy/src/config.rs
zbleness: Remove the invalid reference to the ADR (in 3 places).

lukehinds (approved):

PR #1437, opened 2026-07-17
fix(proxy): don't cross-deny sibling routes sharing an upstream

### LINKED ISSUE #1436: Sibling `custom_credentials` routes sharing one upstream cross-deny each other (403)
### What happened?

When two (or more) `custom_credentials` routes point at the **same upstream host** but carry **disjoint `endpoint_rules`**, every request to that host is rejected with **403**, even for a path that one of the routes explicitly allows.

`custom_credentials` compiles a legacy `endpoint_rules` allow-list into a default-deny endpoint policy. In `select_intercept_route` (`crates/nono-proxy/src/tls_intercept/handle.rs`), the per-candidate loop evaluates every route sharing the upstream `host:port` and the `EndpointPolicyOutcome::Deny` arm returns `RouteSelection::Rejected(403)` immediately. Candidates are evaluated in sorted-prefix order, so the first route whose allow-list does not cover the request path hard-denies the whole request before the sibling route that *does* allow it is ever considered.

This diverges from the documented reference selector `route::select_route` (whose doc comment says to keep the two in sync), where a credential route whose `endpoint_rules` don't match is simply dropped from consideration and the matching route is selected. See also the intended multi-route semantics asserted in `test_route_selection_multi_org_profile`.

Each request should be authorized by whichever sibling route's `endpoint_rules` match it, and that route's credential injected. A path covered by no route should be an un-credentialed passthrough (the host is already gated by `allow_domain`), not a 403. Only an **explicit** `endpoint_policy` `deny` rule (or an explicitly authored `default: deny`) should hard-deny.

### Steps to reproduce

Profile (`bug.json`):

```json
{
  "extends": "default",
  "meta": { "name": "bug", "description": "Profile for reproducing bug" },
  "network": {
    "block": false,
    "allow_domain": ["example.com"],
    "credentials": ["foo", "bar"],
    "custom_credentials": {
      "foo": {
        "upstream": "https://example.com/foo",
        "credential_key": "env://USER",
        "inject_mode": "header",
        "credential_format": "Basic {}",
        "endpoint_rules": [{ "method": "GET", "path": "/foo" }]
      },
      "bar": {
        "upstream": "https://example.com/bar",
        "credential_key": "env://LOGNAME",
        "inject_mode": "header",
        "credential_format": "Basic {}",
        "endpoint_rules": [{ "method": "GET", "path": "/bar" }]
      }
    }
  }
}
```

Run:

```bash
nono run -v --allow-cwd --profile bug -- curl 'https://example.com/foo'
```

### Relevant output or logs

```shell
l7 endpoint policy decision mode=connect_intercept target="example.com" method="GET" path="/foo" decision=Deny endpoint_policy_action="deny" endpoint_policy_rule="endpoint_policy.default"
403
```

### nono version

0.68.0

### Operating system

macOS

### OS version / distro

26.5.1

_No response_

### Profile (if relevant)

```json
{
  "extends": "default",
  "meta": { "name": "bug", "description": "Profile for reproducing bug" },
  "network": {
    "block": false,
    "allow_domain": ["example.com"],
    "credentials": ["foo", "bar"],
    "custom_credentials": {
      "foo": {
        "upstream": "https://example.com/foo",
        "credential_key": "env://USER",
        "inject_mode": "header",
        "credential_format": "Basic {}",
        "endpoint_rules": [{ "method": "GET", "path": "/foo" }]
      },
      "bar": {
        "upstream": "https://example.com/bar",
        "credential_key": "env://LOGNAME",
        "inject_mode": "header",
        "credential_format": "Basic {}",
        "endpoint_rules": [{ "method": "GET", "path": "/bar" }]
      }
    }
  }
}
```

### Additional context

_No response_

zbleness:
select_intercept_route returned 403 as soon as any route for the upstream evaluated to Deny. A legacy endpoint_rules allow-list compiles to a non-explicit default-deny policy, so when several credential routes share one upstream host with disjoint endpoint_rules, the first route (by prefix order) whose rules didn't match the request path hard-denied the whole request, even though a sibling route would have authorized and injected a credential.

Treat a non-explicit default-deny as "this route does not apply" instead of a terminal 403, mirroring route::select_route: drop a managed-credential route, or let a credential-less _ep_ route gate the request via has_endpoint_only_route so the post-loop check still 403s only when nothing authorized it. Explicit endpoint policies keep their authoritative hard-deny. Fix covers both the HTTP/1.1 and HTTP/2 intercept paths.

Add a regression test for two disjoint credential routes on one upstream.

## Linked Issue

Closes #1436 

## Summary

Fixes #1436

## Test Plan

```shell
make ci 
```

## Checklist

### DIFF
--- a/crates/nono-proxy/src/tls_intercept/handle.rs
+++ b/crates/nono-proxy/src/tls_intercept/handle.rs
@@ -645,6 +645,23 @@ pub(crate) async fn select_intercept_route<'a>(
                 }
             }
             EndpointPolicyOutcome::Deny { reason, rule_label } => {
+                // A legacy `endpoint_rules` allow-list compiles to a
+                // non-explicit default-deny policy. When the request path is
+                // not in that list the route simply does not apply — it must
+                // not hard-deny the whole request, because another route
+                // sharing this upstream may still authorize and inject a
+                // credential. Mirror `route::select_route`: drop a managed-
+                // credential route, or let a credential-less endpoint-only
+                // (`_ep_`) route gate the request via `has_endpoint_only_route`
+                // so the post-loop check produces the 403 only when nothing
+                // authorized it. Explicit endpoint policies keep their
+                // authoritative hard-deny below.
+                if !route.endpoint_policy.is_explicit() {
+                    if !route.requires_managed_credential {
+                        has_endpoint_only_route = true;
+                    }
+                    continue;
+                }
                 let deny_reason = reason.unwrap_or("endpoint denied by policy");
                 audit::log_l7_policy_decision(
                     audit_log,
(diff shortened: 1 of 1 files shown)


SequeI (approved):
Makes sense and LGTM, thanks for the fix!

PR #1440, opened 2026-07-17
feat(tool-sandbox): caller-declared env pass-through via export_env

### LINKED ISSUE #1439: Let a command pass named env vars to the commands it invokes (caller-declared export_env)

A tool-sandboxed command can't pass a specific env var down to a command it invokes. `allow_vars` filters against the callee and the dangerous-var blocklist strips interpreter vars; `set_vars` only injects a static value, not the parent's live one.

Caller-declared pass-through: per-command `export_env` + top-level `session_export_env`. Named vars (exact, `PREFIX_*`, or `*`) are copied verbatim from the invoked command's parent env, bypassing `allow_vars` and the blocklist. `PATH`/`NONO_*` always excluded. The caller controls its own exposure; safe because the parent's env is already sanitized at launch.

```json
{
  "command_policies": {
    "session_export_env": ["SOME_VAR"],
    "commands": { "tool-a": { "export_env": ["OTHER_*"] } }
  }
}
```

`allow_vars` (callee-scoped; blocklist still strips). `set_vars` (static value only).

## How is this blocking you?

## Additional context

Applies to both macOS and Linux tool-sandbox paths.

kipz:
## Linked Issue

Closes nolabs-ai/nono#1439

## Summary

Caller-declared env pass-through: per-command `export_env` + top-level `session_export_env`. A command's named vars are copied verbatim to commands it invokes, bypassing `allow_vars` filtering and the dangerous-var blocklist. Patterns: exact, `PREFIX_*`, `*`. `PATH`/`NONO_*` always excluded.

Authored by an AI agent (code, tests, PR, issue). No `unwrap`/`expect`/`panic` in new non-test code; covered by tests.

## Test Plan

* `make ci` passes (clippy `-D warnings`, fmt, tests, doc lint).
* Unit tests: verbatim pass-through; `PREFIX_*` and `*` matching; `PATH`/`NONO_*` excluded under `*`; reserved/malformed patterns rejected.
* Verified end-to-end on Linux: `session_export_env` and a mediated command's `export_env` each deliver a named var downstream that default filtering would drop.

## Checklist

- [X] An issue exists and is linked above
- [X] All commits are signed-off, using [DCO](<https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin>)
- [X] All new code follows the project's coding standards ([CLAUDE.md](<CLAUDE.md>)) and is covered by tests
- [X] Public-facing changes are paired with documentation updates

- [X] I am not prohibited from contributing under this policy
- [X] An issue already exists
- [X] I disclosed that I am an agent in the issue discussion
- [X] I described my intent and approach in the issue discussion
- [X] I reviewed repository coding and security rules for the affected area
- [X] I provided required attribution for reused or adapted code
- [X] I did not use forbidden patterns such as unwrap/expect
- [X] I used NonoError where required
- [X] I validated and canonicalized all relevant paths
- [X] This PR matches the approved or disclosed issue scope

### DIFF
--- a/crates/nono-cli/src/tool-sandbox/env.rs
+++ b/crates/nono-cli/src/tool-sandbox/env.rs
@@ -116,6 +116,49 @@ pub(crate) fn apply_environment_set_vars(
     Ok(())
 }
 
+/// Copies `export_patterns`-matching vars from `request.env` into `env` verbatim,
+/// bypassing `allow_vars` and the dangerous-var blocklist. `PATH`/`NONO_*` and
+/// secrets are always excluded; loader vars need an exact-name pattern.
+pub(crate) fn apply_export_env(
+    env: &mut Vec<Vec<u8>>,
+    request: &ToolSandboxShimRequest,
+    export_patterns: &[String],
+) {
+    if export_patterns.is_empty() {
+        return;
+    }
+    for entry in &request.env {
+        let Some((key, value)) = split_env_entry(entry) else {
+            continue;
+        };
+        let Ok(key_str) = std::str::from_utf8(key) else {
+            continue;
+        };
+        // nono owns PATH and its own NONO_* namespace; never let the export
+        // list re-admit them, including under a bare `*`.
+        if key_str == "PATH" || key_str.starts_with("NONO_") {
+            continue;
+        }
+        if !crate::exec_strategy::matches_env_var_patterns(key_str, export_patterns) {
+            continue;
+        }
+        if crate::exec_strategy::env_sanitization::is_forbidden_secret_env_var(key_str) {
+            continue;
+        }
+        if crate::exec_strategy::env_sanitization::is_loader_injection_env_var(key_str)
+            && !export_patterns.iter().any(|pattern| pattern == key_str)
+        {
+            continue;
+        }
+        let mut prefix = key.to_vec();
+        prefix.push(b'=');
+        env.retain(|existing| !existing.starts_with(&prefix));
+        let mut new_entry = prefix;
+        new_entry.extend_from_slice(value);
+        env.push(new_entry);
+    }
+}
+
 pub(crate) fn inject_chaining_control_env(
     env: &mut Vec<Vec<u8>>,
     socket_path: &Path,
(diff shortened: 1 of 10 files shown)


lukehinds:
hey @kipz , just a couple of fix ups

The new `session_export_env` and `export_env` fields also need to be added to the published profile schema. The embedded JSON Schema uses `additionalProperties: false`, so it could currently rejects any profile using these fields, even though the runtime accepts them. I think the corresponding expectations in schema_shape.rs should be updated too.

While there worth getting the validation to also reject patterns containing repeated trailing wildcards. For example, `A**` currently passes validation, but the runtime matcher silently ignores it, so the profile loads successfully without exporting the expected variables. I would have this enforce the same single-wildcard rule already used for `allow_vars`

kipz:
Makes sense @lukehinds - thanks! All done.

crates/nono-cli/src/tool-sandbox/env.rs
```diff
@@ -116,6 +116,57 @@ pub(crate) fn apply_environment_set_vars(
     Ok(())
 }
```
lukehinds, line 118: if poss trim down a bit , the projects loc count is getting a bit on the high side and you have really good coverage in the docs already!

lukehinds (commented):
Code is sound - would be good to get some docs in before we land, but as this opens up a lot of footguns, i think its worth narrowing down if what you care most about is  Interpreter/shell stuff.

The main concern (and fair dues, you do point it out and gate it): this could open up something huge if misused. The dangerous-var blocklist exists so an adversarial process can't push `LD_PRELOAD/NODE_OPTIONS` into a credential-bearing sandboxed child. `export_env` intentionally reopens that path, and the injected value comes from `request.env` (the intercepted command's parent env), not from the trusted profile. So a broad `session_export_env: ["*"]` (or exporting `LD_PRELOAD`) means any unmediated wrapper's environment can load arbitrary code into a child that then holds an injected credential socket - i.e. a ready-made exfiltration point.

So it's worth considering whether we should still hard-block a subset, even inside `export_env`. Looking at the blocklist, it's really three classes, not one:

- Actual secrets — `OP_SERVICE_ACCOUNT_TOKEN, OP_CONNECT_TOKEN_*.` These aren't tooling vars at all, they're the parent's1Password creds. Right now `["*"]` forwards them verbatim, bypassing the broker - which contradicts the "export_env is for tooling, not credentials" rule this PR sets. I'd make these always-excluded like `PATH/NONO_*`, never forwardable by any pattern.
- Universal loader injection ,  `LD_*/DYLD_*.` These hit any dynamically-linked child, not just an interpreter. Rather than block outright (sanitizers/faketime/etc. legitimately preload),I'd make `*` and `PREFIX_*` skip them, so forwarding one requires naming it exactly  - kills the accidental broad-`*` footgun while leaving a deliberate, auditable opt-out. Costs us nothing on the motivating use case, which is                   `NODE_OPTIONS/PYTHONPATH` anyway.
- Interpreter/shell-scoped - `NODE_OPTIONS, PYTHONPATH, BASH_ENV`, … leave these as-is; they're the actual reason for the feature and only bite the matching interpreter.

crates/nono-cli/src/tool-sandbox/env.rs
kipz: zomg, yes. done

kipz:
> Code is sound - would be good to get some docs in before we land, but as this opens up a lot of footguns, i think its worth narrowing down if what you care most about is Interpreter/shell stuff.
> 
> The main concern (and fair dues, you do point it out and gate it): this could open up something huge if misused. The dangerous-var blocklist exists so an adversarial process can't push `LD_PRELOAD/NODE_OPTIONS` into a credential-bearing sandboxed child. `export_env` intentionally reopens that path, and the injected value comes from `request.env` (the intercepted command's parent env), not from the trusted profile. So a broad `session_export_env: ["*"]` (or exporting `LD_PRELOAD`) means any unmediated wrapper's environment can load arbitrary code into a child that then holds an injected credential socket - i.e. a ready-made exfiltration point.
> 
> So it's worth considering whether we should still hard-block a subset, even inside `export_env`. Looking at the blocklist, it's really three classes, not one:
> 
> * Actual secrets — `OP_SERVICE_ACCOUNT_TOKEN, OP_CONNECT_TOKEN_*.` These aren't tooling vars at all, they're the parent's1Password creds. Right now `["*"]` forwards them verbatim, bypassing the broker - which contradicts the "export_env is for tooling, not credentials" rule this PR sets. I'd make these always-excluded like `PATH/NONO_*`, never forwardable by any pattern.
> * Universal loader injection ,  `LD_*/DYLD_*.` These hit any dynamically-linked child, not just an interpreter. Rather than block outright (sanitizers/faketime/etc. legitimately preload),I'd make `*` and `PREFIX_*` skip them, so forwarding one requires naming it exactly  - kills the accidental broad-`*` footgun while leaving a deliberate, auditable opt-out. Costs us nothing on the motivating use case, which is                   `NODE_OPTIONS/PYTHONPATH` anyway.
> * Interpreter/shell-scoped - `NODE_OPTIONS, PYTHONPATH, BASH_ENV`, … leave these as-is; they're the actual reason for the feature and only bite the matching interpreter.

@lukehinds took a pass at this...

kipz:
@lukehinds all three classes are now enforced, plus docs and the missing test coverage.

**Actual secrets — always excluded, not forwardable by any pattern.** `is_forbidden_secret_env_var` in `exec_strategy/env_sanitization.rs` covers `OP_SERVICE_ACCOUNT_TOKEN`, `OP_CONNECT_TOKEN`, `OP_CONNECT_HOST` and `OP_SESSION_*`, and `apply_export_env` checks it *after* pattern matching, so naming one exactly doesn't help. Test: `export_env_never_forwards_1password_secrets` asserts an empty result even for the list `["OP_SERVICE_ACCOUNT_TOKEN", "OP_SESSION_my", "*"]`.

**Loader injection — exact name only, as you suggested.** `is_loader_injection_env_var` matches `LD_*`/`DYLD_*`; those are skipped unless a pattern equals the variable name, so `*` and `LD_*` both miss them and a deliberate forward has to be spelled out and is therefore auditable in the profile. Tests: `export_env_star_does_not_forward_loader_vars` (list `["*", "LD_*", "DYLD_*"]` → nothing) and `export_env_exact_name_pattern_forwards_loader_var` (`["LD_PRELOAD"]` → forwarded).

**Interpreter/shell-scoped — unchanged.** `NODE_OPTIONS`, `PYTHONPATH`, `BASH_ENV` still forward under `*`, which is the motivating use case. `PATH` and `NONO_*` remain excluded under every pattern.

Docs are in: `docs/cli/features/tool-sandbox.mdx` and the authoring guide both carry the `export_env`/`session_export_env` rows and the exclusion rules, and the fields are in the published JSON schema with matching `schema_shape.rs` expectations (your earlier point about `additionalProperties: false`, and the repeated-trailing-wildcard rule, are both done).

Since your review I've also closed two coverage gaps a second review pass flagged:

- `caller_export_env` had no direct test on either platform — added on both, covering session → `session_export_env`, a named caller → its own `export_env`, and an **unknown** caller → empty. That last case matters: an unmediated wrapper must not inherit the session's exports by falling back.
- `export_env_cannot_displace_the_chaining_control_env` — naming `NONO_TOOL_SANDBOX_SOCKET` exactly must not beat the `NONO_*` exclusion, otherwise a child could point its own shim at a socket it controls. Previously only the bare-`*` case was covered.

On trimming: `env.rs` came down as you asked in the inline thread.

<sub>Posted by an agent on behalf of @kipz.</sub>

lukehinds (approved):
thanks , lgtm!

Issue #986, opened 2026-05-21
filesystem.read_file grants become stale when target is replaced via atomic rename (Linux/Landlock)


oalders:
> _Filed by Claude (Anthropic's Claude Code CLI agent) on behalf of @oalders, who confirmed the repro and reviewed the writeup before posting._

### What happened?

A `filesystem.read_file` grant in a profile stops working partway through a sandbox session if the granted file is replaced via the standard atomic write-tmp + `rename()` pattern that most config tools use (`git config`, editors saving with `:w`, dotfile installers, etc.). The session reads the file fine at first, then silently starts getting `EACCES` after the external rewrite — even though `nono why` (re-resolving the path) still reports `ALLOWED`.

This appears to be the standard Landlock semantic: `PathBeneath` rules are bound to the inode at ruleset load time, and `rename()` replaces the inode at that path. The kernel sees the new inode as ungranted; nono's profile evaluator, walking the path now, sees the grant.

### What did you expect to happen?

Either:
1. The grant follows the path (so atomic rewrites stay accessible), or
2. The behavior is documented and `nono why` reflects the live ruleset rather than the post-hoc path resolution, so users can diagnose without staring at inode numbers.

### Steps to reproduce

```bash
TMP=$(mktemp -d)
echo ORIGINAL > "$TMP/file"
cat > "$TMP/profile.json" <<JSON
{ "filesystem": { "read_file": ["$TMP/file"] } }
JSON

mkfifo "$TMP/in"
( cd "$TMP" && nono run --profile "$TMP/profile.json" --allow-cwd --silent -- \
    bash -s <"$TMP/in" ) &
exec 9>"$TMP/in"

echo "cat $TMP/file" >&9         # -> ORIGINAL
sleep 1
echo REPLACED > "$TMP/file.new"
mv "$TMP/file.new" "$TMP/file"   # atomic rename from OUTSIDE
echo "cat $TMP/file" >&9         # -> cat: ...: Permission denied
echo exit >&9
```

### Relevant output or logs

```shell
=== READ 1 ===
ORIGINAL
inode-from-inside-1: 9970910
=== READ 2 ===
cat: /tmp/nono-inode-repro/file: Permission denied
exit: 1
```

Meanwhile, from outside the sandbox:

```shell
$ nono why --path /tmp/nono-inode-repro/file --op read --profile /tmp/nono-inode-repro/profile.json
ALLOWED
  Reason: granted_path
  Granted by: /tmp/nono-inode-repro/file
[...]

PR #1448, opened 2026-07-20
fix(policy): survive atomic replacement of resolv.conf; report stale …

maufl:
## Linked Issue

References #986 — a partial fix by design (diagnostics plus one built-in policy instance, per the disclosure comment on the issue), so intentionally not `Closes`.

## Summary

Landlock rules bind to the inode open at ruleset build, so a file-level grant goes stale when the file is atomically replaced (write-temp + `rename(2)`). This PR:

1. **Policy** (`data/policy.json`): `system_read_linux_core` additionally grants the directory `/run/systemd/resolve` (read). Its `/etc/resolv.conf` file grant canonicalizes to `stub-resolv.conf`, which systemd-resolved replaces on any resolver-config change, breaking DNS for the rest of the session. Directory rules cover children created later, so they survive; non-systemd systems skip the missing path.
2. **Diagnostics**: the capability state file records `(dev, ino)` per file grant at sandbox start (Linux only; optional fields, compatible both ways). `nono why` re-stats the granted path and reports `DENIED / stale_file_grant` with grant-time vs. current inode and a restart hint — or ALLOWED with a warning when a directory grant still covers the path. Non-`--self` queries load the state leniently (unreadable state ⇒ no hints, never an error); `--self` stays strict.
3. **Docs**: troubleshooting entry + Landlock-internals note.

```
DENIED
  Reason: stale_file_grant
  Details: The sandbox granted this file, but the file at ... was replaced
  after sandbox start (inode 39:175234 when the sandbox started, 39:175245
  now). Landlock rules bind to the inode, not the path, ... Restart the
  sandbox to re-apply the grant to the current file.
```

**Deliberately not done:** Landlock has no path-based rules, so the only way a file grant can survive replacement is granting its parent directory. We did not apply that automatically to all file grants — it silently widens each to its whole directory, a posture change maintainers should make explicitly, per grant (as done here). The issue's repro still reproduces for user/profile file grants, now diagnosable. Other rename-prone builtin file grants (`/etc/ld.so.cache`, `/etc/hosts`, …) remain — their parent `/etc` is too broad. `/run/systemd/resolve` is systemd-specific in-tree policy; per NOGENT.md a nono package may be the longer-term home — guidance welcome.

## Agent Disclosure

- Agent-assisted contribution (Claude Fable 5 via Claude Code), directed and reviewed by @maufl; intent disclosed on #986 before opening this PR.
- Consulted: `AGENTS.md`, `NOGENT.md`, `sandbox/linux.rs` (PathFd inode binding), `capability.rs` (grant-time canonicalization), `policy.rs`, `sandbox_state.rs`, `why_runtime.rs`, `query_ext.rs`, the profile authoring guide, and this template. All code newly written; nothing reused or adapted.
- Confirmed compliant with repository coding and security requirements: NonoError propagation, no unwrap/expect in production code, paths canonicalized, fail-closed behavior preserved (the lenient loader affects diagnostics only, never enforcement).

## Test Plan

- New unit tests: staleness detection (fresh / in-place / replaced / removed), both `why` verdicts, pass-through, inode recording, legacy-state compatibility, lenient-loader failure modes.
- `make check`, `make test`, `cargo audit`, `make lint-aliases`, `make lint-docs` pass locally (remaining failures are pre-existing PTY/AF_UNIX tests needing permissions the sandboxed dev env lacks, all in untouched files).
- Verified end-to-end on Arch/zen (Landlock ABI v6): resolv.conf reads survive stub replacement under the new grant; the output above is from a live session; non-`--self` `why` still answers with an unreadable state file.

## Checklist

- [x] Release note has been added to `CHANGELOG.md` if needed (generated via git-cliff from conventional commits)

- [x] An issue already exists

### DIFF
--- a/crates/nono-cli/data/policy.json
+++ b/crates/nono-cli/data/policy.json
@@ -211,6 +211,7 @@
           "/usr/local/lib",
           "/usr/local/lib64",
           "/etc/resolv.conf",
+          "/run/systemd/resolve",
           "/etc/hosts",
           "/etc/nsswitch.conf",
           "/etc/gai.conf",
--- a/crates/nono-cli/src/query_ext.rs
+++ b/crates/nono-cli/src/query_ext.rs
@@ -55,6 +55,10 @@ pub enum QueryResult {
         source: Option<String>,
         #[serde(skip_serializing_if = "Option::is_none")]
         endpoint_rules: Option<Vec<crate::sandbox_state::EndpointRuleState>>,
+        /// Advisory caveat: policy allows the operation, but enforcement may
+        /// still deny it (e.g. a sibling file grant went stale).
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        warning: Option<String>,
     },
     /// The operation is denied
     #[serde(rename = "denied")]
@@ -181,6 +185,7 @@ pub fn query_path(
             access: Some(cap.access.to_string()),
             source: Some(cap.source.to_string()),
             endpoint_rules: None,
+            warning: None,
         });
     }
 
@@ -271,6 +276,7 @@ pub fn query_network(
                                     )),
                                     source: Some("domain allowlist".to_string()),
                                     endpoint_rules: Some(de.endpoints.clone()),
+                                    warning: None,
                                 }
                             } else {
                                 QueryResult::Denied {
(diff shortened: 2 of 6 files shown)


lukehinds:
Thanks @maufl , I am good with the /run/systemd/resolve grant, there is nothing private there , the alternative (broken DNS on every resolver-config change) is much worse. approved

lukehinds (approved):

PR #1467, opened 2026-07-20
fix(tool-sandbox): grant command interpreter read of its script

### LINKED ISSUE #1466: Sandboxed #!/usr/bin/env <interp> script can't read itself under the tool-sandbox (EACCES)
## What happened?

On Linux, when the tool-sandbox is engaged (the profile defines `command_policies`), running a `#!/usr/bin/env node` script under the sandbox fails: the interpreter gets `EACCES` reading its own script, **even when the script's directory is granted**. A bare profile with no command policies does not reproduce it, and a directly-exec'd binary reading the same file succeeds.

The interpreter is launched to run the script (`argv = ["node", ".../hello.js"]`) inside a per-exec child sandbox that grants the interpreter (`/usr/bin/node`) and its runtime baseline, but not the **script** — which is an argv path, not the launched binary — nor the script's module dependency tree. The session/outer grant for the script's directory does not reach that child sandbox.

## What did you expect to happen?

The interpreter should be able to read the script it is launched to run (and its co-located files) when the caller's own sandbox already permits reading them — as a directly-exec'd binary can.

## Steps to reproduce

1. Profile `repro.json` that engages the tool-sandbox (any `command_policies.commands` entry) and grants a dir:
   ```json
   { "meta": { "name": "repro" }, "extends": ["default"],
     "filesystem": { "allow": ["~/scripts"] },
     "command_policies": { "commands": { "git": { "sandbox": {} } } } }
   ```
2. `~/scripts/hello.js`:
   ```js
   #!/usr/bin/env node
   console.log("hi")
   ```
   (`chmod +x ~/scripts/hello.js`)
3. Run: `nono run --profile repro --allow ~/scripts -- ~/scripts/hello.js`
4. Fails with `EACCES` opening `hello.js`, though `~/scripts` is granted.

## Relevant output or logs

```shell
Error: EACCES: permission denied, open '/home/user/scripts/hello.js'
# nono reports "No path denials were observed" — the interpreter runs in a
# per-exec child sandbox narrower than the session, which the fs grant never reaches.
```

## nono version

0.69.0

## Operating system

Linux

## OS version / distro

Ubuntu 22.04

## Kernel version (Linux only)

6.8.0-1057

## Profile (if relevant)

```json
{
  "meta": { "name": "repro" },
  "extends": ["default"],
  "filesystem": { "allow": ["~/scripts"] },
  "command_policies": { "commands": { "git": { "sandbox": {} } } }
}
```

kipz:
## Linked Issue

Closes #1466

## Summary

On Linux, when the tool-sandbox is engaged, a `#!/usr/bin/env <interp>` script's interpreter is launched to run the script inside a per-exec child sandbox that grants the interpreter + its runtime baseline but not the **script** — which is an argv path, not the launched binary — so the interpreter gets `EACCES` reading its own script (and its module tree), even when the caller's session already grants the script's directory. This grants the child read of argv paths the agent's **outer** sandbox already permits, re-using the covering outer read grant so dependency trees resolve; bounded by `outer_caps`, so the child gains nothing the caller lacked. When such a script is matched, it also grants the session TLS-intercept trust bundle (otherwise given only to commands with their own proxy route), so the interpreter can make the caller's HTTPS calls.

- This PR was authored by an AI agent (Claude) on behalf of the repository owner.
- Areas changed: `crates/nono-cli/src/tool-sandbox/platform/linux.rs` (`build_child_caps` + new `add_interpreted_script_read`/`_inner`); reviewed `CLAUDE.md` coding/security rules and the sibling handling from #1394.
- The grant is bounded by `outer_caps` (`path_covered_with_access`) so it never widens the child beyond the caller's own read set; argv paths are canonicalized before use.

## Test Plan

- New unit test `interpreted_script_read_grants_covered_script_and_trust_bundle` (Linux): asserts a covered script's enclosing outer read grant + the trust bundle are granted, and that an argv path **outside** `outer_caps` is not granted. Passes (`cargo test -p nono-cli`, Linux container).
- Built `nono` for Linux and verified manually: a `#!/usr/bin/env node` script run under a profile with command policies now executes (was `EACCES` on the script before); reads outside the caller's grants remain denied.
- Change is `#[cfg(target_os = "linux")]`; host `make ci` clippy/fmt/tests pass (one unrelated attestation-signing test is a known parallel-run flake — passes in isolation).

## Checklist

- [x] An issue already exists

### DIFF
--- a/crates/nono-cli/src/command_policy.rs
+++ b/crates/nono-cli/src/command_policy.rs
@@ -2905,7 +2905,10 @@ fn read_command_prefix(path: &Path, max_bytes: usize) -> nono::Result<Vec<u8>> {
 }
 
 #[cfg(any(test, target_os = "linux", target_os = "macos"))]
-fn classify_executable_shape(path: &Path, bytes: &[u8]) -> nono::Result<ResolvedExecutableShape> {
+pub(crate) fn classify_executable_shape(
+    path: &Path,
+    bytes: &[u8],
+) -> nono::Result<ResolvedExecutableShape> {
     if bytes.starts_with(b"\x7fELF") {
         return Ok(ResolvedExecutableShape {
             kind: ResolvedExecutableKind::Elf,
--- a/crates/nono-cli/src/tool-sandbox/platform/linux.rs
+++ b/crates/nono-cli/src/tool-sandbox/platform/linux.rs
@@ -4,7 +4,8 @@ use crate::audit_integrity::{
 };
 use crate::command_policy::{
     CommandFromConfig, CommandPoliciesConfig, CommandSandboxConfig, ResolvedCommandBinaries,
-    ResolvedCommandBinary, ResolvedExecutableKind, has_explicit_self_invocation_entry,
+    ResolvedCommandBinary, ResolvedExecutableKind, classify_executable_shape,
+    has_explicit_self_invocation_entry,
 };
 use crate::lineage_cgroup::LineageMarker;
 use crate::profile;
@@ -3451,6 +3452,7 @@ fn build_child_caps(
     )?);
     add_runtime_baseline(&mut caps, &state.baseline_cache, &binary.canonical_path)?;
     add_executable_shape_baseline(&mut caps, state, binary)?;
+    add_interpreted_script_read(&mut caps, state, binary, request)?;
     add_chaining_control_caps(&mut caps, state)?;
     add_policy_fs(
         &mut caps,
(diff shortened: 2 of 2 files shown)


kipz:
Thanks — addressed the findings:

- **[HIGH · security] CA/grant leaked to any command via argv file path.** Real concern, fixed. Note the suggested `binary.shape.kind == ShebangScript` gate wouldn't work here: the launched binary is the **interpreter** (`/usr/bin/node`, `kind == Elf`) with the script as an argv path, so that gate would disable the fix entirely. Instead the grant now fires only when an argv path **is a shebang script whose resolved interpreter is the launched binary** (`classify_executable_shape` + `env_shebang_target_interpreter`, basename match). `curl /path/file` (plain file) and `curl script.sh` (shebang for a different interpreter) both match nothing, so they get neither the read grant nor the trust bundle. Covered by new negative test cases.

- **[MEDIUM · bug] `path_covered_with_access` precheck ignores file capabilities.** Removed the precheck entirely; coverage is now decided in the per-capability loop, which handles both file and directory caps.

- **[MEDIUM · design] called for all commands.** The call stays in `build_child_caps`, but the shebang-interpreter check above scopes it precisely — ordinary (non-interpreter) commands match nothing, so no argv read grant or trust bundle is added for them.

- **[LOW · bug] macOS divergence.** This is Linux/Landlock-specific — under macOS Seatbelt the interpreter reads its script without this failure in our testing, so no macOS change is included here. Happy to add the equivalent if a macOS repro surfaces.

kipz:
Fixed both — the coverage check now runs before the file is opened/classified (`is_shebang_script_for` is only called once a covering `outer_caps` grant is found), and relative argv paths are now resolved against `request.cwd` before canonicalizing instead of being skipped. Added regression tests for both (`interpreted_script_read_grants_nothing_for_an_uncovered_script`, `interpreted_script_read_resolves_relative_argv_against_cwd`).

lukehinds (approved):

PR #1476, opened 2026-07-21
feat: mediate vault login -method=oidc (custom inject header + per-command open_port)

### LINKED ISSUE #1473: Support brokering credentials for CLIs that use a non-Bearer auth header and a localhost OIDC callback

I want to sandbox a CLI (a coding agent / build tool) that runs HashiCorp Vault's `vault login -method=oidc`, so the real Vault token never enters the sandbox. `oauth_capture` is the right mechanism — capture the token at the OIDC callback, hand the sandbox a phantom, redeem it on egress — but three things about the Vault CLI flow don't fit today:

1. **Vault sends the token in `X-Vault-Token`, not `Authorization: Bearer`.** The `api_hosts` redemption route only injects `Authorization: Bearer {}`, so Vault rejects the request.
2. **The Vault OIDC callback endpoint is capture-only.** After the browser round-trip, `vault` GETs `/v1/auth/<mount>/oidc/callback` and the token arrives in the *response*; it is never re-sent in a request body. But `oauth_capture` requires `request_nonce_fields` on every token endpoint, so this endpoint can't be declared.
3. **`vault login` binds a localhost callback listener (`localhost:8250`) that a mediated command can't bind.** If `vault` runs as a `command_policies` child, `network.tcp_bind_ports` is rejected for tool-sandbox children on macOS, and there is no per-command `open_port` (the top-level policy has one). The bind fails with `listen: operation not permitted` and the login can't complete.

Support these so `vault login -method=oidc` can be mediated:

1. Optional `inject_header` / `credential_format` on `oauth_capture` providers (default `Authorization` / `Bearer {}`) — set `X-Vault-Token` / `{}` for Vault.
2. Optional `request_nonce_fields` for capture-only endpoints like the Vault OIDC callback. It only marks where to swap a phantom by value in an outbound refresh body; it does not gate capture, and `api_hosts` is mandatory, so omitting it is safe.
3. `open_port` / `open_port_range` on per-command network policies (mirroring the top-level field) so the mediated `vault` command can bind its `localhost:8250` callback (enforceable on macOS, unlike `tcp_bind_ports`).

With these, a `vault_oidc` provider captures `auth.client_token` at the callback and the sandboxed `vault` only ever sees a phantom, redeemed to the real token on egress to the Vault API.

Raw seatbelt bind rules for the callback (ignored for tool-sandbox children); `network.allow_all` on `vault` (allows the bind but drops the proxy trust-bundle grant, breaking TLS interception); `tcp_bind_ports` (rejected on macOS).

## How is this blocking you?

It's blocking production use of nono — `vault login -method=oidc` can't be mediated at all, so a Vault-dependent workflow either leaks the token into the sandbox or can't run.

## Additional context

Prototyped and proven end-to-end against a real `vault login -method=oidc`: token captured at the callback, phantom-only inside the sandbox, redeemed on egress. Draft PR ready.

kipz:
## Linked Issue

Closes #1473

## Summary

Enables mediating HashiCorp Vault's `vault login -method=oidc` so the real Vault token never enters the sandbox — captured at the OIDC callback, replaced with a phantom, redeemed on egress to the Vault API. Three generic `oauth_capture`/policy gaps blocked the Vault CLI flow:

- **Optional `inject_header` / `credential_format` on `oauth_capture` providers** — Vault authenticates with `X-Vault-Token: <raw token>`, not `Authorization: Bearer {}`. Defaults unchanged; set `X-Vault-Token` / `{}` for Vault. Validated at parse time: `credential_format` must contain `{}`; `inject_header` must be a valid HTTP header token.
- **`request_nonce_fields` now optional** — the Vault OIDC callback (`/v1/auth/<mount>/oidc/callback`) is capture-only: the token arrives in the response and is never re-sent in a request body. The field only marks where to swap a phantom by value in an outbound refresh body — it does not gate capture, and `api_hosts` is mandatory, so omitting it is safe.
- **`open_port` / `open_port_range` on per-command network policies** — so the mediated `vault` command can bind its `localhost:8250` OIDC callback listener. Mirrors the top-level `network.open_port`, wired through `add_policy_network` (macOS + Linux) via the existing `localhost_port_ranges` path (enforceable on macOS, unlike `tcp_bind_ports`).

## Agent Disclosure

This PR was authored by an AI agent (disclosed on #1473).

- Consulted: `credential_provider.rs`, `proxy_runtime.rs`, `command_policy.rs`, `tool-sandbox/platform/{macos,linux}.rs`, `oauth_capture/{rewrite,endpoint}.rs`, and `CLAUDE.md`.
- Composes with existing mechanisms (`add_localhost_port_range` -> `proxy_bind_port_ranges`; the `api_hosts` route builder) rather than adding new machinery.
- No `unwrap`/`expect` in new code; uses `NonoError`; adds no new filesystem paths.

## Test Plan

- `make ci` clean (fmt, clippy -D warnings, tests) on a fresh rebase onto `upstream/main`.
- New unit tests: custom `inject_header` on the api_hosts route; `open_port` merge + parse round-trip; `open_port` -> localhost bind ranges (macOS + Linux); `credential_format` / `inject_header` validation.
- Verified end-to-end against a real `vault login -method=oidc`: `auth.client_token` captured at the callback and replaced with a phantom, the sandboxed `vault` only ever holds the phantom, and it is redeemed to the real token on egress to the Vault API.
- Docs updated: `sandboxed-oauth-logins.mdx` (new fields + Vault callback-bind note) and `tool-sandbox.mdx` (`open_port` reference).

## Checklist

## Agent Compliance Check

- [x] An issue already exists

### DIFF
--- a/crates/nono-cli/src/profile/credential_provider.rs
+++ b/crates/nono-cli/src/profile/credential_provider.rs
@@ -198,6 +198,17 @@ pub(super) fn validate_credential_provider_entries(profile: &Profile) -> Result<
                 "credential_providers.{name}.credential_format must contain the '{{}}' token placeholder"
             )));
         }
+        // Same raw header-value path as the header name; reject control chars
+        // to prevent header injection.
+        if provider
+            .credential_format
+            .as_deref()
+            .is_some_and(|format| format.bytes().any(|b| b.is_ascii_control() && b != b'\t'))
+        {
+            return Err(NonoError::ProfileParse(format!(
+                "credential_providers.{name}.credential_format must not contain control characters"
+            )));
+        }
         if provider
             .inject_header
             .as_deref()
@@ -483,6 +494,30 @@ mod tests {
             err.to_string().contains("credential_format"),
             "unexpected error: {err}"
         );
+    }
+
+    #[test]
+    fn credential_format_with_control_chars_is_rejected() {
+        for bad in [
+            "{}\r\nEvil: 1",
+            "Bearer {}\n",
+            "{}\0",
+            "{}\x01",
+            "{}\x0b",
+            "{}\x0c",
+            "{}\x7f",
+        ] {
+            let profile = provider_with(Some("X-Vault-Token"), Some(bad));
+            let err = validate_credential_provider_entries(&profile)
+                .expect_err(&format!("must reject format {bad:?}"));
+            assert!(
+                err.to_string().contains("credential_format"),
+                "unexpected error for {bad:?}: {err}"
+            );
+        }
+        // Horizontal tab is a legal field-value character.
+        let ok = provider_with(Some("X-Vault-Token"), Some("Bearer\t{}"));
+        validate_credential_provider_entries(&ok).expect("tab is allowed");
     }
 
     #[test]
--- a/crates/nono-cli/src/profile/credential_provider.rs
+++ b/crates/nono-cli/src/profile/credential_provider.rs
@@ -15,6 +15,15 @@ pub struct CredentialProviderDef {
     /// API origins where phantom tokens are resolved on egress.
     #[serde(default)]
     pub api_hosts: Vec<String>,
+    /// Header the resolved credential is injected into on `api_hosts` egress.
+    /// Defaults to `Authorization` (OAuth Bearer APIs). Vault, for example,
+    /// needs `X-Vault-Token`.
+    #[serde(default)]
+    pub inject_header: Option<String>,
+    /// Format applied to the resolved credential before injection; `{}` is the
+    /// token. Defaults to `Bearer {}`. Use `{}` for a raw token (e.g. Vault).
+    #[serde(default)]
+    pub credential_format: Option<String>,
     /// Optional provider-specific logout/session detection.
     #[serde(default)]
     pub credential_store: Option<CredentialProviderStore>,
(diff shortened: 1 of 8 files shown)


crates/nono-cli/src/profile/credential_provider.rs
```diff
@@ -178,6 +186,27 @@ pub(super) fn validate_credential_provider_entries(profile: &Profile) -> Result<
...
+        }
+        if provider
+            .inject_header
+            .as_deref()
+            .is_some_and(|header| !is_valid_http_header_name(header))
+        {
+            return Err(NonoError::ProfileParse(format!(
+                "credential_providers.{name}.inject_header must be a valid HTTP header name (RFC 7230 token)"
+            )));
+        }
```
lukehinds, line 220: looks good @kipz , just one small extra worth getting in. `credential_format`'s value template isn't CRLF-checked but the header name is, and it flows through the same raw-string path. It's operator-supplied/trusted, so it's a consistency/defense-in-depth nit, not a blocker 

perhaps something like this on line 200 of your last commit (above the check for .inject_header):

```rust
if provider
    .credential_format
    .as_deref()
    .is_some_and(|format| format.bytes().any(|b| matches!(b, b'\r' | b'\n' | b'\0')))
{
    return Err(NonoError::ProfileParse(format!(
        "credential_providers.{name}.credential_format must not contain control characters (CR, LF, NUL)"
    )));
}
```

kipz: @lukehinds done in 2c84b730. I widened the guard while there: rejects all HTTP field-value control bytes except horizontal tab (`b.is_ascii_control() && b != b'\t'`), not just CR/LF/NUL. added tests plus a positive case that tab is allowed.

SequeI (approved):
lgtm! thanks for the fix too

PR #1488, opened 2026-07-22
feat: implement nono profile init --draft

### LINKED ISSUE #1479: Create draft profiles from the CLI

Nono's CLI has the ability to validate and promote draft profiles, but not create them. This is problematic because it would require tools to hardcode the `draft-profiles` path, and more importantly, for existing profiles, you also need to generate the `.base` file when creating a draft

* `nono profile init my-agent --draft` checks to see if `my-agent` exists. If so, it copies the existing profile as-is into the drafts folder, otherwise it generates a new one (same as before, just in the drafts folder)
* `nono profile init my-agent --draft --source=template my-agent` errors if `my-agent` already exists. Otherwise, it copies over the existing my-agent, and also writes the .base file
* `nono profile init my-agent --draft --source=existing my-agent` uses the existing profile as the starting point, and errors out if it doesn't exist. This is the inverse of --source=template

Using a shell script to generate the .base value:
shasum -a 256 "$HOME/.config/nono/profiles/${PROFILE_NAME}.json" | awk '{print $1}' \
  > "$HOME/.config/nono/profile-drafts/${PROFILE_NAME}.base"

### How is this blocking you?

### Additional context

_No response_

intentionally-left-nil:
This extends the functionality of profile init to support draft profiles If the target profile already exists, it is used as a starting point, otherwise the existing template is used.

The behavior of init with drafts can be further controlled with --source to force the usage of a template or the existing profile.

init --draft also writes the .base file needed for nono promote to work later

## Linked Issue

Closes #1479

## Summary
This adds --draft and --source flags to nono profile init, and wires up the new behavior according to the issue.
Some flags are incompatible with --draft mode (namely --output and --full) and those are checked. Next, draft mode determines if the profile is new, or editing an existing one. The correct source profile is used, modified according to the remaining CLI arguments, and then written to .draft-profiles/profile-name

AI was used to generate the tests, and assisted with code generation under supervision.

## Test Plan

TBD

## Checklist


crates/nono-cli/src/profile_cmd.rs
```diff
@@ -197,6 +250,94 @@ fn profile_validate_target(args: &ProfileInitArgs, output_path: &Path) -> String
...
+            .entry("groups".to_string())
+            .or_insert_with(|| serde_json::Value::Object(serde_json::Map::new()));
+        if let serde_json::Value::Object(groups_map) = groups {
+            groups_map.insert(
+                "include".to_string(),
+                serde_json::Value::Array(
+                    args.groups
+                        .iter()
+                        .cloned()
+                        .map(serde_json::Value::String)
```
SequeI, line 329: `groups_map.insert` replaces `groups.include` entierly, if the existing profile already has groups, `--groups c` will drop them and write `["c"]`. Is that intentional? If so, worth a note in the flag description; if not, this should merge rather than overwrite.

crates/nono-cli/src/profile_cmd.rs
```diff
@@ -197,6 +250,94 @@ fn profile_validate_target(args: &ProfileInitArgs, output_path: &Path) -> String
     }
 }
 
+fn resolve_init_source(args: &ProfileInitArgs) -> Result<ProfileInitSource> {
+    if !args.draft {
+        return Ok(ProfileInitSource::Template);
+    }
+    let target_path = profile::get_user_profile_path(&args.name)?;
+    let target_exists = regular_file_exists(&target_path, "target profile")?;
+    match (args.source, target_exists) {
```
SequeI, line 259: When `!args.draft`, this function resolves the target path and checks if it exists, then immediately returns `Template` without using either result. The early return should come before the `get_user_profile_path` call.

SequeI:
No tests for `--draft --force` when a draft already exists. `--force` is not rejected in `--draft` mode, so the existing-file check in `cmd_init` should let it through, but it is an untested path. We should add tests for this mode aswell just to be sure

left a few nit comments, otherwise looks good and once those are addressed, we can look at getting this in!

PR #1508, opened 2026-07-27
fix(registry): suppress X-Nono-UUID when update check is opted out

### LINKED ISSUE #1507: Registry requests still send X-Nono-UUID after update-check opt-out
## Summary

Registry requests to `registry.nono.sh` continue to send the persistent installation identifier `X-Nono-UUID` **even after the user has opted out of the update check**, whenever the update-check state file already exists on disk (which is the default after the very first `nono` run).

This makes the documented opt-out (`NONO_NO_UPDATE_CHECK` / `[updates] check = false`) inaccurate: it stops the `update.nono.sh` ping but does **not** stop the per-install identifier being transmitted to the registry on every `pull` / `search` / `status` / pack-hint refresh.

## Where it happens

`build_context_headers()` in `crates/nono-cli/src/registry_client.rs` reads the UUID unconditionally:

```rust
if let Some(uuid) = crate::update_check::read_installation_uuid() {
    headers.push(("X-Nono-UUID", uuid));
}
```

`read_installation_uuid()` (`crates/nono-cli/src/update_check.rs`) returns the stored UUID whenever the state file exists. Neither path consults the opt-out predicate used by `start_background_check()` (env `NONO_NO_UPDATE_CHECK`, or config `[updates] check = false`).

Because the very first non-silent `nono` run generates and writes `~/.local/state/nono/update-check.json` before the user has any chance to opt out, the residual-UUID transmission is the default rather than an edge case.

## Expected behaviour

When the user has opted out of the update check (via `NONO_NO_UPDATE_CHECK` or `[updates] check = false`), nono must not transmit the persistent installation identifier to the registry, regardless of whether a state file already exists.

## Proposed approach

1. Add a single authoritative opt-out predicate `update_check_opted_out()` in `update_check.rs` (env var presence + config `check = false`), and route the existing `start_background_check()` checks through it.
2. Gate `read_installation_uuid()` on that predicate — return `None` when opted out. Since its only caller is the registry header builder, this fail-safe suppresses the UUID everywhere it could be transmitted, while keeping its existing no-write contract.
3. Reuse the same predicate in `pack_update_hint::is_opted_out()` to remove the duplicated env/config logic.

Coarse fingerprinting headers (`X-Nono-Platform`, `X-Nono-Arch`, `X-Nono-CI`, `X-Nono-Install-Source`) are intentionally **out of scope** here — they remain gated by the separately-tracked "master telemetry switch" work. This change is limited to the persistent identifier.

## Risks / tradeoffs

- Behavioural change only for opted-out installs; opted-in behaviour is unchanged.
- The state file is **not** auto-deleted (that would be surprising destructive behaviour); users who want the on-disk UUID removed still delete the file manually. This change stops transmission, which is the privacy-relevant boundary.

## Verification

- New unit test asserting `read_installation_uuid()` returns `None` when `NONO_NO_UPDATE_CHECK` is set even though a valid state file exists, and returns the UUID when not opted out.
- `make ci` (clippy + fmt + tests).

---
_Filed by an AI-assisted contributor at the request of the maintainer, ahead of any code change, per the repository Coding Agent Contribution Policy._

scp7:
## Summary

Fixes #1507.

Registry requests to `registry.nono.sh` continued to send the persistent installation identifier `X-Nono-UUID` **even after the user opted out** of the update check, whenever the update-check state file already existed on disk (the default after the first `nono` run). The opt-out (`NONO_NO_UPDATE_CHECK` / `[updates] check = false`) gated only the `update.nono.sh` ping, not the registry header, so the documented opt-out was inaccurate in the common case.

## Approach

- Added a single authoritative opt-out predicate `update_check_opted_out()` in `crates/nono-cli/src/update_check.rs` (env `NONO_NO_UPDATE_CHECK` presence — any value — or config `[updates] check = false`).
- Routed `start_background_check()` through it (behaviour-preserving).
- **Gated `read_installation_uuid()` on the predicate** — it now returns `None` when opted out, suppressing `X-Nono-UUID` at the source for every consumer, even if a state file was written on an earlier run. The state file is intentionally **not** deleted (transmission stops; users still remove the file to erase the on-disk UUID).
- De-duplicated `pack_update_hint::is_opted_out()` to reuse the shared predicate.
- Corrected the header doc comment in `registry_client.rs`.

The coarse fingerprinting headers (`X-Nono-Platform`/`Arch`/`CI`/`Install-Source`) are intentionally **out of scope** — they remain gated by the separately-tracked master telemetry switch. This change is limited to the persistent identifier.

## Tests

- `update_check::test_read_installation_uuid_respects_opt_out` — reader returns the UUID when opted in; `None` under both the env-var and config opt-outs; state file left on disk.
- `registry_client::context_headers_gate_uuid_on_opt_out` — `X-Nono-UUID` present when opted in, absent when opted out, while `X-Nono-Platform`/`X-Nono-Install-Source` remain (pins the scope boundary).

Both tests are hermetic (serialized on `ENV_LOCK`, all env vars guard-managed, isolated temp `HOME`/`XDG_STATE_HOME`/`XDG_CONFIG_HOME`). `make ci` is clean locally.

## Files consulted

`crates/nono-cli/src/update_check.rs`, `crates/nono-cli/src/registry_client.rs`, `crates/nono-cli/src/pack_update_hint.rs`, `crates/nono-cli/src/config/user.rs`, `crates/nono-cli/src/state_paths.rs`, `crates/nono-cli/src/config/mod.rs`, `crates/nono-cli/src/test_env.rs`.

## Contributor disclosure

This PR was prepared by an AI-assisted contributor at the request of the maintainer, following the repository Coding Agent Contribution Policy (issue filed and disclosed before any code change).

## Agent Compliance Check

- [x] An issue already exists (#1507)
- [x] I did not use forbidden `unwrap`/`expect` in production code (test code uses `.expect()` consistent with the repo's existing test convention; `clippy::unwrap_used` passes under `make ci`)

### DIFF
--- a/crates/nono-cli/src/pack_update_hint.rs
+++ b/crates/nono-cli/src/pack_update_hint.rs
@@ -299,13 +299,7 @@ fn is_opted_out() -> bool {
     if std::env::var(NO_PACK_UPDATE_HINTS_ENV).is_ok() {
         return true;
     }
-    if std::env::var("NONO_NO_UPDATE_CHECK").is_ok() {
-        return true;
-    }
-    match crate::config::user::load_user_config() {
-        Ok(Some(config)) => !config.updates.check,
-        _ => false,
-    }
+    crate::update_check::update_check_opted_out()
 }
 
 fn is_newer(installed: &str, latest: &str) -> bool {
--- a/crates/nono-cli/src/registry_client.rs
+++ b/crates/nono-cli/src/registry_client.rs
@@ -29,7 +29,9 @@ pub struct RegistryClient {
 /// Build the installation-context headers to attach to every registry request.
 ///
 /// Headers included:
-/// - `X-Nono-UUID`: installation UUID if a state file exists (omitted otherwise)
+/// - `X-Nono-UUID`: installation UUID (omitted when the update check is opted
+///   out via `NONO_NO_UPDATE_CHECK` or `[updates] check = false`, or when no
+///   state file exists)
 /// - `X-Nono-Platform`: OS name (`std::env::consts::OS`)
 /// - `X-Nono-Arch`: CPU architecture (`std::env::consts::ARCH`)
 /// - `X-Nono-CI`: `"true"` when a recognised CI environment is detected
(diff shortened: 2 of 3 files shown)


SequeI (approved):
lgtm!

PR #1521, opened 2026-07-28
fix: allow non-UTF-8 command line arguments

### LINKED ISSUE #1504: nono panics on any non-UTF-8 command-line argument (std::env::args unwrap)
### What happened?

Any `nono` invocation that includes an argument which is not valid UTF-8 aborts with a
Rust panic before doing anything, including before printing usage. It exits with 134
(SIGABRT).

```console
$ nono $'\xff'

thread 'main' panicked at library/std/src/env.rs:876:51:
called `Result::unwrap()` on an `Err` value: "\xFF"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
$ echo $?
134
```

This is not specific to a subcommand — `nono audit list $'\xff'` and
`nono run --allow-cwd -- echo $'\xff'` panic identically, so it is top-level argument
handling rather than anything in a particular command.

On macOS and Linux arguments are byte strings, so this is reachable in ordinary use —
for example passing a filename that is not valid UTF-8, or any wrapper script that
forwards argv it did not create.

A normal error, or for the argument to be accepted and handled. Concretely, for
`nono run -- echo $'\xff'` I would expect the sandbox to start and `echo` to receive the
argument, since the argument is meant for the sandboxed program and not for nono.

Failing that, a clean diagnostic and a non-abort exit code would be a big improvement
over a panic.

### Steps to reproduce

1. `nono $'\xff'` (zsh or bash `$'...'` syntax produces the raw `0xFF` byte)
2. Observe the panic and exit code 134.

Also reproduces as:

- `nono audit list $'\xff'`
- `nono run --allow-cwd -- echo $'\xff'`

Control: `nono --version` works normally.

### Relevant output or logs

```shell
$ RUST_BACKTRACE=1 nono $'\xff'

thread 'main' panicked at library/std/src/env.rs:876:51:
called `Result::unwrap()` on an `Err` value: "\xFF"
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

### nono version

0.69.0

### Operating system

macOS

### OS version / distro

macOS 26.5.2 (25F84), Mac16,11 (M4 Pro)

### Additional context

`std::env::args()` panics on any argument that is not valid UTF-8, unlike
`std::env::args_os()`. The reachable call looks like `collect_legacy_network_warnings()`
in `crates/nono-cli/src/cli_bootstrap.rs`, which is invoked very early from `main`:

```rust
// crates/nono-cli/src/cli_bootstrap.rs:22
let args: Vec<String> = std::env::args().skip(1).collect();
```

That matches the observed behaviour — it runs before clap, which is why even an unknown
subcommand panics rather than producing a usage error.

Interestingly the adjacent deprecated-flag scan already does the right thing, on the very
next line of `main`:

```rust
// crates/nono-cli/src/main.rs:125
let os_args: Vec<_> = std::env::args_os().collect();
```

Since `collect_legacy_network_warnings` only compares against ASCII legacy flag names, a
lossy conversion is sufficient and preserves the existing behaviour for all valid input:

```rust
let args: Vec<String> = std::env::args_os()
    .skip(1)
    .map(|a| a.to_string_lossy().into_owned())
    .collect();
```

Installed via Homebrew. Happy to open a DCO-signed PR if that would help.

kurtisc:
## Linked Issue

Closes #1504

## Summary

Read CLI arguments as `OsString` so they can be passed to child programs directly.

## Test Plan
Steps to reproduce fixed

Tests added

## Checklist

### DIFF
--- a/crates/nono-cli/src/cli.rs
+++ b/crates/nono-cli/src/cli.rs
@@ -5,6 +5,7 @@
 
 use clap::builder::styling::{Style, Styles};
 use clap::{Parser, Subcommand, ValueEnum};
+use std::ffi::OsString;
 use std::path::PathBuf;
 
 const STYLES: Styles = Styles::plain().header(Style::new().bold());
@@ -2102,7 +2103,7 @@ pub struct RunArgs {
 
     /// Command to run inside the sandbox (optional if profile specifies `binary`)
     #[arg(hide = true)]
-    pub command: Vec<String>,
+    pub command: Vec<OsString>,
 
     /// Print help
     #[arg(long, short = 'h', action = clap::ArgAction::Help, help_heading = "OPTIONS")]
@@ -2150,7 +2151,7 @@ pub struct WrapArgs {
 
     /// Command to run inside the sandbox
     #[arg(required = true, hide = true)]
-    pub command: Vec<String>,
+    pub command: Vec<OsString>,
 
     /// Print help
     #[arg(long, short = 'h', action = clap::ArgAction::Help, help_heading = "OPTIONS")]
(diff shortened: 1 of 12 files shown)


connrg (approved):
lgtm

Issue #1532, opened 2026-07-30
env_credentials: two entries mapping different keystore accounts to the same env var race non-deterministically (no override, no warning)


dselivanov:
### Summary

When two `env_credentials` entries — typically a base profile and a child that `extends` it — map **different keystore accounts to the same destination env var**, both entries survive the merge and the final value of that env var is chosen in **random order per launch**. There is no child-override and no collision warning, so the symptom is a silent, intermittent "wrong token" failure.

This contrasts with the sibling field `environment.set_vars`, which merges **child-wins-by-destination**.

### Reproduce

`base.json`:

```json
{ "env_credentials": { "ACCOUNT_A": "MY_TOKEN" } }
```

`child.json` (extends base, maps a *different* account to the *same* env var):

```json
{ "extends": ["base"], "env_credentials": { "ACCOUNT_B": "MY_TOKEN" } }
```

With both `ACCOUNT_A` and `ACCOUNT_B` present in the keystore (service `nono`), `MY_TOKEN` flips between the two secret values across launches — no error, no warning.

Real-world case that bit us: a shared base mapped `GITHUB_TOKEN→GITHUB_TOKEN`; a GHE child mapped `GH_TRUSTELLIGENCEAI_TOKEN→GITHUB_TOKEN`. `GITHUB_TOKEN` was sometimes the github.com token, sometimes the GHE token, changing between restarts → intermittent `401` from `git`/`gh`.

### Why it happens (validated against `main`, nono 0.70.0)

1. `SecretsConfig.mappings` is keyed by the **keystore account** (source), value is the env var (destination):
   `crates/nono-cli/src/profile/mod.rs` — `pub struct SecretsConfig { pub mappings: HashMap<String, String> }` (~L1724).

2. The `extends` merge overrides by **account**, not by destination env var — `crates/nono-cli/src/profile/mod.rs`, `merge_profiles` (~L3633):

   ```rust
   env_credentials: SecretsConfig {
       mappings: {
           let mut merged = base.env_credentials.mappings;
           merged.extend(child.env_credentials.mappings);   // HashMap::extend keys on account
           merged
       },
   },
   ```

   `HashMap::extend` only replaces an entry when the **key** matches. `ACCOUNT_A` and `ACCOUNT_B` are different keys, so **both** entries survive and both target `MY_TOKEN`.

3. There is no destination-collision check. `validate_env_credential_keys` (~L1434) validates URI formats and the destination-name blocklist only.

[...]

PR #1533, opened 2026-07-30
fix(profile): child overrides base in env_credentials by destination env var

dselivanov:
Fixes the `env_credentials` race reported in #1532: when a base profile and a child that `extends` it map **different keystore accounts** to the **same destination env var**, both entries survived the merge and the winning value was chosen non-deterministically per launch.

## Root cause

`SecretsConfig.mappings` is a `HashMap<account, env_var>` — keyed by the **keystore account** (source), with the env var as the value (destination). The `extends` merge used `HashMap::extend`:

```rust
let mut merged = base.env_credentials.mappings;
merged.extend(child.env_credentials.mappings);
```

`extend` only replaces an entry when the **key (account)** matches. Two different accounts targeting the same env var (e.g. a base's `GITHUB_TOKEN -> GITHUB_TOKEN` and a child's `GH_ENTERPRISE_TOKEN -> GITHUB_TOKEN`) therefore both survived. At load time `keystore::load_secrets` iterates the map in random per-process order, and the environment is assembled last-write-wins (`execution_runtime.rs`), so the final value of that env var flipped between launches — silent and intermittent, with no collision warning.

## Fix

Drop any base entry whose destination env var a child overrides before extending, so the child wins by destination — matching `set_vars`, whose merge key *already is* the destination env var (so it has always been child-wins-by-destination):

```rust
let child_destinations: std::collections::HashSet<String> =
    child.env_credentials.mappings.values().cloned().collect();
let mut merged = base.env_credentials.mappings;
merged.retain(|_account, dest| !child_destinations.contains(dest));
merged.extend(child.env_credentials.mappings);
```

`merge_profiles` is the single merge path used by `extends` (single + array) and `platform_overrides`, so this covers all inheritance routes.

## Tests

Two new unit tests in `profile::tests`:

- `env_credentials_child_overrides_base_by_destination_env_var` — base `ACCOUNT_A -> MY_TOKEN` + child `ACCOUNT_B -> MY_TOKEN` => child wins, base entry dropped, unrelated base entry preserved.
- `env_credentials_child_extends_base_without_collision` — no destination collision => both kept.

(The existing `test_merge_profiles_env_credentials_child_wins` covers same-account override, which already worked; these cover the different-account-same-destination case that did not.)

Full `nono-cli` binary suite: **1816 passed, 0 failed.**

## Scope / not in this PR

This fully fixes the **cross-profile** case (base + `extends` child). A rarer, same-profile variant — two accounts mapping to one env var within a *single* profile's `env_credentials` (`{ "A": "T", "B": "T" }`) — still races and is not resolved by this merge change; it would need a validation warning (e.g. via `collect_profile_warnings`). Left as a follow-up noted in #1532 to keep this PR focused.

Refs #1532.


SequeI (approved):
This makes sense, thank you for the fix!

dselivanov:
it;s and amazing how smooth contributions happen nowadays with the correct ai coding agent setup and right engineering culture. My second PR here based on my experience with nono. Like it so far!
PS never written a single line in Rust myself.

PR #1554, opened 2026-07-31
fix(why): fail on unresolvable network policy instead of degrading

### LINKED ISSUE #1548: nono why --profile silently ignores network_policy resolution failures and reports every host as allowed
### What happened?

`resolve_allowed_domains` in `crates/nono-cli/src/why_runtime.rs` silently swallows errors from `load_network_policy` and `resolve_network_profile`. When a profile's domain policy comes only from `network_profile` and that name fails to resolve, the allowlist collapses to empty, `query_network` takes the unconditional-allow path, and `nono why --profile --host` reports every host as allowed.

This reproduces the exact symptom and reason string of #1366 on the code that closed it (#1372, `c344efb`). #1372 fixed the `AllowAll` arm of `query_network` to consult `HostFilter` when `allowed_domains` is non-empty. It is correct, but it depends on the allowlist being populated. This function can empty it without telling anyone.

A single-character typo is enough. Valid names in the embedded policy are `opencode`, `developer`, `codex`, `claude-code`, `minimal`, `enterprise`, several of them hyphenated.

| Profile | `nono why --host definitely-not-allowed.invalid` |
|---|---|
| `network_profile: "claude_code"` (underscore) | `allowed` / `network_allowed` |
| `network_profile: "claude-code"` (correct) | `denied` / `proxy_filtered` |

`nono run` rejects the same profile outright, so the diagnostic contradicts the runtime it describes:

```
$ nono run --profile typo.jsonc -- /bin/echo hi
nono: Configuration parse error: Network profile 'claude_code' not found in policy
```

The failure direction is what makes this worth reporting. A typo converts a fail-closed runtime error into a fail-open diagnostic, and callers reasonably treat `allowed` as authoritative. In our case a downstream tool that only ever queries by host silently discarded every domain request because `why` told it the host was already permitted.

This appears to contradict two rules in `AGENTS.md`: *"Configuration load failures must be fatal. If security lists fail to load, abort"* and *"Fail Secure: On any error, deny access. Never silently degrade to a less secure state."* It is also the documented "Silent fallbacks" footgun.

`nono why --profile` should fail with the same configuration error `nono run` produces, rather than answering from a silently emptied allowlist:

```
nono: Configuration parse error: Network profile 'claude_code' not found in policy
```

More generally, a `why` answer should never be more permissive than what the runtime would enforce for the same profile. If the allowlist cannot be resolved, erroring is the only safe outcome, since both `allowed` and `denied` would be guesses.

### Steps to reproduce

```sh
# 1. Profile whose domain policy comes only from network_profile, with a typo
cat > /tmp/typo.jsonc <<'EOF'
{ "network": { "network_profile": "claude_code" } }
EOF

# 2. why reports an arbitrary host as allowed
nono why --profile /tmp/typo.jsonc --host definitely-not-allowed.invalid --port 443 --json

# 3. run refuses to start on the identical profile
nono run --profile /tmp/typo.jsonc -- /bin/echo hi

# 4. Control: the correct name behaves properly
cat > /tmp/ok.jsonc <<'EOF'
{ "network": { "network_profile": "claude-code" } }
EOF
nono why --profile /tmp/ok.jsonc --host definitely-not-allowed.invalid --port 443 --json
```

### Relevant output or logs

```shell
$ nono why --profile /tmp/typo.jsonc --host definitely-not-allowed.invalid --port 443 --json
{
  "status": "allowed",
  "reason": "network_allowed",
  "access": "Connection to definitely-not-allowed.invalid:443 would be allowed"
}

$ nono run --profile /tmp/typo.jsonc -- /bin/echo hi
  mode supervised (proxy, supervisor)
nono: Configuration parse error: Network profile 'claude_code' not found in policy

$ nono why --profile /tmp/ok.jsonc --host definitely-not-allowed.invalid --port 443 --json
{
  "status": "denied",
  "reason": "proxy_filtered",
  "details": "Domain filtering is active. host definitely-not-allowed.invalid is not in the allowlist",
  "policy_source": "proxy domain filter"
}
```

### nono version

0.70.0 (built from `main` at `c22b7537`)

### Operating system

macOS

### OS version / distro

26.5.1

_No response_

### Profile (if relevant)

```json
{ "network": { "network_profile": "claude_code" } }
```

### Additional context

**Scope.** The fail-open case is specifically a profile whose domain policy comes *only* from `network_profile`. Add any `allow_domain` entry alongside the bad `network_profile` and the result is correctly `proxy_filtered`, because the non-empty allowlist keeps `HostFilter` engaged. `--path` queries are unaffected. That "only `network_profile`" shape is the one #1366 was originally filed against.

**Source.** `crates/nono-cli/src/why_runtime.rs:19`, unchanged since `85f0acca` (predates #1372):

```rust
let net_policy = match network_policy::load_network_policy(policy_json) {
    Ok(p) => p,
    Err(_) => {
        return profile.network.allow_domain.iter()
            .map(|e| e.domain().to_string()).collect();   // policy load failure ignored
    }
};

if let Some(net_profile_name) = profile.network.resolved_network_profile()
    && let Ok(resolved) = network_policy::resolve_network_profile(&net_policy, net_profile_name)
{                                                         // resolution failure ignored
```

Two distinct silent failures. The `load_network_policy` arm returns raw unexpanded entries, which can also produce false *denials* where `expand_proxy_allow` would have widened the list. The `resolve_network_profile` arm drops the profile's hosts entirely.

**Other triggers besides typos.** Any `network_profile` name valid in one nono version and renamed or removed in another, and `extends` chains that reference a name the embedded policy no longer ships. These fail the same way, on profiles that used to work.

**Suggested fix.** Change the signature to `Result<Vec<String>>` and propagate both errors with `?`. The single call site in `run_why` already returns `Result`, so it is a one-character change there. I have this working locally against `main` at `c22b7537`, with 8 tests covering the error path plus the exact-match and wildcard-apex semantics (`auth0.com` not granting `foo.auth0.com`, `*.auth0c.com` granting `x.auth0c.com` but not the apex). Verified: `cargo test -p nono-cli --bins`, `cargo test -p nono --lib`, clippy with `-D warnings -D clippy::unwrap_used`, and `cargo fmt --check` all clean on the changed file. Happy to open a PR if you would like it.

**Disclosure.** This issue was investigated and written by an AI agent (Claude Opus 5) working with a human collaborator. Findings were verified by building `main` at `c22b7537` and running the commands above; the version-table and typo results are measured output, not inference.

fyockm:
## Linked Issue

Closes #1548

## Summary

`resolve_allowed_domains` in `crates/nono-cli/src/why_runtime.rs` swallowed errors from both `load_network_policy` and `resolve_network_profile`:

```rust
let net_policy = match network_policy::load_network_policy(policy_json) {
    Ok(p) => p,
    Err(_) => { return profile.network.allow_domain.iter()... }  // silent fallback
};
```

Two distinct silent failures, in opposite directions:

- The `load_network_policy` arm returned raw unexpanded entries, skipping `expand_proxy_allow`, which can produce false **denials**.
- The `resolve_network_profile` arm dropped the network profile's hosts entirely. For a profile whose domain policy comes *only* from `network_profile`, the allowlist collapsed to empty, `query_network` took the unconditional-allow path, and every host reported `allowed` / `network_allowed` — a false **allow**.

A single-character typo was enough to turn a fail-closed runtime error into a fail-open diagnostic. `nono run` rejects the same profile outright, so the diagnostic contradicted the runtime it describes.

Both errors now propagate with `?`. The single call site in `run_why` already returns `Result`, so it is a one-character change there.

This is the "Silent fallbacks" footgun from AGENTS.md, and it contradicted two stated rules: *"Configuration load failures must be fatal"* and *"Fail Secure: on any error, deny access. Never silently degrade to a less secure state."* It matters more than usual here because callers treat a `why` answer as authoritative.

## Agent Disclosure

This PR was prepared by an AI agent (Claude Opus 4.8) working with @fyockm.

Files consulted: `why_runtime.rs`, `query_ext.rs`, `network_policy.rs`, `execution_runtime.rs`, `proxy_runtime.rs`, `profile/mod.rs`, `capability.rs`, `CONTRIBUTING.md`, and `AGENTS.md`.

Attribution: all logic in this PR is newly written. It builds on `main` at `c22b7537`, which already contains #1372 (`c344efb`, Aleksy Siek) — that fix made `query_network` consult `HostFilter` in the `AllowAll` arm when `allowed_domains` is non-empty. This change addresses the remaining path by which that allowlist could be silently emptied.

## Test Plan

The exact reproduction from #1548, against the built binary:

| Scenario | Before | After |
|---|---|---|
| `network_profile: "claude_code"` (typo) | `allowed` / `network_allowed`, exit 0 | `Configuration parse error: Network profile 'claude_code' not found in policy`, exit 1 |
| `claude-code` + `definitely-not-allowed.invalid` | `denied` / `proxy_filtered` | unchanged |
| `claude-code` + `api.anthropic.com` | `allowed` / `proxy_allowed` | unchanged |

The `after` output matches the issue's stated expectation exactly.

Eight tests added to `why_runtime.rs`. `unknown_network_profile_is_an_error` is the regression test; the rest pin the matching semantics #1372 restored so a future refactor cannot regress them silently:

- `profile_plain_entry_does_not_grant_subdomains` — matching is exact rather than subtree, so `auth0.com` does not grant `foo.auth0.com`
- `profile_wildcard_grants_subdomain_not_apex` — `*.auth0c.com` grants `x.auth0c.com` and denies the bare apex
- `profile_host_in_allowlist_is_allowed`, `profile_host_outside_allowlist_is_denied`
- `known_network_profile_contributes_hosts`, `profile_without_network_config_reports_allowed`, `profile_proxy_denies_cloud_metadata`

Verification on macOS (native toolchain):

| Check | Result |
|---|---|
| `cargo test -p nono-cli --bins` | 1835 passed, 2 pre-existing failures |
| `cargo test -p nono --lib` | 747 passed |
| `cargo clippy --no-deps -p nono-cli --all-targets -- -D warnings -D clippy::unwrap_used` | clean |
| `cargo fmt --check -p nono-cli` | clean |

### Pre-existing failures, not introduced here

Both confirmed against the baseline with `git stash` (unmodified `main` at `c22b7537` gives `1827 passed, 2 failed` — the same two). Left alone rather than bundling unrelated fixes.

- `cli::tests::test_extends_requires_profile`
- `profile::tests::test_load_nonexistent_profile`

Both appear environment-dependent. Happy to file a separate issue if useful.

## Related: the same swallow pattern remains in `execution_runtime.rs`

Deliberately **not** fixed here, since it is outside the scope of #1548 and would change `nono run` startup behavior rather than a diagnostic.

`crates/nono-cli/src/execution_runtime.rs:235-269` resolves the same allowlist for the sandbox state file (the data `nono why --self` later answers from) and degrades on both the same errors, with only a `warn!`:

```rust
Err(e) => {
    warn!("failed to resolve network_profile for sandbox state: {e}");
}
...
Err(e) => {
    warn!("failed to load network policy for sandbox state: {e}");
    plain_domain_strs
}
```

The exposure is narrower than the `why --profile` case, because `nono run` validates the profile elsewhere and aborts on an unknown `network_profile` before reaching this code. So the reachable consequence today is a state file that under-reports the allowlist rather than over-reports it — a false denial in `why --self`, not a false allow. That is why it is a separate concern and not part of this fix.

It is still a silent degrade on a security-relevant list, so it is worth its own issue. Two options for maintainers, happy to follow up with either:

1. Propagate, matching this change and `proxy_runtime.rs:2285` (which already uses `?`).
2. Keep it non-fatal but make the degrade explicit in the state file, so `why --self` can say "allowlist incomplete" instead of answering from a truncated list.

Flagging rather than choosing, since the call depends on whether a state-write failure should be able to abort a run that would otherwise be correctly sandboxed.

## Checklist

- [x] Public-facing changes are paired with documentation updates — n/a, no documented `why` exit-code contract changes (checked `docs/cli/usage/flags.mdx`, `docs/cli/getting_started/quickstart.mdx`); `CHANGELOG.md` is generated from commit messages

## Agent Compliance Check

- [x] An issue already exists (#1548)
- [x] I disclosed that I am an agent in the issue discussion (#1548 carries an AI-agent disclosure)
- [x] I described my intent and approach in the issue discussion (#1548 states the suggested fix, the `Result` signature change, and the test coverage this PR implements)
- [x] I did not use forbidden patterns such as unwrap/expect (test-only `expect` inside `#[cfg(test)]`)
- [x] I used NonoError where required (both propagated errors are `NonoError::ConfigParse`, unchanged from their source)
- [x] I validated and canonicalized all relevant paths (no path handling changed)


SequeI (approved):
lgtm, thank you!!

PR #1585, opened 2026-08-06
fix(proxy): fail closed on non-granted endpoint approvals in TLS intercept

### LINKED ISSUE #1584: Endpoint-policy `approve` fails on the TLS-intercept path
## Summary

On the CONNECT/TLS-intercept proxy path, a **non-granted** endpoint-policy
approval (denied, backend error, or timeout) on a managed-credential route does
not block the request. Instead the route is dropped from selection and the
request falls through to the "no route matched" branch, which forwards it
upstream **without** the managed credential.

The net effect is a fail-open on the approval gate: a request an operator
explicitly denied still egresses. The reverse-proxy path already fails closed
(returns 403), so this is a divergence between the two proxy modes.

## Severity / scope

I did consider if this needed an embargoed fix, but its a feature undocumented, not surfaced in nono as a useable flag and the risk is low, but a genuine enforcement fail-open worth fixing:

- **Not a credential leak.** The dropped route means no injection happens, so
  the upstream receives the request without the managed secret. Confidentiality
  of the token is preserved; what breaks is the deny/approval gate on egress.
- **Unused in practice.** No shipped or embedded profile uses endpoint-policy
  `approve` on a managed-credential route (the only in-repo references to
  `approval_backends` are in `nono-profile.schema.json`). The decision value is
  documented at `docs/cli/features/tool-sandbox.mdx:209`, but there is no worked
  example and no default profile that exercises it.
- Because there is no affected user population and no disclosure risk, this is a
  normal public bug fix, not a security-advisory/embargo case.

## Affected code

`crates/nono-proxy/src/tls_intercept/handle.rs`, in `select_intercept_route`,
the `EndpointPolicyOutcome::Approve` arm. On a non-granted decision the code did:

```rust
Ok(Ok(Ok(_))) => {
    // ...audit approve_denied...
    if !route.requires_managed_credential {
        has_endpoint_only_route = true;
    }
    // falls through — managed-credential route silently dropped
}
```

A managed-credential route (`requires_managed_credential == true`) hits neither
branch, so it is dropped and route selection ends at:

```
tls_intercept: no endpoint_rules matched … forwarding without credentials
```

The explicit `EndpointPolicyOutcome::Deny` arm in the same function correctly
returns `RouteSelection::Rejected(403)`; the approve-denied path did not.

## Reproduction

Profile with a managed-credential route whose endpoint policy routes a path to
an approval backend:

```jsonc
"credentials": {
  "httpbin": {
    "type": "proxy",
    "upstream": "https://httpbingo.org",
    "credential_key": "env://HTTPBIN_TOKEN",
    "env_var": "HTTPBIN_TOKEN",
  }
}
// curl command grants credential "httpbin" with:
"endpoint_policy": {
  "default": "deny",
  "approve": [{ "method": "GET", "path": "/headers", "backend": "local-webhook" }],
  "allow":   [{ "method": "GET", "path": "/get" }]
}
```

Run an approval backend that **denies** the `/headers` request, then:

```bash
nono run -s --allow-cwd --profile ./profile.json -- curl -s https://httpbingo.org/headers
```

**Expected:** proxy returns 403; request never reaches the upstream.

**Actual (before fix):** curl gets a 200 from the upstream, with no
`Authorization` header (request forwarded un-credentialed). The audit trail for
the session shows `approve_requested` → `approve_denied` → **`allow`** for the
same `GET /headers`.

## Secondary issue: approval reason dropped from audit

Denied approvals record a hardcoded reason instead of the backend's actual one,
so the audit trail says *that* a request was denied but not *why*:

- Command approvals (`crates/nono-cli/src/tool-sandbox/platform/{macos,linux}.rs`)
  record `reason: "approval_denied"`.
- Endpoint approvals log `"endpoint approval denied"`.

The backend's `ApprovalDecision::Denied { reason }` is available at both sites
and should be propagated.

## Environment

- nono 0.71.0
- macOS (Darwin 24.6); the fail-open is on the CONNECT/TLS-intercept path. The
  reverse-proxy path was already fail-closed.

lukehinds:
A denied, errored, or timed-out endpoint approval on a managed-credential route was dropped from route selection instead of rejecting the request, so the request forwarded upstream without its credential — approval 'deny' only withheld injection rather than blocking. Return 403 from all three arms, matching the explicit deny arm and the reverse-proxy path.

Also propagate the approval backend's stated reason into audit events and deny messages instead of hardcoded strings: command approvals now record 'approval_denied: <backend reason>' and endpoint approvals 'endpoint approval denied: <backend reason>', so the audit trail shows who refused and why.

## Linked Issue

Closes #1584

## Checklist


SequeI (approved):
lgtm, it's a nice refactor too. Thanks!

PR #1589, opened 2026-08-06
fix(proxy): redeem phantom nonces on absolute-form forward-proxy requests

### LINKED ISSUE #1588: Absolute-form forward-proxy requests bypass credential-route matching and phantom redemption
## What happened?

A profile can declare a `custom_credentials` route with a loopback `upstream` (e.g. `http://127.0.0.1:8787`) and `redeem_phantoms` configured, so a captured credential is swapped in before forwarding. This works over `CONNECT` (TLS-intercept).

But when a client is configured to hit that loopback address directly and does so via nono's `HTTP_PROXY` env injection, the request arrives in **absolute-form** (`GET http://127.0.0.1:8787/path HTTP/1.1`). `handle_connection` in `crates/nono-proxy/src/server.rs` dispatches this to `handle_forward_http`, which never consults `route_store` for a matching route, forwards headers verbatim including any phantom nonce, and logs nothing about the request at all.

## What did you expect to happen?

`handle_forward_http` should check whether the target host:port matches a configured route's upstream (the same way `CONNECT` does via `RouteStore::lookup_by_upstream`/`is_route_upstream`) and redeem any phantom nonce before forwarding — consistent with the same route's behavior over `CONNECT`.

## Steps to reproduce

1. Run a local HTTP server on `127.0.0.1:9000` that echoes back its `Authorization` header.
2. Declare a `custom_credentials` route:
   ```json
   {
     "network": {
       "credentials": ["local"],
       "custom_credentials": {
         "local": {
           "upstream": "http://127.0.0.1:9000",
           "redeem_phantoms": ["my-cred"],
           "endpoint_rules": [{ "method": "*", "path": "/*" }]
         }
       }
     }
   }
   ```
3. Capture a credential named `my-cred` via a `capture_credential` intercept so the agent receives a phantom (`nono_<hex>`) instead of the real value.
4. Inside `nono run --profile ...`, send a request with `Authorization: Bearer <phantom>` to `http://127.0.0.1:9000` using a client that honors `HTTP_PROXY` (e.g. `curl`).
5. The mock server's echoed `Authorization` header still contains the raw `nono_<hex>` phantom.

## Relevant output or logs

```
DEBUG Loading route 'local' -> http://127.0.0.1:9000
INFO    http://127.0.0.1:9000 | creds: none | intercept: on | endpoint_rules: 1
DEBUG Accepted connection from 127.0.0.1:<ephemeral>
DEBUG Supervised child exited with code 0
```
No policy-decision or redemption log line for the request, despite the route loading and the connection reaching nono's proxy.

## nono version

0.71.0

## Operating system

macOS

## OS version / distro

macOS 15 (Sequoia)

## Kernel version (Linux only)

N/A

## Profile (if relevant)

```json
{
  "network": {
    "credentials": ["local"],
    "custom_credentials": {
      "local": {
        "upstream": "http://127.0.0.1:9000",
        "redeem_phantoms": ["my-cred"],
        "endpoint_rules": [{ "method": "*", "path": "/*" }]
      }
    }
  }
}
```

kipz:
## Linked Issue

Closes #1588

## Summary

`handle_forward_http` (the absolute-form `http://…` dispatch branch) never consulted `route_store`, so a `custom_credentials` route with a loopback upstream and `redeem_phantoms` configured was silently unreachable when a client sent an absolute-form request via `HTTP_PROXY` — the phantom nonce passed straight through to the real upstream. This adds a route lookup (by upstream host:port, mirroring the existing `CONNECT` path) and redeems any phantom nonce in headers before forwarding.

## Test Plan

- `make fmt && make ci` — clean (clippy `-D warnings -D clippy::unwrap_used`, fmt-check, full test suite, audit, lint-aliases, lint-docs). One pre-existing, order-independent flake (`command_runtime::tests::shell_dry_run_rejects_block_net_with_upstream_proxy`, a TMPDIR/state-root race under parallel test execution, unrelated to this change) — confirmed passing in isolation, both before and after this change.
- Added 3 new unit tests for the new `strip_and_redeem_proxy_headers` helper: redeems an admitted nonce, leaves an unadmitted nonce unchanged (fail-open, matching existing CONNECT-path behavior), leaves headers with no nonce unchanged.
- Verified live end-to-end against a real downstream reproduction: a real captured phantom sent via absolute-form `HTTP_PROXY` request to a real local loopback service returned `401` before this fix (phantom forwarded unredeemed) and a real successful response after this fix, with the new debug log line confirming the route match and redemption path fired.

## Checklist

### DIFF
--- a/crates/nono-proxy/src/server.rs
+++ b/crates/nono-proxy/src/server.rs
@@ -177,6 +177,48 @@ fn strip_proxy_headers(header_bytes: &[u8]) -> Vec<u8> {
     out
 }
 
+/// Like [`strip_proxy_headers`], but also redeems any `nono_…` phantom nonce
+/// via `resolver`/`consumer`, fail-open per [`tls_intercept::handle::resolve_nonce_in_header_value`].
+fn strip_and_redeem_proxy_headers(
+    header_bytes: &[u8],
+    consumer: &str,
+    resolver: &dyn crate::token::NonceResolver,
+) -> Vec<u8> {
+    let header_str = match std::str::from_utf8(header_bytes) {
+        Ok(s) => s,
+        Err(_) => return header_bytes.to_vec(),
+    };
+    let mut out = Vec::with_capacity(header_bytes.len());
+    for line in header_str.split_inclusive("\r\n") {
+        let Some((name, rest)) = line.split_once(':') else {
+            out.extend_from_slice(line.as_bytes());
+            continue;
+        };
+        let trimmed_name = name.trim();
+        if trimmed_name.eq_ignore_ascii_case("proxy-connection")
+            || trimmed_name.eq_ignore_ascii_case("proxy-authorization")
+        {
+            continue;
+        }
+        let (value, line_ending) = match rest.strip_suffix("\r\n") {
+            Some(v) => (v, "\r\n"),
+            None => (rest, ""),
+        };
+        match tls_intercept::handle::resolve_nonce_in_header_value(value.trim(), consumer, resolver)
+        {
+            Some(resolved) => {
+                out.extend_from_slice(name.as_bytes());
+                out.push(b':');
+                out.push(b' ');
+                out.extend_from_slice(resolved.as_bytes());
+                out.extend_from_slice(line_ending.as_bytes());
+            }
+            None => out.extend_from_slice(line.as_bytes()),
+        }
+    }
+    out
+}
+
 #[must_use]
 fn proxy_diagnostic_code_label(code: crate::diagnostic::ProxyDiagnosticCode) -> &'static str {
     code.as_str()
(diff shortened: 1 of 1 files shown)


kipz:
Added an end-to-end test (`forward_http_matching_route_upstream_redeems_phantom`) exercising the fix via a real proxy instance and route config, not just the header-rewriting helper in isolation.

crates/nono-proxy/src/server.rs
```diff
@@ -1982,9 +2042,10 @@ async fn handle_forward_http(
         log: Some(&state.audit_log),
         mode: audit::ProxyMode::Reverse,
         event_ctx: audit::EventContext {
+            route_id: matched_route.as_deref(),
             auth_mechanism: Some(nono::undo::NetworkAuditAuthMechanism::ProxyAuthorization),
             auth_outcome: Some(nono::undo::NetworkAuditAuthOutcome::Succeeded),
-            managed_credential_active: Some(false),
+            managed_credential_active: Some(matched_route.is_some()),
```
SequeI: `managed_credential_active` should reflect whether a credential was actually swapped in for this request, not just whether the target matched a route.

 A route can match with no credential configured (e.g. a declarative route that only exists to supply a `*_BASE_URL` env var), or a request can hit a credentialed route without carrying a valid/admitted nonce (the fail-open case)

In both cases `matched_route.is_some()` is true but no real credential was ever injected. Can we base this on whether `strip_and_redeem_proxy_headers` actually performed a swap, rather than on route match alone?"

SequeI (changes requested):
One nit for the audit log, otherwise looks good! I appreciate the test coverage :)

SequeI (approved):
lgtm, thanks Kipz!

PR #1617, opened 2026-08-10
fix(proxy): bound per-line reads of the WebSocket upgrade response

### LINKED ISSUE #1616: nono-proxy: unbounded per-line buffering when reading a WebSocket upgrade response
## Problem

`read_response` in the TLS-intercept WebSocket path reads the upstream upgrade
response with `BufReader::read_line`, which grows its buffer until it sees a
`\n`:

- [`websocket.rs:42`](https://github.com/nolabs-ai/nono/blob/4363d4890223bab5912988236138c0db74689e34/crates/nono-proxy/src/tls_intercept/websocket.rs#L42) — status line
- [`websocket.rs:52`](https://github.com/nolabs-ai/nono/blob/4363d4890223bab5912988236138c0db74689e34/crates/nono-proxy/src/tls_intercept/websocket.rs#L52) — each header line

There are two existing limits, and neither bounds a single line:

- `MAX_HEADER_SIZE` (64 KiB, [`websocket.rs:64`](https://github.com/nolabs-ai/nono/blob/4363d4890223bab5912988236138c0db74689e34/crates/nono-proxy/src/tls_intercept/websocket.rs#L64)) is only checked *after* a
  complete line has been buffered, and not at all for the status line.
- `UPSTREAM_HEADER_TIMEOUT` (30s, [`websocket.rs:80`](https://github.com/nolabs-ai/nono/blob/4363d4890223bab5912988236138c0db74689e34/crates/nono-proxy/src/tls_intercept/websocket.rs#L80)) bounds elapsed time, not
  memory.

So an upstream that streams bytes without ever sending a newline makes the
proxy allocate for 30 seconds at line rate before anything trips. On a fast
link that is a large allocation per connection, in the supervisor process that
holds the broker's real credentials.

## Threat model

The upstream host is reachable only via a configured intercept route, so this
is not arbitrary-attacker-controlled. It is still a remote input that the proxy
buffers without a size bound, and the fail-secure rule in `AGENTS.md` says the
limit should be explicit rather than implied by a timeout.

## Proposed fix

Add a `MAX_LINE_SIZE` cap (8 KiB) and a small `read_line_limited` helper over
`AsyncBufRead` that errors with `InvalidData` once that many bytes accumulate
without a newline, checking the cap on both the newline-found and no-newline
paths. `read_response` uses it for the status line and for each header line;
`MAX_HEADER_SIZE` keeps bounding the cumulative total.

## Notes

- Filed by an agent on behalf of @kipz.
- The change was originally written as review follow-up on #1469 and removed
  from that PR because it is unrelated to that PR's feature; this issue exists
  so it can land on its own.

kipz:
Closes #1616

`read_line` grows its buffer until it sees a newline, so an upstream that
streams bytes without one made the proxy allocate for the full 30s header
timeout. `MAX_HEADER_SIZE` only applied once a whole line had been buffered,
and never to the status line at all.

Adds `MAX_LINE_SIZE` (8 KiB) and a `read_line_limited` helper over
`AsyncBufRead` that errors with `InvalidData` once that many bytes accumulate,
checking the cap on both the newline-found and no-newline branches.
`MAX_HEADER_SIZE` still bounds the cumulative total.

Contributed by an agent on behalf of @kipz.

## Agent Compliance Check

- [x] An issue already exists

### DIFF
--- a/crates/nono-proxy/src/external.rs
+++ b/crates/nono-proxy/src/external.rs
@@ -11,8 +11,9 @@ use crate::audit;
 use crate::config::ExternalProxyConfig;
 use crate::error::{ProxyError, Result};
 use crate::filter::ProxyFilter;
+use crate::line_reader;
 use crate::token;
-use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
+use tokio::io::{AsyncWriteExt, BufReader};
 use tokio::net::TcpStream;
 use tracing::debug;
 use zeroize::Zeroizing;
@@ -61,15 +62,18 @@ pub async fn connect_via_proxy(
 
     let mut buf_reader = BufReader::new(&mut proxy_stream);
     let mut response_line = String::new();
-    buf_reader
-        .read_line(&mut response_line)
-        .await
-        .map_err(|e| {
-            ProxyError::ExternalProxy(format!(
-                "failed to read response from external proxy: {}",
-                e
-            ))
-        })?;
+    line_reader::read_line_limited_string(
+        &mut buf_reader,
+        &mut response_line,
+        line_reader::MAX_LINE_SIZE,
+    )
+    .await
+    .map_err(|e| {
+        ProxyError::ExternalProxy(format!(
+            "failed to read response from external proxy: {}",
+            e
+        ))
+    })?;
 
     let status = parse_status_code(&response_line)?;
     if status != 200 {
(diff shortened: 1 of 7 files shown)


SequeI:
Once comments left by nogent are addressed, we can get this in. Thanks!

kipz:
Addressed all 4 findings from nogent's review in 25b87194:

- `server.rs:1375` (`handle_connection`) — first request line and header lines now read via a bounded line reader.
- `tls_intercept/handle.rs:309` (`parse_inner_request`) — request/status line and header lines now bounded.
- `tls_intercept/http1.rs:191` (`relay_chunked`) — chunk `size_line` and `trailer` reads from the upstream now bounded.
- `external.rs:65` (`connect_via_proxy`) — status line and header lines from the enterprise proxy now bounded.

Extracted the bounded-read helper this PR already introduced for the WebSocket upgrade response (`tls_intercept/websocket.rs`) into a shared `crate::line_reader` module (`read_line_limited` / `read_line_limited_string`), and reused it at all 4 sites plus the existing WebSocket call site, so there's one audited implementation instead of five copies. Each read is capped at 8KB per line (`MAX_LINE_SIZE`).

`make ci` (clippy `-D warnings -D clippy::unwrap_used`, fmt, full workspace test suite) is green.

SequeI (approved):
lgtm!

PR #1650, opened 2026-08-16
fix(tool-sandbox): fall back to the interpreter's directory for RPATH-less ELF dependencies

### LINKED ISSUE #1649: tool-sandbox: ELF dependency resolution fails on Nix/NixOS when a NEEDED library has no RPATH/RUNPATH (e.g. gcc's libgcc_s.so.1 → libc.so.6)
### What happened?

Running any command through `nono run` on NixOS fails during sandbox setup whenever the target binary's dependency closure includes a shared library that has no `DT_RPATH`/`DT_RUNPATH` of its own and needs another library by bare soname — for example `libgcc_s.so.1` (from the `gcc-*-libgcc` package) needing `libc.so.6`. `resolve_shared_library` in `crates/nono-cli/src/tool-sandbox/platform/linux.rs` only searches the referencing object's own RPATH/RUNPATH plus a hardcoded list of FHS directories (`/lib`, `/usr/lib`, etc.), none of which exist as such on NixOS, so resolution fails.

nono resolves the dependency the same way the real dynamic linker does, and
the sandbox starts normally.

### Steps to reproduce

1. On NixOS (or with a Nix-installed toolchain on any distro), install a package whose runtime depends on `libgcc_s.so.1` — e.g. `nix profile install nixpkgs#claude-code`, or anything else linked against `libgcc`.
2. Run it through nono: `nono run --profile <any> --allow-cwd -- <path-to-binary> --help`.

### Relevant output or logs

```shell
nono: Sandbox initialization failed: failed to resolve ELF dependency 'libc.so.6' for /nix/store/<hash>-gcc-15.3.0-libgcc/lib/libgcc_s.so.1


Confirmed with `readelf -d`: `libgcc_s.so.1`'s `RUNPATH` is empty and it `NEEDED`s `libc.so.6` by bare soname. `ldd` on the same file resolves it fine — via the ELF interpreter (`ld-linux-x86-64.so.2`) that the *requesting* top-level binary's own `PT_INTERP` names, which lives right next to `libc.so.6` in that interpreter's own store output directory.
```

### nono version

0.73.0

### Operating system

Linux

### OS version / distro

NixOS 26.11

6.18.44

### Profile (if relevant)

```json
Any profile that has a `command_policies` set seems to reproduce.
```

### Additional context

This issue and its linked fix were prepared with AI assistance (Claude Code), per this repository's `AGENTS.md` Coding Agent Contribution Policy.

- **Intent**: extend `resolve_shared_library`'s fallback search to cover   the case above, without changing behavior on FHS/non-Nix systems.
- **Planned approach**: thread the directory of the *top-level* binary's  own ELF interpreter (`PT_INTERP`) through the dependency-closure walk as an extra fallback search directory, tried after the referencing object's own RPATH/RUNPATH and before the FHS defaults. This mirrors how the real dynamic linker resolves the same case (its own compiled-in default search path is that same directory) instead of hardcoding anything Nix-specific.
- **Risks/tradeoffs**: none identified for non-Nix systems — there, the interpreter directory is normally `/lib` or `/lib64`, already covered by the existing default list, so the change is a no-op. On Nix/NixOS it can grant read access to a few additional libraries that live in the interpreter's own directory; those are exactly the libraries the real dynamic linker would map into the process anyway, so this doesn't broaden the sandbox's effective permission surface.

I'll open a draft PR implementing this immediately, and use it as a patch on my local system — if the suggested approach works for you I'll be happy to un-draft it!

### LINKED ISSUE #1724: tool-sandbox: outer exec gate denies top-level shebang/wrapper commands when command_policies is active (exit 127 on NixOS)
### What happened?

When a profile defines `command_policies`, the Linux supervised child applies the
tool-sandbox outer exec gate (`build_outer_exec_files` → `apply_outer_exec_gate`),
which allow-lists `Execute` per file: shims, direct bypasses, and executables
enumerated from PATH search dirs plus their ELF dependency closures.

The initial top-level command is never explicitly added to that allow-list. It only
works by coincidence of living inside an enumerated search dir — and even then only
its `#!` interpreter is followed, not what the script executes at runtime.

On NixOS this breaks most binaries, because nixpkgs executables are multi-level
bash wrapper chains:

```
pi          → #!/nix/store/byi2zpy2bgcf3dr6y0l8m50rmjj8z7q1-bash-5.3p15/bin/bash   (wrapper)
  .pi-wrapped → exec /nix/store/glcp73hgagq2b24i80jlgbvj28vdb6kk-nodejs-24.19.0/bin/node …
```

Running `nono run --profile pi-cmd -- pi --version` fails with exit 127 and a
misleading "its directory is not readable" hint (that diagnostic compares the
unresolved symlink path against canonicalized capabilities). The real failure is
EACCES from the stacked Execute-only Landlock layer, which the path-denial observer
doesn't see.

A top-level command that the outer session sandbox can already read should be
executable whether it is an ELF binary, a shebang script, or a multi-stage NixOS
wrapper chain — without weakening the gate's core property that controlled/deny-only
binary identities may only run through their shims.

### Steps to reproduce

```bash
# 1. Any profile with an active command_policies section (e.g. one defining gh/kubectl).
# 2. Run any NixOS wrapper-wrapped binary as the top-level command:
nono run --profile pi-cmd -- pi --version

# 3. Bisect under the same profile:
nono run --profile pi-cmd -- /bin/sh -c 'echo ok'    # works (resolves into an enumerated system dir)

printf '#!/run/current-system/sw/bin/sh\necho ok\n' > /tmp/a.sh; chmod +x /tmp/a.sh
nono run --profile pi-cmd -- /tmp/a.sh               # works

printf '#!/nix/store/byi2zpy2bgcf3dr6y0l8m50rmjj8z7q1-bash-5.3p15/bin/bash\necho ok\n' > /tmp/b.sh
chmod +x /tmp/b.sh
nono run --profile pi-cmd -- /tmp/b.sh               # exit 127

nono run --profile pi-cmd -- /nix/store/byi2zpy2bgcf3dr6y0l8m50rmjj8z7q1-bash-5.3p15/bin/bash -c 'echo ok'   # exit 127
nono run --profile pi-cmd -- /nix/store/glcp73hgagq2b24i80jlgbvj28vdb6kk-nodejs-24.19.0/bin/node --version   # exit 127
```

### Relevant output or logs

```shell
Failed to execute command (exit code 127).

The executable 'pi' was resolved at:
  /home/user/.nix-profile/bin/pi
but its directory is not readable inside the sandbox.

Fix: grant read access to the binary's directory:
  nono run --read /home/user/.nix-profile/bin ...

No path denials were observed during this session.
The failure may be unrelated to sandbox restrictions.


(Note: read access to that directory *is* granted and confirmed via
`nono why --path /home/user/.nix-profile/bin/pi --op read`; the message is
unrelated to the actual failure.)
```

### nono version

0.74.0

### Operating system

Linux

### OS version / distro

NixOS (Landlock ABI V6)

7.2.0

### Profile (if relevant)

```json
Any profile with an active `command_policies` section, e.g. one defining
policy-controlled commands (`gh`, `kubectl`, ...). Normal filesystem/network grants;
`~/.nix-profile/bin` and `/nix/store` are readable and confirmed allowed via
`nono why`.
```

### Additional context

This is after applying the fix from https://github.com/nolabs-ai/nono/pull/1650. Without it, it gets blocked on ELF issues

Root cause sketch:

1. `build_outer_exec_files()` relies on PATH-dir enumeration for non-controlled
   executables; shebang interpreters and wrapper targets in other directories are
   invisible to it, and the initial resolved program itself is never passed into
   prepare-time gate construction.
2. `elf_dependency_closure()` follows the `#!` interpreter, but nothing follows
   absolute executable paths referenced inside a wrapper script's body
   (`.pi-wrapped`, node), so even granting interpreters would not cover NixOS
   wrapper chains.

Possible fix direction: pass the initial resolved program into
`ToolSandboxPrepare`, add it (plus closure) to `allowed_outer_exec_files` while
skipping controlled/deny-only inode identities, and extend the closure walk so
shebang scripts contribute absolute executable paths found in their body (bounded
scan, best-effort). Safety: stacked Landlock layers intersect and the main layer
maps `AccessMode::Read` → `ReadFile | ReadDir | Execute`, so extra entries cannot
widen effective access beyond the outer session grants.

Related: same root-cause family as #1699 (shebang helpers via `exec_paths` in
brokered child sandboxes); this covers the distinct top-level/outer-gate surface
and the NixOS wrapper-chain case that fix would not reach.

Ekleog:
## Linked Issue

Closes #1649
Closes #1724 

## Summary

`resolve_shared_library`'s ELF dependency-closure walk only searched a
`NEEDED` library's own `DT_RPATH`/`DT_RUNPATH` plus a hardcoded FHS default
list (`/lib`, `/usr/lib`, etc.). On Nix/NixOS this fails whenever a shared
object has no RPATH/RUNPATH of its own — e.g. `libgcc_s.so.1` (from the
`gcc-*-libgcc` package) needing `libc.so.6` by bare soname — because none of
the FHS defaults exist as such there, so `nono run` refused to start any
command whose dependency closure reached that library.

The real dynamic linker resolves this case fine: a Nix-built `ld.so` has its
own directory compiled in as its default search path, so an RPATH-less
`NEEDED` entry still resolves through it. This PR threads the *top-level*
binary's own ELF interpreter (`PT_INTERP`) directory through the recursive
closure walk as an extra fallback search directory — tried after the
referencing object's own RPATH/RUNPATH and before the FHS defaults — so
`resolve_shared_library` mirrors that real resolution order instead of
hardcoding anything Nix-specific.

On FHS/non-Nix systems this is a no-op: the interpreter directory there is
normally `/lib` or `/lib64`, already covered by the existing default list.
The `ELF_LIB_CACHE` memoization key was widened to include the interpreter
directory so a resolution computed under one binary's interpreter doesn't
leak into a different binary's closure within the same launch.

Verified directly against `libgcc_s.so.1` on NixOS with `readelf -d` (empty
`RUNPATH`) and `ldd` (resolves via the interpreter's own directory).

This PR was prepared with AI assistance (Claude Code). Files/sections
consulted: `crates/nono-cli/src/tool-sandbox/platform/linux.rs`
(`elf_dependency_closure`, `resolve_elf_recursive`, `resolve_shared_library`,
`parse_elf`/`parse_dynamic`, and the surrounding `ELF_*_CACHE` memoization),
this repository's `AGENTS.md` (Coding Agent Contribution Policy, coding
standards, security considerations), and the linked issue.

## Test Plan

- `cargo test -p nono-cli --bins tool_sandbox::linux` — 34 passed, including
  6 new/changed tests covering the fallback, its precedence under RPATH,
  the not-found error path, and the real `PT_INTERP`-parsing path via
  `std::env::current_exe()`.
- `cargo fmt --all -- --check` and `cargo build --release -p nono-cli` — clean.
- Manual repro: `nono run --profile claude --allow-cwd -- ~/.nix-profile/bin/claude --dangerously-skip-permissions --help` on NixOS no longer fails with the ELF-resolution error.

## Checklist

- [x] Public-facing changes are paired with documentation updates — n/a, this is internal sandbox-initialization behavior with no public API or documented contract change

- [x] An issue already exists
- [x] I provided required attribution for reused or adapted code — n/a, no reused/adapted code

### DIFF
--- a/crates/nono-cli/src/tool-sandbox/platform/linux.rs
+++ b/crates/nono-cli/src/tool-sandbox/platform/linux.rs
@@ -5162,9 +5162,10 @@ thread_local! {
     /// launch latency. Canonicalization is a pure function of the (read-only,
     /// during prep) filesystem, so caching is safe within a pass.
     static ELF_CANON_CACHE: RefCell<HashMap<PathBuf, PathBuf>> = RefCell::new(HashMap::new());
-    /// Memoizes shared-library name resolution, keyed by `(soname, search_dirs)`
-    /// — the only inputs that determine the result.
-    static ELF_LIB_CACHE: RefCell<HashMap<(String, Vec<String>), PathBuf>> =
+    /// Memoizes shared-library name resolution, keyed by
+    /// `(soname, search_dirs, interpreter_dir)` — the only inputs that
+    /// determine the result.
+    static ELF_LIB_CACHE: RefCell<HashMap<(String, Vec<String>, Option<PathBuf>), PathBuf>> =
         RefCell::new(HashMap::new());
     /// Memoizes `parse_elf` (which reads the whole file) keyed by canonical path,
     /// so each shared object is read+parsed once per pass rather than once per
@@ -5223,12 +5224,49 @@ fn cached_file_id(canonical: &Path) -> Result<FileId> {
 fn elf_dependency_closure(binary: &Path) -> Result<Vec<PathBuf>> {
     let mut seen = HashSet::new();
     let mut result = Vec::new();
-    resolve_elf_recursive(binary, &mut seen, &mut result)?;
+    let interpreter_dir = interpreter_dir_of(binary)?;
+    resolve_elf_recursive(binary, interpreter_dir.as_deref(), &mut seen, &mut result)?;
     Ok(result)
 }
 
+/// Directory containing `binary`'s own ELF interpreter (`PT_INTERP`), if any.
+///
+/// A dynamic linker built by Nix (`ld-linux-x86-64.so.2` under a specific
+/// `glibc` store path) resolves `NEEDED` entries that carry no
+/// `DT_RPATH`/`DT_RUNPATH` of their own — e.g. `libgcc_s.so.1` depending on
+/// `libc.so.6` — by falling back to its own compiled-in default search path,
+/// which is the directory it itself lives in. Threading that directory
+/// through as an extra fallback mirrors what the real dynamic linker does at
+/// runtime, without hard-coding anything Nix-specific: on FHS systems it is
+/// already covered by the standard defaults below.
+///
+/// Verified directly: on a NixOS system, `libgcc_s.so.1` (from the
+/// `gcc-*-libgcc` package) carries an empty `DT_RUNPATH` and needs
+/// `libc.so.6` by bare soname; `ldd` still resolves it, via the interpreter
+/// named in the requesting binary's own `PT_INTERP`
+/// (`<glibc>/lib/ld-linux-x86-64.so.2`), to `libc.so.6` sitting right next to
+/// that same interpreter in `<glibc>/lib`.
+fn interpreter_dir_of(binary: &Path) -> Result<Option<PathBuf>> {
+    let canonical = cached_canonicalize(binary)?;
+    let parsed = parse_elf_cached(&canonical)?;
+    // `parsed.interpreter` is already canonicalized by `read_cstr_path`.
+    Ok(parsed
+        .interpreter
+        .and_then(|i| i.parent().map(Path::to_path_buf)))
+}
+
+/// Recursively walks `path`'s ELF dependency graph into `result`.
+///
+/// `interpreter_dir` is fixed for the whole closure — it comes from the
+/// top-level binary's own `PT_INTERP` (see [`interpreter_dir_of`]) and must
+/// be threaded through unchanged at every recursion depth, not recomputed
+/// per nested file: a real process has exactly one dynamic linker instance,
+/// so its default search path fallback applies uniformly to every
+/// transitively `NEEDED` library, not just the top-level binary's direct
+/// dependencies.
 fn resolve_elf_recursive(
     path: &Path,
+    interpreter_dir: Option<&Path>,
     seen: &mut HashSet<FileId>,
     result: &mut Vec<PathBuf>,
 ) -> Result<()> {
@@ -5239,11 +5277,12 @@ fn resolve_elf_recursive(
     result.push(canonical.clone());
     let parsed = parse_elf_cached(&canonical)?;
     if let Some(interpreter) = parsed.interpreter {
-        resolve_elf_recursive(&interpreter, seen, result)?;
+        resolve_elf_recursive(&interpreter, interpreter_dir, seen, result)?;
     }
     for needed in parsed.needed {
-        let dep = resolve_shared_library(&needed, &parsed.search_dirs, &canonical)?;
-        resolve_elf_recursive(&dep, seen, result)?;
+        let dep =
+            resolve_shared_library(&needed, &parsed.search_dirs, interpreter_dir, &canonical)?;
+        resolve_elf_recursive(&dep, interpreter_dir, seen, result)?;
     }
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


jocelynthode:
Hey I tried to build nono from this PR. It fixed the command_policies failing, but I hit then another bug. When having a `command_policies` defined  and trying to run for example `pi`. It fails. This seems related to the fact that pi is a wrapper shell script (common in nixpkgs)

any #!-script on `$PATH` fails ELF parsing in `parse_elf()`, the error propagates out of `add_outer_exec_file_with_deps`, and the entry is silently skipped (debug-level "outer exec gate skipped"). Neither the script nor its shebang interpreter is added to the execute allow-list, so launching it under any profile that has command policies fails with exit 127.

Not sure if fixing this should be part of this PR or a new issue.

SequeI:
Fixing both issues in one if the scope doesnt expand greatly would be very appreciated!

jocelynthode:
@Ekleog Is that something you would be willing to look into? I'm not well versed in Rust and would not feel comfortable vibe coding and submitting a PR that I would not thoroughly understand.

welociraptor:
Confirming this fixes #1649 on NixOS with nono 0.74.0 — the issue was filed against 0.73.0, so it's still current. The patch applies to tag v0.74.0 with offsets only and leaves Cargo.lock alone, so carrying it in nixpkgs is a one-line overrideAttrs.

Would you consider un-drafting it?

Ekleog:
I just undrafted, but unfortunately it looks like CI is failing and I'll probably not be able to get back to this for the next few weeks (maybe until early next year depending on how things evolve on my personal side).

If someone could take the PR over and push it to completion, it'd be awesome. Sorry for being unable to pass the finish line!

SequeI:
If I have free time tmwr, I will take care of this; will write it down on my todo's!

jocelynthode:
@SequeI Thanks for taking over! will you fix https://github.com/nolabs-ai/nono/issues/1724 in the same PR as well?

SequeI:
@jocelynthode Yup, just orienting myself around this right now. Will tag you once again to give it a whirl when the fix is up

SequeI:
@jocelynthode Let me know if both issues are resolved now off of this PR please

jocelynthode:
@SequeI I did a quick test running on this branch with
```
./target/debug/nono run --profile pi-cmd -- pi
```
pi worked fine. I also have command policies in this profile (for gh and kubectl) and both worked fine, so from my cursory test it appears to be fixed as these 2 tests where failing before! Thanks!

SequeI (approved):

PR #1629, opened 2026-08-12
fix(macos): refuse noninteractive proxy CA trust mutation

nickreese:
## Linked Issue

Closes #1627

## Summary

- reuse an already-trusted macOS proxy CA in noninteractive invocations
- refuse before `SecTrustSettingsSetTrustSettings` and before child launch when a first trust or re-trust would require unavailable terminal interaction
- preserve the existing interactive prompt, cancellation, and ephemeral-fallback behavior
- document the noninteractive contract for `--trust-proxy-ca`

## Agent Disclosure

I am an AI coding agent contributing at the repository owner's request. Before changing source, I reviewed `AGENTS.md`, `README.md`, `CONTRIBUTING.md`, `SECURITY.md`, `CLAUDE.md`, the affected `macos_trust.rs` and `proxy_runtime.rs` paths, and the related CLI documentation. Issue #1627 disclosed my identity, intent, approach, test plan, risks, and tradeoff before implementation.

No external implementation was copied or adapted. The change is newly written against the repository's existing CA lifecycle and `NonoError` patterns.

## Test Plan

- `cargo test -p nono-cli macos_trust` — 7 passed
- `cargo clippy --workspace --all-targets --all-features -- -D warnings -D clippy::unwrap_used` — passed
- `cargo fmt --all -- --check` — passed
- `cargo test -p nono` plus schema, manifest, and doctest stages — passed (748 core tests, 40 schema tests, 16 manifest tests, 7 doctests)
- `cargo test -p nono-cli` inside the aggregate `make ci` run — 1,909 passed, 1 ignored, and one unrelated parallel-only macOS temp-state test failed; that test passes alone at both untouched base `0a7852dc` and this head. A prior aggregate run had two such parallel-only failures, both of which also pass alone at base and head. Hosted CI is left to adjudicate the clean source in its ordinary environment.

No real Keychain mutation or persistent trust change was performed while testing.

## Checklist

- [x] An issue exists and is linked above
- [x] All commits are signed-off, using [DCO](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin)
- [x] All new code follows the project's coding standards ([CLAUDE.md](CLAUDE.md)) and is covered by tests
- [x] Public-facing changes are paired with documentation updates

## Agent Compliance Check

- [x] I am not prohibited from contributing under this policy
- [x] An issue already exists
- [x] I disclosed that I am an agent in the issue discussion
- [x] I described my intent and approach in the issue discussion
[...]

PR #1691, opened 2026-08-20
fix(macos): renew the trusted proxy CA before and during a session

### LINKED ISSUE #1681: TLS interception dies when the proxy CA expires mid-session, with no way to recover without a restart
### What happened?

When the interception CA's `not_after` passes, a running proxy stops being able to serve intercepted TLS, and stays that way for the rest of the process's life:

- Cached leaves are clamped to the CA's `not_after`, so they expire the moment it does — the client rejects them with `certificate has expired`.
- A newly-seen SNI can't be minted at all: `mint_leaf` fails with `CA certificate has expired`, no chain is resolved, and rustls aborts the handshake.
- The CA is only checked at startup, so nothing notices, and there's no path for a replacement CA to take effect in the running process. Every intercepted route is dead until the agent is restarted.

The handshake WARN also blames the client (`Agent likely pins certs or carries a hard-coded trust list. Remove endpoint_rules / credential_key from the route…`) when the cause is our own expired CA, which sends people the wrong way.

That the session recovers on its own — a fresh CA takes effect in the running proxy without a restart. How that happens is open; the missing piece is that today there is no mechanism for it at all.

### Steps to reproduce

1. Save the profile below as `/tmp/ca-expiry/profile.json` — two intercepted routes, `ca_validity: "120s"`.
2. `nono proxy --profile /tmp/ca-expiry/profile.json --port 18080 --no-auth --verbose` — note the printed trust-bundle path.
3. `curl --proxy http://127.0.0.1:18080 --cacert <bundle> https://example.com/` → 200, interception working, leaf cached.
4. Wait past 120s so the CA expires.
5. Re-request `https://example.com/` (cached leaf) → fails, expired certificate.
6. Request `https://example.net/`, not contacted before, so it forces a fresh mint → fails, no chain resolved.

`ca_validity: "120s"` only compresses the timeline; the same thing happens to a long-lived trusted CA whenever a session outlives its certificate.

### Relevant output or logs

```shell
$ openssl x509 -in <bundle> -noout -dates   # last cert in the bundle
notBefore=Aug 19 13:05:08 2026 GMT
notAfter=Aug 19 13:07:08 2026 GMT

=== t0 example.com (first contact, mints a leaf) ===  at 13:05:59Z
http_code=200

=== t+135 example.com (leaf cached pre-expiry) ===    at 13:08:14Z
curl: (60) SSL certificate problem: certificate has expired

=== t+135 example.net (new host, forces a fresh mint) === at 13:08:15Z
curl: (35) LibreSSL/3.3.6: error:1404B419:SSL routines:ST_CONNECT:tlsv1 alert access denied

# proxy, --verbose
13:05:08 INFO TLS interception active for 2 route(s); trust bundle at .../intercept-ca.pem
13:05:59 INFO l7 proxy response mode=connect_intercept target="example.com" method="GET" path="/" status=200
13:08:15 WARN tls_intercept: handshake failed for example.com:443 — received fatal alert: CertificateExpired. Agent likely pins certs or carries a hard-coded trust list. Remove endpoint_rules / credential_key from the route to fall back to a transparent CONNECT tunnel.
13:08:15 WARN tls_intercept: failed to mint leaf for SNI 'example.net': Configuration error: CA certificate has expired; cannot mint leaf for 'example.net'
13:08:15 WARN tls_intercept: handshake failed for example.net:443 — unexpected error: no server certificate chain resolved. Agent likely pins certs or carries a hard-coded trust list. Remove endpoint_rules / credential_key from the route to fall back to a transparent CONNECT tunnel.
```

### nono version

0.73.0

### Operating system

macOS

### OS version / distro

macOS 26.5.2 (arm64)

### Profile (if relevant)

```json
{
  "network": {
    "allow_domain": [
      { "domain": "example.com", "endpoints": [{ "method": "GET", "path": "/**" }] },
      { "domain": "example.net", "endpoints": [{ "method": "GET", "path": "/**" }] }
    ],
    "tls_intercept": { "ca_lifecycle": "session", "ca_validity": "120s" }
  }
}
```

kipz:
## Linked Issue

Closes #1681

## Summary

An expired interception CA killed every intercepted route until the agent restarted: `mint_leaf` refuses past `not_after`, cached leaves are clamped to it and expire with it, and the certificate was only ever checked at launch.

The root of it is that the key and the certificate were treated as one object — the expired branch discarded the private key and minted a whole new anchor, which is why renewal could only happen at launch. Renewal now re-issues over the key already in the Keychain, so subject, SPKI and the key-derived AKI are unchanged and leaves minted under the old certificate still chain to the new one.

That makes the live anchor swappable, so a running session can pick up a renewal without a restart. A supervisor prefers adopting a certificate another launch already renewed — no prompt and no trust-store write — and renews itself only when nothing newer exists, under a cross-process lock. Validity is tri-state, so launch rotates while the old certificate is still valid and a declined prompt costs nothing. If renewal cannot complete, the session keeps serving its current certificate and backs off.

Renewal and its trust-settings handling are macOS-only. The misleading "agent likely pins certs" WARN that triage flagged is not addressed here.

## Agent Disclosure

This PR was prepared by an AI coding agent acting on my behalf. Files and sections consulted: `AGENTS.md` (Coding Agent Contribution Policy, error-handling and forbidden-pattern rules), `CLAUDE.md`, `crates/nono-cli/src/macos_trust.rs`, `crates/nono-proxy/src/tls_intercept/*`, `crates/nono-cli/src/proxy_runtime.rs`. Intent, approach and limitations were disclosed in #1681 before implementation, and this PR matches that scope.

## Test Plan

`make ci` on this branch.

Unit tests cover re-issue producing the same SPKI with a later `not_after`, rejection of a mismatched key/cert pair, `replace_ca` clearing the leaf cache, the rotator swapping both the live certificate and the bundle, tri-state validity and headroom boundaries, and lock exclusivity across holders.

Verified live on macOS with a short `ca_validity` so a session outlives its certificate:

- Mid-session renewal without a restart — an already-expired CA (`no server certificate chain resolved` on every handshake), renewed in place, next request to a fresh host succeeded.
- Adoption — a second session minted a new anchor; the running session logged `Adopted renewed proxy CA from Keychain` with no prompt, rewrote its bundle, and continued serving freshly minted leaves.
- No CA key material reaches the on-disk bundle before or after a rotation.

## Checklist

## Agent Compliance Check

- [x] An issue already exists

### DIFF
--- a/crates/nono-cli/src/macos_ca_renewal.rs
+++ b/crates/nono-cli/src/macos_ca_renewal.rs
@@ -1,0 +1,273 @@
+//! Keeps a running session's interception CA usable across its own expiry, which
+//! would otherwise kill every intercepted route until a restart.
+//!
+//! Adopting a certificate another launch already renewed is preferred over renewing
+//! here, because only the latter writes to the trust store and prompts.
+
+use crate::macos_trust::{self, CertValidity};
+use nono_proxy::tls_intercept::InterceptCaRotator;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+use tracing::{debug, info, warn};
+
+/// Each check is a Keychain read, and a 7-day certificate needs nothing tighter.
+const MAX_CHECK_INTERVAL: Duration = Duration::from_secs(3600);
+
+/// Floor, so a misconfigured tiny `ca_validity` can't spin.
+const MIN_CHECK_INTERVAL: Duration = Duration::from_secs(5);
+
+/// After a declined prompt or failed trust write: re-prompting hourly would be
+/// worse than the expiry.
+const BACKOFF_AFTER_FAILURE: Duration = Duration::from_secs(6 * 3600);
+
+/// Watch the live interception CA and keep it renewed for the life of the process.
+///
+/// A plain thread, not a tokio task: every step is a blocking Security.framework
+/// call, and it must outlive the caller's stack frame.
+pub(crate) fn spawn_supervisor(rotator: Arc<InterceptCaRotator>, validity: Duration) {
+    let interval = check_interval(validity);
+    std::thread::Builder::new()
+        .name("nono-ca-renewal".to_string())
+        .spawn(move || {
+            let mut delay = interval;
+            loop {
+                std::thread::sleep(delay);
+                delay = match supervise_once(&rotator, validity) {
+                    Ok(Outcome::Deferred) => BACKOFF_AFTER_FAILURE,
+                    Ok(_) => interval,
+                    Err(e) => {
+                        warn!("Proxy CA renewal check failed: {e}");
+                        BACKOFF_AFTER_FAILURE
+                    }
+                };
+            }
+        })
+        .map(|_| ())
+        .unwrap_or_else(|e| warn!("Could not start proxy CA renewal supervisor: {e}"));
+}
+
+/// Several checks per headroom, so the renewal window is never missed however
+/// short the configured validity is.
+fn check_interval(validity: Duration) -> Duration {
+    let headroom = macos_trust::renewal_headroom(validity.as_secs().min(i64::MAX as u64) as i64);
+    (headroom / 2).clamp(MIN_CHECK_INTERVAL, MAX_CHECK_INTERVAL)
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum Outcome {
+    /// Live certificate has life left and nothing newer is stored.
+    NothingToDo,
+    /// Swapped in a certificate another process had already renewed.
+    Adopted,
+    /// Re-issued and installed a certificate ourselves.
+    Renewed,
+    /// Renewal was needed but couldn't be completed; still serving the old cert.
+    Deferred,
+}
+
+fn supervise_once(rotator: &InterceptCaRotator, validity: Duration) -> nono::Result<Outcome> {
+    let live = rotator
+        .current_cert_pem()
+        .map_err(|e| nono::NonoError::SandboxInit(format!("{e}")))?;
+
+    if let Some(outcome) = try_adopt(rotator, &live)? {
+        return Ok(outcome);
+    }
+    if macos_trust::cert_validity(&live)? == CertValidity::Valid {
+        return Ok(Outcome::NothingToDo);
+    }
+
+    // Renewal writes to the trust store and prompts; one winner per machine.
+    let Some(_lock) = RenewalLock::acquire()? else {
+        debug!("another process is renewing the proxy CA; deferring");
+        return Ok(Outcome::Deferred);
+    };
+    // It may have finished while we waited for the lock.
+    if let Some(outcome) = try_adopt(rotator, &live)? {
+        return Ok(outcome);
+    }
+
+    match macos_trust::renew_shared_ca(validity)? {
+        Some(renewed) => {
+            install(rotator, &renewed.key_der, &renewed.cert_pem)?;
+            info!("Proxy CA renewed mid-session; no restart needed");
+            Ok(Outcome::Renewed)
+        }
+        None => {
+            warn!(
+                "Proxy CA could not be renewed (authentication declined or trust write failed). \
+                 Still serving the current certificate; will retry later. Intercepted routes will \
+                 fail once it expires — restart the agent to recover."
+            );
+            Ok(Outcome::Deferred)
+        }
+    }
+}
+
+/// Swap in the stored CA if it supersedes what we're serving.
+fn try_adopt(rotator: &InterceptCaRotator, live: &str) -> nono::Result<Option<Outcome>> {
+    let Some((key_der, cert_pem)) = macos_trust::read_shared_ca()? else {
+        return Ok(None);
+    };
+    if !worth_adopting(&cert_pem, live)? {
+        return Ok(None);
+    }
+    if !macos_trust::stored_cert_is_trusted(&cert_pem)? {
+        debug!("stored proxy CA is newer but not trusted; not adopting");
+        return Ok(None);
+    }
+    install(rotator, &key_der, &cert_pem)?;
+    info!("Adopted renewed proxy CA from Keychain");
+    Ok(Some(Outcome::Adopted))
+}
+
+fn worth_adopting(stored: &str, live: &str) -> nono::Result<bool> {
+    Ok(stored != live && macos_trust::supersedes(stored, live)?)
+}
+
+fn install(rotator: &InterceptCaRotator, key_der: &[u8], cert_pem: &str) -> nono::Result<PathBuf> {
+    rotator
+        .rotate(key_der, cert_pem)
+        .map_err(|e| nono::NonoError::SandboxInit(format!("failed to install renewed CA: {e}")))
+}
+
+/// Advisory lock so two sessions renewing at once produce one prompt, and one
+/// Keychain write, rather than two.
+struct RenewalLock {
+    _lock: nix::fcntl::Flock<std::fs::File>,
+}
+
+impl RenewalLock {
+    fn acquire() -> nono::Result<Option<Self>> {
+        let dir = crate::state_paths::user_state_dir()?;
+        std::fs::create_dir_all(&dir).map_err(|e| {
+            nono::NonoError::SandboxInit(format!(
+                "cannot create state dir '{}': {e}",
+                dir.display()
+            ))
+        })?;
+        let path = dir.join("proxy-ca-renewal.lock");
+        let file = std::fs::OpenOptions::new()
+            .create(true)
+            .truncate(false)
+            .write(true)
+            .open(&path)
+            .map_err(|e| {
+                nono::NonoError::SandboxInit(format!(
+                    "cannot open renewal lock '{}': {e}",
+                    path.display()
+                ))
+            })?;
+        match nix::fcntl::Flock::lock(file, nix::fcntl::FlockArg::LockExclusiveNonblock) {
+            Ok(lock) => Ok(Some(Self { _lock: lock })),
+            Err((_, nix::errno::Errno::EWOULDBLOCK)) => Ok(None),
+            Err((_, e)) => Err(nono::NonoError::SandboxInit(format!(
+                "cannot lock '{}': {e}",
+                path.display()
+            ))),
+        }
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+mod tests {
+    use super::*;
+    use nono_proxy::tls_intercept::{CertCache, EphemeralCa};
+
+    fn rotator_with(ca: EphemeralCa, dir: &std::path::Path) -> InterceptCaRotator {
+        InterceptCaRotator::new(
+            Arc::new(CertCache::new(Arc::new(ca))),
+            dir.to_path_buf(),
+            "intercept-ca.pem",
+            None,
+        )
+    }
+
+    #[test]
+    fn the_check_interval_fits_inside_the_renewal_headroom() {
+        // 7 days: 24h headroom, so hourly is the cap that bites.
+        assert_eq!(
+            check_interval(Duration::from_secs(7 * 86400)),
+            MAX_CHECK_INTERVAL
+        );
+        // 120s: 30s headroom, so checks must be far tighter than hourly or the
+        // window closes unnoticed.
+        assert_eq!(
+            check_interval(Duration::from_secs(120)),
+            Duration::from_secs(15)
+        );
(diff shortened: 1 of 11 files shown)


Frankie-Xu (commented):
This review is from an AI coding agent (Cursor) on behalf of @Frankie-Xu.

Re-issuing over the Keychain key (stable SPKI) plus adopt-before-renew is the right shape for this bug: a running `--trust-proxy-ca` session can survive CA expiry without minting a second trust anchor or prompting twice. Unit tests around re-issue, rotation, tri-state validity, and the flock look solid.

**Blocking — `crates/nono-cli/src/macos_ca_renewal.rs:36-44` and `:82-84`.** Lock contention and a declined prompt share `Outcome::Deferred`, which always becomes `BACKOFF_AFTER_FAILURE` (6h). If session B loses `RenewalLock` to session A, B sleeps six hours instead of retrying on `check_interval` and adopting the certificate A just wrote. That undercuts the “adopt with no prompt” path the PR description is selling. Please split outcomes (`Busy` vs declined/failed trust) so lock contention retries on the normal interval.

**Blocking — `crates/nono-proxy/src/tls_intercept/rotate.rs:52-60`.** `replace_ca` runs before `write_bundle`. `write_bundle` can still fail after that (native-roots load, IO). The live interceptor then serves a CA that `SSL_CERT_FILE` does not contain. Fail-secure order is: validate the pair → atomically replace the bundle → swap the live CA only if the write succeeded. A test that injects a bundle-write failure and asserts `current_cert_pem()` is unchanged would lock this in.

**Blocking — `crates/nono-cli/src/macos_ca_renewal.rs:125-127` and `crates/nono-cli/src/macos_trust.rs:116-118`.** `worth_adopting` / `supersedes` only compare PEM inequality and `not_after`. A later-expiring Keychain cert with a **different** SPKI (the `reissue_with_cn` failure path that falls through to `generate_and_trust_new_ca` in `macos_trust.rs:164-170`) would be adopted into a process whose child already loaded the previous anchor. Same-key re-issue is the invariant that makes mid-session rotation safe; please require matching SPKI (or refuse adopt and leave the live CA alone until restart). `only_a_longer_lived_stored_cert_is_worth_adopting` currently only covers the same-key case.

**Nit — `crates/nono-cli/src/macos_trust.rs:471-478`.** `killall trustd` on every successful rotation is a surprising side effect (other local TLS clients). Worth a one-line comment in the user-facing docs, and consider scoping to the current user. `file.flush().ok()` in `crates/nono-proxy/src/tls_intercept/bundle.rs:173` also swallows a flush error before `rename`.

**Nit — tests.** Happy-path coverage is good; the two blocking cases above (lock-contention backoff, bundle-write failure, different-SPKI adopt) are the gaps. The test plan says CA key material never reaches the on-disk bundle — an `assert!(!bundle.contains("PRIVATE KEY"))` next to `rotate_installs_the_new_cert_and_rewrites_the_bundle` would match that claim. Unrelated to this PR’s scope, but `#1629` (refuse noninteractive trust mutation) will interact with the mid-session `trust_cert` prompt; `ERR_SEC_INTERACTION_NOT_ALLOWED` already maps to `UserCancelled`, which is fine as long as that stays a 6h deferred retry rather than a tight prompt loop.

I ran locally: `cargo test -p nono-proxy --lib -- tls_intercept::{ca,rotate,cert_cache,bundle}` (30 passed) and `cargo test -p nono-cli --bin nono -- macos_ca_renewal macos_trust` (13 passed) on macOS.

kipz:
Addressed all three blocking points from @Frankie-Xu's review (a04401a):

1. **Lock contention vs. declined/failed trust** — split into `Outcome::Busy` (lost the lock race) and `Outcome::Deferred` (declined prompt / failed trust write). Only `Deferred` triggers the 6h backoff; `Busy` now retries on the normal `check_interval`, so a session that lost the race adopts the winner's certificate on its next check. See `next_delay` in `macos_ca_renewal.rs`, covered by `losing_the_lock_race_retries_on_the_normal_interval_not_backoff` and `a_declined_or_failed_trust_attempt_backs_off`.

2. **Bundle-write ordering** — `rotate()` now writes the bundle before swapping the live CA, so a `write_bundle` failure leaves `current_cert_pem()` untouched instead of serving a CA the on-disk bundle doesn't contain. Added `rotate_leaves_the_live_ca_untouched_if_the_bundle_write_fails`, which points the bundle dir at a path with no writable parent and asserts the live CA is unchanged after the error.

3. **SPKI matching in `worth_adopting`/`supersedes`** — `supersedes` now requires the candidate and current cert to share the same SPKI before comparing `not_after`, so a different-key cert (the `generate_and_trust_new_ca` fallback path) is never adopted into a running session. Added `supersedes_requires_matching_spki`.

Also picked up both nits: `rotate.rs`'s bundle write now propagates a flush error instead of swallowing it (`bundle.rs`), and `flush_trust_cache` has a short comment noting unprivileged `killall` only reaches the current user's `trustd`.

`make ci` is clean on the updated branch.

panga:
@kipz good work, this is a meaningful improvement over the current behavior. I do have a few technical concerns:

- It restarts `trustd` on every renewal. That introduces a brief interruption to system TLS validation for all user traffic, not just `nono`. Can we avoid it? If we proceed with this approach, the documentation should call this out prominently. I am also not confident this will behave reliably in constrained enterprise environments where users are restricted from terminating certain processes.

- An authentication prompt can now appear mid-session rather than only at launch. That is problematic for the background `nono proxy`, which may surface a prompt when no user is present to respond. In that case, the resulting error path should be explicit and actionable.

kipz:
Thanks for the review, both points are fair. Here's what I found.

**trustd restart**: `killall -q trustd` here is unprivileged, so it only restarts the invoking user's own trustd (see the comment at `macos_trust.rs:489-493`). It doesn't touch other users or the system. I checked for a supported way to invalidate trustd's cache without restarting it, and there isn't one on current macOS. Apple's only supported route for silent trust deployment is an MDM root-certificate payload, which doesn't fit a per-user opt-in flag like `--trust-proxy-ca`. The flush is also best-effort: if `killall` fails, we `debug!`-log it and move on, trustd picks up the new anchor on its own eventually. So a locked-down environment gets a delay, not a failure. I don't see a way to avoid the restart. Let me know if you know of one.

**Mid-session prompt with no user present**: `errSecInteractionNotAllowed` maps to the same `Deferred` outcome as a declined prompt. That backs off 6h and logs a warning instead of hanging, and the whole check runs on its own thread so it can't block the proxy. The real gap is visibility: for a headless `nono proxy` daemon there's nothing beyond that log line today. Rather than build a notification mechanism, I'd propose a flag for `nono proxy` to opt out of mid-session renewal and keep launch-time renewal only, accepting that a long-lived headless proxy eventually needs a restart. Would that address it, or did you have something else in mind?

panga:
`trustd` restart: the main question is whether cache invalidation is actually needed. Based on the agent review, it probably shouldn’t be necessary, though that may still need validation. If a restart is truly required, that needs to be clearly documented.

Mid-session prompt with no user present: agreed, the expected behavior there is a fast-fail deny.

kipz:
Thanks @panga.

**`trustd` restart:** confirmed. Trust changes (add/remove via `security`) took effect on the very next TLS handshake without restarting `trustd`. Removed `flush_trust_cache()`.

**Mid-session prompt with no user present:** `errSecInteractionNotAllowed` is now classified separately from a declined prompt and treated as terminal. It logs once at error level and stops retrying for the session, rather than backing off and re-prompting every 6h.

145393bb

Frankie-Xu (commented):
This review is from an AI coding agent (Cursor) on behalf of @Frankie-Xu.

Follow-up on `ab22d495` → HEAD `04fdeb61`. All three previous Blocking items are absorbed and covered by tests: `Outcome::Busy` vs `Deferred` (`macos_ca_renewal.rs` `next_delay`, `losing_the_lock_race_retries_on_the_normal_interval_not_backoff`), `write_bundle` before `replace_ca` (`rotate.rs:55-63`, `rotate_leaves_the_live_ca_untouched_if_the_bundle_write_fails`), and `supersedes` requiring matching SPKI (`macos_trust.rs:144-148`, `supersedes_requires_matching_spki`). `killall trustd` is gone; bundle flush errors now propagate.

The remaining issues are in `04fdeb61`'s `NoInteraction` handling.

**Blocking — `crates/nono-cli/src/macos_trust.rs:65-68` and `:261-266`.** `UserCancelled` while `RenewDue` returns `RotateOutcome::Cert` with the old still-valid CA. `NoInteractionAvailable` always returns `RotateOutcome::NoInteraction`, and launch maps that to `Ok(None)` — same as `Unavailable`. The log says "Continuing on the current certificate," but `try_ensure_trusted_ca` then falls back to an ephemeral CA. A headless start during headroom therefore drops a trusted, still-valid Keychain anchor and serves an untrusted one; Go clients that would have kept working until `not_after` fail immediately. Please keep the stored CA when `state == CertValidity::RenewDue`, and only ephemeral-fallback when it is already `Expired`.

**Blocking — `crates/nono-cli/src/macos_ca_renewal.rs:36-40` and `:80-83`.** `next_delay(NoInteraction) == None` exits the supervisor for good, so `try_adopt` never runs again. Prompting cannot succeed later in this process; adoption still can — another interactive `--trust-proxy-ca` session can re-issue into Keychain with no prompt required here. Stopping the loop undercuts the adopt-before-renew path. Please keep checking on `check_interval` with prompting disabled (adopt only), rather than tearing the supervisor down.

**Blocking — `crates/nono-cli/src/proxy_command.rs:75-128`.** `04fdeb61` is written for a headless `nono proxy`, but `run_proxy` still never calls `spawn_supervisor` (only `start_proxy_runtime` does). The new terminal outcome and mid-session adopt/renew never run for the command that discussion was about. A long-lived standalone proxy still dies on CA expiry until restart; the launch-time mapping above is the only new headless behavior that actually fires.

**Nit — `docs/cli/usage/flags.mdx:540`.** The mid-session bullet still reads as "keep using the current certificate rather than re-prompting." It does not mention that a headless session stops retrying, or that launch can ephemeral-fallback. Once the two launch/supervisor behaviors above are settled, one sentence here would match.

I ran locally on `04fdeb61`: `cargo test -p nono-proxy --lib -- tls_intercept` (112 passed) and `cargo test -p nono-cli --bin nono -- macos_ca_renewal macos_trust` (17 passed) on macOS.

kipz:
Rebased onto latest `main` and addressed the outstanding review findings:

1. **Launch path dropped a still-valid CA to ephemeral on a headless session.** `RotateOutcome::NoInteraction` now carries an optional `fallback: Option<PreloadedCa>`, set when the stored cert is merely due for renewal (not expired) — mirroring the existing `UserCancelled` branch. `try_ensure_trusted_ca` returns that fallback instead of `None`, so a headless launch keeps serving the still-valid trusted cert instead of silently downgrading to an untrusted ephemeral one. `renew_shared_ca` (mid-session path) still maps to `RenewOutcome::NoInteraction` regardless of the fallback, so it doesn't get misrouted through `Unchanged`/`Deferred`'s "declined or trust write failed" backoff logging.

2. **Supervisor exited its polling loop for good on `NoInteraction`.** It now keeps polling at the normal interval instead of stopping. A `renewal_disabled` flag (sticky once a `NoInteraction` outcome is observed) skips the `RenewalLock`/`renew_shared_ca` attempt on later iterations, avoiding a pointless repeated prompt attempt, but `try_adopt` still runs every interval — so a headless session can adopt a cert renewed by another, interactive session instead of being stuck until restart. Updated `next_delay`'s signature and the test that had locked in the old stop-for-good behavior.

3. **Standalone `nono proxy` never got mid-session renewal.** `run_proxy` never called `spawn_supervisor`, unlike the sandboxed path's `start_proxy_runtime`. Added the same post-handle supervisor spawn (gated on `trust_proxy_ca` and an available rotator) to `run_proxy`, since it blocks until Ctrl-C the same way a sandboxed session does.

`make ci` (clippy, fmt, full test suite) passes on the rebased branch.

kipz:
@panga following up on the fast-fail deny point.

I initially implemented that (145393bb: `errSecInteractionNotAllowed` classified as terminal, stops retrying for the session). Frankie-Xu's follow-up review flagged that as blocking though: a supervisor that stops for good on `NoInteraction` can never later adopt a certificate that a *different*, interactive session renews via the shared Keychain key. Since adopt-before-renew (no prompt) is the mechanism this PR relies on to avoid re-prompting, permanently disabling it defeats that path for the lifetime of the headless process.

What's pushed now: the supervisor still stops *attempting to renew* (no repeated prompt attempts, matching your fast-fail intent for the interactive path) but keeps polling on the normal interval to *adopt* a cert another session renews. So a headless `nono proxy` never re-prompts, but isn't stuck on its original cert until restart either.

panga:
Thank you @kipz it makes sense to me.

PR #1692, opened 2026-08-20
feat(profile,proxy): glob patterns for env var and hostname allow/deny lists

### LINKED ISSUE #1642: Support whole-label wildcards in non-leading hostname positions

Nono can currently express exact hosts and leading suffix wildcards, but it cannot express a constrained family of hosts with fixed labels on both sides of a variable label.

For example, I need to express this host pattern:

```text
jenkins.*.ci.example.com
```

It should match these hosts:

```text
jenkins.prod.ci.example.com
jenkins.stage.ci.example.com
```

It should not match these hosts:

```text
jenkins.ci.example.com
jenkins.foo.bar.ci.example.com
other.prod.ci.example.com
```

The closest currently supported pattern is broader.

```text
*.ci.example.com
```

In Nono 0.70.0, leading wildcard patterns use suffix matching. Broadening the rule this way permits unrelated sibling services and deeper hostnames under the suffix.

This is particularly important for credential injection. The intended hostname restriction needs to be enforced by the same Nono matcher that controls network access and selects the `custom_credentials` route.

A narrower check inside a `credential_capture` command cannot restore the intended network boundary after Nono has already allowed the broader hostname and selected a route for it.

Support `*` as a complete hostname label in a non-leading position.

For example:

```text
jenkins.*.ci.example.com
```

A non-leading wildcard label should match exactly one non-empty label.

Suggested matching behavior:

```text
jenkins.prod.ci.example.com          match
jenkins.stage.ci.example.com         match
jenkins.ci.example.com               no match
jenkins.foo.bar.ci.example.com       no match
other.prod.ci.example.com            no match
```

Matching should remain case-insensitive.

At minimum, the same hostname-pattern implementation should be used for:

- `network.allow_domain`
- `network.custom_credentials[*].upstream`
- route protection and credential-route selection
- any corresponding policy checks that consume the same hostname grammar

Using one shared parser and matcher would ensure that network authorization and credential scoping cannot disagree.

I am not proposing a change to the existing leading `*.` suffix semantics. This request is specifically for a wildcard occupying a complete non-leading hostname label. Partial-label globs such as `jenkins-*.example.com` can remain invalid unless they are separately designed and documented.

One option is to enumerate every exact hostname. This preserves the security boundary, but it becomes brittle when environment labels are added or removed.

Another workaround is to broaden the Nono rule.

```text
jenkins.*.ci.example.com
```

becomes:

```text
*.ci.example.com
```

A host-side `credential_capture` command can then inspect `NONO_REQUEST_HOST` and refuse to return credentials unless the original narrower pattern matches.

This is not an adequate replacement for native support because:

- Nono’s network allowlist still operates on the broader pattern
- credential-route selection still operates on the broader pattern
- the external check only protects credential material
- the external check cannot prevent connectivity to unintended sibling hosts
- the external matcher becomes a second security-critical hostname implementation that can drift from Nono’s behavior

Native support would keep network authorization, route selection, and credential scoping under one authoritative matcher.

### How is this blocking you?

### Additional context

Relevant context:
- `nono version` output is `nono 0.70.0`
- Operating system is macOS
- The relevant v0.70.0 network matcher only treats hosts beginning with `*.` as wildcard patterns
- The relevant v0.70.0 credential-route matcher also only handles exact hosts and leading `*.` patterns
- This is distinct from #1242, which concerned interception for an already-supported leading wildcard route

Relevant source:
- [`HostFilter` wildcard handling](https://github.com/nolabs-ai/nono/blob/v0.70.0/crates/nono/src/net_filter.rs#L121-L135)
- [`HostFilter` suffix matching](https://github.com/nolabs-ai/nono/blob/v0.70.0/crates/nono/src/net_filter.rs#L227-L235)
- [`host_port_matches` credential-route matching](https://github.com/nolabs-ai/nono/blob/v0.70.0/crates/nono-proxy/src/route.rs#L646-L669)

### LINKED ISSUE #1663: Support full-name glob matching in `environment.allow_vars` and `environment.deny_vars`

Nono’s environment-variable policy currently supports exact names and trailing-`*` prefix patterns such as:

```jsonc
"environment": {
  "allow_vars": ["PATH", "HOME", "AWS_*"],
  "deny_vars": ["GITHUB_*"]
}
```

This cannot express suffix, contains, or infix patterns such as:

```text
*_TOKEN
*SECRET*
AWS_*_TOKEN
```

A concrete use case is allowing most of a developer’s host environment while removing variables whose names look like credentials before the sandbox starts.

For example, in our internally we create a custom Go prefilter that performs case-insensitive substring matching for secret-like environment-variable names before invoking Nono:

That filtering cannot currently be represented in the Nono profile itself. As a result:

* The complete environment security policy is split between the wrapper and Nono.
* `nono profile validate` cannot validate or explain the complete policy.
* Other Nono consumers need to implement similar filtering independently.
* The external matcher becomes security-sensitive code that can drift from Nono’s execution behavior.

The relevant Nono execution configuration currently documents environment patterns as exact names or prefix patterns:

https://github.com/nolabs-ai/nono/blob/main/crates/nono-cli/src/exec_strategy.rs#L2707-L2728

Support anchored glob matching over the complete environment-variable name, with `*` allowed in leading, trailing, and infix positions.

Suggested matching behavior:

```text
Pattern             Variable                    Result
GH_TOKEN            GH_TOKEN                    match
GH_*                GH_TOKEN                    match
*_TOKEN             GH_TOKEN                    match
*TOKEN*             JENKINS_TOKEN               match
AWS_*_TOKEN         AWS_SESSION_TOKEN           match
*_TOKEN             GH_SECRET                   no match
```

A bare `*` should continue to match every variable name.

Existing exact and trailing-`*` patterns should retain their current behavior, making the change backward compatible.

For example, a profile could express:

```jsonc
"environment": {
  "allow_vars": ["*"],
  "deny_vars": [
    "*TOKEN*",
    "*KEY*",
    "*PASS*",
    "*SECRET*",
    "*CREDENTIAL*"
  ]
}
```

The use case also needs an optional case-insensitive matching mode. The current wrapper filter normalizes names before matching so that variants such as these cannot bypass the policy:

```text
JENKINS_TOKEN
jenkins_token
Jenkins_Token
```

The exact configuration schema is flexible. This could be an environment-level option, structured pattern rules, or another representation that preserves existing string-pattern compatibility.

For example:

```jsonc
"environment": {
  "allow_vars": ["*"],
  "deny_vars": [
    "*TOKEN*",
    "*KEY*",
    "*PASS*",
    "*SECRET*",
    "*CREDENTIAL*"
  ],
  "pattern_case": "insensitive"
}
```

### Expected policy semantics

The generalized matcher should preserve the existing environment-policy ordering:

1. Host environment variables are evaluated against the built-in dangerous-variable rules.
2. `deny_vars` takes precedence over `allow_vars`.
3. Explicit `environment.set_vars` values are added after host filtering.
4. Nono-generated credential and proxy variables are added after host filtering and continue to bypass `allow_vars` and `deny_vars`.

The last property is important for credential injection. A profile should be able to reject a real host-provided `GH_TOKEN` or `JENKINS_TOKEN` while still allowing Nono to provide a phantom variable with the same name.

Current execution ordering can be seen here:

https://github.com/nolabs-ai/nono/blob/main/crates/nono-cli/src/exec_strategy.rs#L2918-L2959

The patterns should match variable names only. They should never inspect or match variable values.

The same matcher should be used wherever Nono applies profile environment policy, including:

* Direct execution
* Supervised execution
* Tool-sandbox or command-policy children
* Profile validation
* Profile explanation and diagnostics

#### Enumerating exact variable names

A profile can list known names such as:

```jsonc
"deny_vars": [
  "GH_TOKEN",
  "GITHUB_TOKEN",
  "NPM_TOKEN",
  "JENKINS_TOKEN",
  "OPENAI_API_KEY"
]
```

This is brittle. New tools and vendors introduce new credential names, and each profile must continually update its list.

#### Using prefix patterns

Prefix matching works for a known namespace such as `AWS_*`, but it cannot express common cross-tool conventions such as `*_TOKEN` or `*_API_KEY`.

Using broad prefixes would also remove unrelated non-secret variables.

#### Filtering before invoking Nono

This is the current wrapper workaround.

It works for inherited process environment variables, but it means the effective policy is not fully represented by the Nono profile. It also creates a second security-critical matcher outside Nono.

Native support would allow the wrapper to remove this custom launch-time filtering and make the Nono profile the authoritative description of the environment boundary.

#### Expanding Nono’s built-in dangerous-variable list

Making every variable containing words such as `TOKEN` or `KEY` globally dangerous may be too opinionated and could break existing users.

Profile-configurable patterns would let consumers choose the appropriate policy without changing Nono’s defaults.

### How is this blocking you?

### How is this blocking you?

This is not a hard blocker because the external prefilter works.

It prevents us from:

* Removing custom security-sensitive environment-filtering code.
* Expressing the complete sandbox environment policy in Nono.
* Validating the complete policy through Nono tooling.
* Reusing the same policy consistently across profiles and execution strategies.

### Scope and non-goals

This request concerns environment variables inherited when Nono constructs the sandboxed child environment.

It does not ask Nono to monitor or rewrite arbitrary environment mutations made by a process after `exec`, such as a running interactive shell exporting a new variable from `.zshrc`, `.bashrc`, direnv, or another shell hook. That is a separate lifecycle problem.

This request also does not propose changing the hostname-pattern grammar or reusing hostname glob semantics for environment variables.

### Acceptance criteria

* `*` may appear at the beginning, end, or middle of an environment-variable pattern.
* Matching is anchored to the complete variable name.
* Existing exact, bare-`*`, and trailing-`*` behavior remains compatible.
* A profile can opt into case-insensitive matching.
* `deny_vars` continues to override `allow_vars`.
* `set_vars` and Nono-injected credential variables retain their current post-filter behavior.
* The same matching implementation is used by all execution strategies.
* Invalid patterns fail during profile validation with a clear error.
* Documentation describes the matching grammar and precedence.
* Tests cover exact, prefix, suffix, contains, infix, case-sensitive, and case-insensitive matching.

### Related issue

This is conceptually related https://github.com/nolabs-ai/nono/issues/1642

That issue requests richer wildcard semantics for hostname matching and argues that network authorization and credential routing should use one authoritative native matcher.

This request applies the same architectural principle to environment-variable names: environment ingress policy should be expressible and enforceable inside Nono rather than requiring an external security-critical matcher.

The policy domains and matching grammars are otherwise separate.

SequeI:
## Linked Issue

Closes #1663
Closes #1642 

@btiernay Tested manually, and all your usecases worked fine for me, let me know whenever you have time if this is all good on your end too if possible. Thanks!

## Summary

Generalize two pattern matchers that previously only supported exact
names or a trailing-* prefix, closing #1663 and #1642.

environment.allow_vars/deny_vars now accept * anywhere in the name
(leading, trailing, or infix: "*_TOKEN", "*SECRET*", "AWS_*_TOKEN"),
with an optional case_insensitive_vars flag. The same relaxed grammar
now applies to export_env/session_export_env, since they share the
same matcher. case_insensitive_vars is sticky across `extends` so a
child profile can't silently downgrade it.

network.allow_domain/deny_domain now support a whole-label wildcard in
a non-leading position ("jenkins.*.ci.example.com"), matching exactly
one label there. The matcher lives in nono::net_filter and is shared
by allow/deny_domain and nono-proxy's route/credential matching, so
network authorization and credential routing can't disagree.

## Checklist


btiernay:
Thanks for this change set @SequeI! Some observations below:

The centralization is mostly sound but there are 3 inconsistencies I see:
- exact environment patterns currently match any value sharing their prefix, so `PATH` matches `PATH_EXTRA` 
- `HostFilter` and `RouteStore` still normalize hostnames differently, allowing an equivalent hostname such as a trailing dot FQDN to pass authorization while missing its protected route and falling through to transparent `CONNECT`.
- command policy child environments and export fields still reject multiple wildcards, so the documented `*SECRET*` grammar is not actually shared across execution strategies. 

I would probably also make invalid patterns fail during preparation and update the env documentation before merge. Thoughts?

crates/nono-cli/src/exec_strategy/env_sanitization.rs
```diff
@@ -416,10 +445,43 @@ mod tests {
     }
 
     #[test]
-    fn test_env_var_allowed_mid_star_ignored() {
+    fn test_env_var_allowed_mid_star_matches_infix_wildcard() {
```
lukehinds, line 461: While you're adding tests here - I think the exact-match bug above slipped past CI because nothing asserts that an exact pattern *doesn't* match a superstring (`test_env_var_denied_no_match` only uses unrelated names). Could you add a couple of regression cases so it can't come back?

```rust
assert!(!is_env_var_allowed("PATH_X", &["PATH".into()]));
assert!(!is_env_var_denied("GH_TOKEN_BAK", &["GH_TOKEN".into()]));
```

lukehinds (commented):
sorry was testing an agent flow, I will have a review humane style shortly.

SequeI:
Thanks for the reviews, and catching the inconsistencies! All should be fixed up now and standardized

cong-or (approved):

connrg (approved):
lgtm

PR #1722, opened 2026-08-23
fix(sandbox): grant Refer on command_policies outer exec gate

### LINKED ISSUE #1689: command_policies in profile causes same-filesystem rename to fail with EXDEV
### What happened?

A cross-directory rename beneath an explicit read/write `/tmp` grant succeeds with the default sandbox but fails with `EXDEV` when a minimal profile containing `command_policies` is enabled.

Both paths are beneath the same `/tmp` grant and on the same filesystem. The profile does not add any filesystem restrictions affecting the Node process; it only defines one minimal command policy for `git`.

Enabling `command_policies` should not remove Landlock `REFER` permission from filesystem operations already permitted by the outer sandbox. The rename should succeed with or without the profile.

### Steps to reproduce

1. Create `~/.config/nono/profiles/refer-repro.json` using the profile from the Profile field below.

2. Run this baseline command without the profile:

   ```
   nono run --allow /tmp -- /home/linuxbrew/.linuxbrew/opt/node/bin/node -e 'const fs = require("node:fs"); const base =
   `/tmp/nono-refer-test-${process.pid}`; try { fs.mkdirSync(`${base}/pending`, { recursive: true });
   fs.writeFileSync(`${base}/pending/result.json`, "{}"); fs.renameSync(`${base}/pending/result.json`, `${base}/result.json`);
   console.log("rename succeeded"); } finally { fs.rmSync(base, { recursive: true, force: true }); }'
   ```

   It prints `rename succeeded`.

3. Run the same command with the minimal command-policy profile:

   ```
   nono run --profile refer-repro --allow /tmp -- /home/linuxbrew/.linuxbrew/opt/node/bin/node -e 'const fs =
   require("node:fs"); const base = `/tmp/nono-refer-test-${process.pid}`; try { fs.mkdirSync(`${base}/pending`, { recursive: true
   }); fs.writeFileSync(`${base}/pending/result.json`, "{}"); fs.renameSync(`${base}/pending/result.json`, `${base}/result.json`);
   console.log("rename succeeded"); } finally { fs.rmSync(base, { recursive: true, force: true }); }'
   ```

4. The second command fails with `EXDEV`.

### Relevant output or logs

```shell
kernel  Landlock V4
        degraded: IoctlDev, Scoping (upgrade kernel for full support)
Applying sandbox...

node:fs:1248
  binding.rename(
          ^

Error: EXDEV: cross-device link not permitted, rename '/tmp/nono-refer-test-1547140/pending/result.json' ->
'/tmp/nono-refer-test-1547140/result.json'
    at Object.renameSync (node:fs:1248:11)
    at [eval]:3:60
    at runScriptInThisContext (node:internal/vm:219:10)
    at node:internal/process/execution:483:12
    at [eval]-wrapper:6:24
    at runScriptInContext (node:internal/process/execution:481:60)
    at evalFunction (node:internal/process/execution:315:30)
    at evalTypeScript (node:internal/process/execution:327:3)
    at node:internal/main/eval_string:71:3 {
  errno: -18,
  code: 'EXDEV',
  syscall: 'rename',
  path: '/tmp/nono-refer-test-1547140/pending/result.json',
  dest: '/tmp/nono-refer-test-1547140/result.json'
}
```

### nono version

0.74.0

### Operating system

Linux

### OS version / distro

Ubuntu 24.04.4 LTS (Noble Numbat)

6.8.0-134-generic

### Profile (if relevant)

```json
{
  "extends": "default",
  "meta": {
    "name": "refer-repro",
    "version": "1.0.0"
  },
  "command_policies": {
    "commands": {
      "git": {
        "from": {
          "session": {
            "fs_read": [
              "/tmp"
            ]
          }
        }
      }
    }
  }
}
```

### Additional context

The command succeeds without `--profile refer-repro` and fails when the minimal profile above is enabled, suggesting the presence of `command_policies` to be behind the issue.

Both rename paths are beneath the explicit `--allow /tmp` grant and are on the same filesystem.

This appears related to #1396 and its fix #1397, but still reproduces on nono 0.74.0. The existing fix may not cover every
Landlock layer activated during Tool Sandbox command-policy preparation.

The existing regression appears to cover moving a file into a newly created subdirectory. This reproduction moves a file from a
newly created child directory to its parent:

    /tmp/test/pending/result.json -> /tmp/test/result.json

SebTardif:
## Linked Issue

Closes #1689

## Summary

`command_policies` stacks a Landlock execute-restriction layer that did not grant `Refer`. Same-filesystem rename then failed with `EXDEV` even when the outer sandbox already allowed the paths. This grants bare `Refer` on `/` in `apply_outer_exec_gate`, matching `restrict_execute`.

## Problem

A same-filesystem rename such as `pending/result.json` to `result.json` succeeds under the default sandbox and fails with `EXDEV` once a profile includes `command_policies`. Both paths sit under an explicit write grant. The profile does not add filesystem restrictions that should affect that rename.

## Cause

`apply_outer_exec_gate` created a stacked Landlock ruleset that only called `handle_access(AccessFs::Execute)`. Landlock requires `Refer` in every stacked layer for rename and link. Missing it turns an already-permitted same-FS rename into `EXDEV`. PR #1397 fixed the same class of bug in `restrict_execute`; the tool-sandbox outer exec gate was not updated.

## Change

- Handle `AccessFs::Refer` at `BestEffort` and grant it on `/` when the ABI supports it.
- Bare `Refer` does not widen write or execute rights.
- Add a `command_policies` suite test that creates `pending/result.json` and renames it to `result.json`.
- Add a stacked-gate unit test in the same style as the #1397 rename regression.

## Test Plan

- `cargo fmt --check`
- `cargo clippy -p nono-cli --all-targets -- -D warnings`
- `cargo test -p nono-cli command_policies_allows_same_fs_rename -- --nocapture`
- `cargo test -p nono-cli outer_exec_gate -- --nocapture`
- Linux Landlock tests are `#[cfg(target_os = "linux")]`. They compile into CI on Linux. This checkout is Darwin, so local execution skips them.

## Checklist

### DIFF
--- a/crates/nono-cli/src/tool-sandbox/platform/linux.rs
+++ b/crates/nono-cli/src/tool-sandbox/platform/linux.rs
@@ -3091,6 +3091,12 @@ fn add_outer_exec_file_with_deps(
     Ok(())
 }
 
+/// Stack a Landlock layer that restricts execute to `paths` and `writable_dirs`.
+///
+/// Also grants bare `Refer` on `/`. Landlock requires Refer in every stacked
+/// layer for rename/link, so omitting it here silently breaks same-FS
+/// renames the outer sandbox already permits (`command_policies` / #1689).
+/// Bare `Refer` alone cannot widen access. Mirrors `restrict_execute`.
 fn apply_outer_exec_gate(
     paths: &[PathBuf],
     writable_dirs: &[PathBuf],
@@ -3112,6 +3118,12 @@ fn apply_outer_exec_gate(
             ))
         })?
         .set_compatibility(CompatLevel::BestEffort)
+        .handle_access(AccessFs::Refer)
+        .map_err(|err| {
+            NonoError::SandboxInit(format!(
+                "tool-sandbox outer exec gate cannot handle Refer: {err}"
+            ))
+        })?
         .create()
         .map_err(|err| {
             NonoError::SandboxInit(format!(
@@ -3150,6 +3162,21 @@ fn apply_outer_exec_gate(
                 NonoError::SandboxInit(format!(
                     "tool-sandbox outer exec gate add_rule for {}: {err}",
                     dir.display()
+                ))
+            })?;
+    }
+
+    if abi.has_refer() {
+        let root_fd = PathFd::new("/").map_err(|err| {
+            NonoError::SandboxInit(format!(
+                "tool-sandbox outer exec gate cannot open / for Refer grant: {err}"
+            ))
+        })?;
+        ruleset = ruleset
+            .add_rule(PathBeneath::new(root_fd, AccessFs::Refer))
+            .map_err(|err| {
+                NonoError::SandboxInit(format!(
+                    "tool-sandbox outer exec gate add_rule for / (Refer): {err}"
                 ))
             })?;
     }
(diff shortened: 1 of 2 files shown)


SequeI (approved):
LGTM, thanks very much for this fix and for adding the appropriate tests!

Will run some more manual tests on my machine in the morning just to confirm the issue is solved and get this merged. Welcome to the nono community!

PR #1733, opened 2026-08-24
feat(broker-path): sanitize PATH for host-side credential and URL brokers

SequeI:
## Linked Issue

Closes #

## Summary

Host-side brokers (open/xdg-open, op/bw/security, ambient and credential_capture command sources) resolve executables by bare name via PATH while running unsandboxed. A sandbox-writable directory that is also on PATH let the sandboxed process plant a trojan for these brokers to execute with real user privileges.

Add nono::sanitize_broker_path_for_binary, checking each PATH entry and the exact resolved binary against the capability set, including symlink-chain hops and file-scoped grants that a directory-only check would miss. Wire it into every broker call site across nono-cli, nono (keystore), and nono-proxy (CredentialStore). Move the shared primitive into the nono library so nono::keystore can use it without crossing the library/CLI boundary.

Document the mechanism in the security model and credential-injection docs, including the standalone nono proxy exemption (no sandboxed child, so no capability set to sanitize against).

## Checklist


connrg (approved):
lgtm

PR #1734, opened 2026-08-24
docs(security): security model clarification 

lukehinds:
This PR repurposes the security model to highlight nono's key capability (fine grained , capability based sandboxing) and how it complements more monolithic isolation technologies such as microvms or containers. 

Various other documents were tuned up to reflect the same. I also drop statements to the effect of "Cannot be modified after application" as we have approval backends now since this was first written.


SequeI (approved):
lgtm, feel free to merge after taking a look through the nogent comments and seeing if any apply. Thanks!

PR #1816, opened 2026-09-07
fix(cli): gate diagnostic remedies on observed evidence

### LINKED ISSUE #1646: Pack guidance routes sandbox-reinit and ordinary permission errors to path grants that cannot apply
### What happened?

Every `nolabs-ai` agent pack ships a `nono-sandbox` skill, and the `codex` pack additionally injects a `developer_instructions` block into `~/.codex/config.toml`. Both tell the agent that a list of permission-error strings means the nono sandbox boundary, and both offer only two remedies: `nono run --allow <path>` and drafting a profile for `nono profile promote`. Both widen the sandbox, and neither has a branch for the failure not being a sandbox problem.

The sandbox-reinit case is the clearest example. From the codex pack's `skills/nono-sandbox/SKILL.md` line 21: "On macOS, `sandbox-exec: sandbox_apply: Operation not permitted` often means Codex tried to apply its own sandbox inside the already-running nono Seatbelt sandbox; treat it as the same capability problem." Nono classifies this failure as a blocked system service, `forbidden-sandbox-reinit`. The denial names no path, so the `--allow` / `--read` / `--write` grants on offer have nothing to attach to, and applying them does not change the outcome.

The description is also narrower than the cause. Nono blocks sandbox re-initialization for anything running under the profile, and in practice that is usually a process the agent spawned rather than the agent itself: SwiftPM evaluating `Package.swift`, xcodebuild's plugin execution, Chrome's zygote, `ttyd` under vhs. Codex is one instance because it wraps each command in `sandbox-exec`. An agent following this guidance will recognise the self-sandboxing case and has nothing to go on for the more common spawned-process case. The fix in each case is to disable the inner sandbox — `swift build --disable-sandbox`, `xcodebuild -IDEPackageSupportDisableManifestSandbox=1 -IDEPackageSupportDisablePluginExecutionSandbox=1`, Chrome `--no-sandbox`, Codex `-c sandbox_mode="danger-full-access"` — and none of those is reachable from the two remedies offered.

`nono run` prints the same path-grant advice for this failure. Having labelled the denial `forbidden-sandbox-reinit`, it still suggests `nono run --allow <path>` and `nono why --path <path>`, so an agent reading nono's own output rather than the pack guidance arrives at the same place.

A second problem is the breadth of the trigger list combined with an instruction not to consider the ordinary causes. The codex `developer_instructions` block says: "When a tool output includes `sandbox-exec: sandbox_apply: Operation not permitted`, `Operation not permitted`, `Permission denied`, `EACCES`, `EPERM`, `landlock`, or `sandbox denied`, treat it as the outer nono OS sandbox boundary, not macOS TCC, Full Disk Access, Unix permissions, chmod/chown/sudo, or a Codex approval issue." `Permission denied`, `EACCES` and `EPERM` are among the most common error strings in ordinary software, and the agent is told in advance that they are never Unix permissions or tool-level errors. The skill's `description` field carries the same conclusion, so it reaches the agent through skill metadata before the body is read.

Two further instructions prevent the agent from checking. "Do not run shell/file checks to discover the active profile; those checks may themselves be blocked by the current sandbox" rules out the verification that would settle it. "For this local dev install, assume the active profile is `codex-dev` unless the user says they launched Codex with a custom profile" hardcodes a guess, so on any other setup the drafted profile extends the wrong parent.

Guidance that leads with verification rather than a conclusion, and keeps a branch for the failure not being a capability problem. Specifically: route `sandbox-exec: sandbox_apply: Operation not permitted` and `forbidden-sandbox-reinit` to disabling the inner sandbox rather than to a path grant; drop the assertion that `Permission denied` / `EACCES` / `EPERM` are never Unix permissions or ordinary tool errors; permit `nono why --self` instead of forbidding checks and hardcoding `codex-dev`; and offer "not a sandbox issue" as an outcome alongside the two widening remedies.

### Steps to reproduce

1. Reproduce the underlying failure directly, with no agent or toolchain involved: `nono run -p <any-profile> -- sandbox-exec -p '(version 1)(allow default)' /bin/echo hi`. It fails with `sandbox-exec: sandbox_apply: Operation not permitted`, exit code 71, and `Sandbox blocked system services: forbidden-sandbox-reinit`. The same command outside `nono run` succeeds. Nono's "Next steps" suggest `--allow <path>` although the denial has no path.
2. `nono pull nolabs-ai/codex` and launch Codex under a nono profile.
3. Trigger the same failure through a build so the agent sees it in tool output — `swift build` without `--disable-sandbox`, or `xcodebuild` on a project with SPM dependencies.
4. The agent reports a nono capability boundary and offers `nono run --allow <path>` or a promoted profile. Neither addresses a sandbox-reinit denial.
5. The same misdiagnosis occurs for `EACCES` / `EPERM` failures unrelated to the sandbox, since the trigger list covers them and the guidance rules out the ordinary causes by name.

### Relevant output or logs

```shell
$ nono run -p codex-local -- sandbox-exec -p '(version 1)(allow default)' /bin/echo hi
  Applying sandbox...
sandbox-exec: sandbox_apply: Operation not permitted
Command exited with code 71.
Sandbox blocked system services:
  forbidden-sandbox-reinit
To grant additional access, re-run with:
  --allow <path>     read+write access to directory
  --read <path>      read-only access to directory
  --write <path>     write-only access to directory
Next steps:
  Add permissions: nono run --allow <path> -- sandbox-exec -p '(version 1)(allow default)' /bin/echo hi
  Query policy: nono why --path <path> --op <read|write|readwrite>

$ sandbox-exec -p '(version 1)(allow default)' /bin/echo hi     # same command, no nono
hi

$ sed -n '21p' ~/.config/nono/packages/nolabs-ai/codex/skills/nono-sandbox/SKILL.md
When you see any of these on a Bash, apply_patch, or MCP file-tool failure, it is a nono boundary — not macOS TCC, not Full Disk Access, not Unix file permissions, not a Codex approval. On macOS, `sandbox-exec: sandbox_apply: Operation not permitted` often means Codex tried to apply its own sandbox inside the already-running nono Seatbelt sandbox; treat it as the same capability problem.

$ sed -n '2p' ~/.config/nono/packages/nolabs-ai/codex/wiring/codex-block.toml
When a tool output includes `sandbox-exec: sandbox_apply: Operation not permitted`, `Operation not permitted`, `Permission denied`, `EACCES`, `EPERM`, `landlock`, or `sandbox denied`, treat it as the outer nono OS sandbox boundary, not macOS TCC, Full Disk Access, Unix permissions, chmod/chown/sudo, or a Codex approval issue.

$ for p in claude codex opencode pi; do printf '%-9s ' $p; grep -cE '\-\-disable-sandbox|sandbox_apply|nest' ~/.config/nono/packages/nolabs-ai/$p/skills/nono-sandbox/SKILL.md; done
claude    0
codex     3
opencode  0
pi        0
```

### nono version

0.73.0

### Operating system

macOS

### OS version / distro

macOS 26.5.2

_No response_

### Profile (if relevant)

```json
Not profile-specific. The skill ships with every pack and the `developer_instructions` block is injected regardless of which profile is active. The reproduction above uses `codex-local` but behaves the same on the built-in `default` profile.
```

### Additional context

The scope is wider than the codex pack. All four packs — `claude`, `codex`, `opencode`, `pi` — ship their own `skills/nono-sandbox/SKILL.md`, each a different file, all with the same two-remedy shape. Only the codex pack also injects `developer_instructions`. The codex skill is the only one of the four that mentions nested sandboxes, and it is the one that misclassifies them; the other three do not raise the possibility. Changing the codex block alone would leave the same failure mode reachable through the skill on all four agents.

Happy to open a PR with replacement wording if that would be useful.

roli-lpci:
## Problem

The supervised diagnostic footer recommended path and network grants even when the session recorded no evidence that either caused the failure. This made application and system-service failures look actionable through unrelated sandbox permissions.

Fixes #1646

## Changes

- show path grant and `nono why --path` guidance only when diagnostics contain path evidence
- show `--allow-net` only when diagnostics contain network-denial evidence
- preserve existing remedies for logged denials, stderr path hints, missing paths, and real network blocks
- add focused regression and positive-control coverage for the evidence-gated branches

## Impact

This changes CLI guidance only; sandbox enforcement, capabilities, policy resolution, and diagnostic records are unchanged.

This contribution was autonomously selected and produced by agents through [Hermes Labs](https://hermes-labs.ai)’ engineering infrastructure. [Rolando Bosch](https://github.com/roli-lpci) is the responsible human contributor and authorized publication from his personal GitHub account.

### DIFF
--- a/crates/nono-cli/src/diagnostic/formatter.rs
+++ b/crates/nono-cli/src/diagnostic/formatter.rs
@@ -1182,10 +1182,14 @@ impl<'a> DiagnosticFormatter<'a> {
         };
         self.format_likely_sandbox_list(&mut lines, additional);
 
-        if self.blocked_protected_file.is_none() && !has_stderr_findings {
+        // Same evidence rule as the supervised footer: path grants and path
+        // queries are prescribed only when this session named a path.
+        if self.blocked_protected_file.is_none()
+            && !has_stderr_findings
+            && self.has_observed_path_evidence(diagnostics)
+        {
             lines.push("[nono]".to_string());
-            self.format_grant_help(&mut lines);
-            lines.push("[nono]".to_string());
+            self.format_grant_help(&mut lines, diagnostics);
             self.format_follow_up_from_diagnostics(&mut lines, diagnostics);
         }
 
@@ -1208,15 +1212,7 @@ impl<'a> DiagnosticFormatter<'a> {
         let system_service_diagnostics = self.system_service_diagnostics(diagnostics);
         let has_path_findings =
             !path_diagnostics.is_empty() || !pathname_unix_diagnostics.is_empty();
-        // Path-specific remedies (`--allow/--read/--write <path>`, `nono why
-        // --path`) only make sense when something in this session named a
-        // path: a logged filesystem denial, a stderr line that looks like a
-        // sandbox denial on a path, or a stderr "No such file" report. A
-        // system-service block or an application error names no path, so
-        // there is nothing for those flags to attach to (issue #1646).
-        let has_observed_path_evidence = has_path_findings
-            || !stderr_likely_sandbox_diagnostics(diagnostics).is_empty()
-            || stderr_missing_path_diagnostic(diagnostics).is_some();
+        let has_observed_path_evidence = self.has_observed_path_evidence(diagnostics);
 
         if !has_path_findings
             && ipc_diagnostics.is_empty()
@@ -1262,8 +1258,7 @@ impl<'a> DiagnosticFormatter<'a> {
             }
             if has_observed_path_evidence {
                 lines.push("[nono]".to_string());
-                self.format_grant_help(&mut lines);
-                lines.push("[nono]".to_string());
+                self.format_grant_help(&mut lines, diagnostics);
                 self.format_follow_up_from_diagnostics(&mut lines, diagnostics);
             } else if stderr_network_diagnostic(diagnostics).is_some() {
                 // Same principle as has_observed_path_evidence: a blocked
@@ -1415,7 +1410,7 @@ impl<'a> DiagnosticFormatter<'a> {
         lines: &mut Vec<String>,
         diagnostics: &[NonoDiagnostic],
     ) {
-        lines.push("[nono] Next steps:".to_string());
+        let mut steps = Vec::new();
         for diagnostic in diagnostics {
             let Some(ref remediation) = diagnostic.remediation else {
                 continue;
@@ -1423,19 +1418,19 @@ impl<'a> DiagnosticFormatter<'a> {
             match remediation {
                 NonoRemediation::RunDiscovery => {
                     if let Some(command) = self.format_command_for_run() {
-                        lines.push(format!(
+                        steps.push(format!(
                             "[nono]   Add permissions: nono run --allow <path> -- {}",
                             command
                         ));
                     } else {
-                        lines.push(
+                        steps.push(
                             "[nono]   Add permissions: nono run --allow <path> -- <your command>"
                                 .to_string(),
                         );
                     }
                 }
                 NonoRemediation::CheckPolicy => {
-                    lines.push(
+                    steps.push(
                         "[nono]   Query policy: nono why --path <path> --op <read|write|readwrite>"
                             .to_string(),
                     );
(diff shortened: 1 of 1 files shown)


SequeI (approved):
LGTM, thank you for the fix and welcome to the community!

PR #1852, opened 2026-09-10
fix(cli): ignore unhandled keys in the denied-path review selector

### LINKED ISSUE #1845: Review denied path approval backend closes on any invalid key press
### What happened?

After I shutdown the sandbox the following menu appears

```shell
 [nono] Review denied paths
  ↑/↓ move  ·  Space cycle  ·  a grant-all  ·  d deny-all  ·  Enter confirm  ·  Esc cancel

  ▶  grant     ~/  (read dirs)
```

when I press an invalid key (like `right arrow`) the menu closes, disappears and is never seen again. This is especially annoying when it was _laborous_ to get to that state.

1. invalid key presses should simply be ignore. only close it with `Esc` (as stated) (or `Ctrl+C` or `Ctrl+D`).
2. print a help message on how to get the menu back after it was closed

### Steps to reproduce

1. run any profile
2. shutdown app

### Relevant output or logs

```shell

```

### nono version

0.75

### Operating system

Linux

### OS version / distro

Debian 13

_No response_

### Profile (if relevant)

```json

```

### Additional context

similar issue: https://github.com/nolabs-ai/nono/issues/1506

MateSaary:
## Linked Issue

Closes #1845

## Summary

The denied-path review selector's escape-sequence reader collapsed every unrecognized sequence into `Key::Char('\x1b')`, which the event loop matched in the same arm as Ctrl+C and treated as cancel. It also read at most 3 bytes, so longer sequences like `ESC [ 1 ; 5 C` left trailing bytes to be misread as further keypresses.

Split `Key::Esc` (a real Esc: nothing followed within the VTIME window) from `Key::Unknown` (consumed but unhandled), added `Key::CtrlD`, and changed the decoder to handle both CSI (`ESC [`) and SS3 (`ESC O`) introducers, consuming through the final byte (`0x40..=0x7e`) with a 32-byte bound. Only Esc, Ctrl+C, and Ctrl+D now cancel; everything else is ignored. 

## Test Plan

Reproduced against the pre-fix binary by running `nono run --profile default -- cat ~/.zshrc` under a PTY and injecting one keypress: `x` redrew the menu, `ESC [ C` erased it. Re-ran the same harness after the fix. `ESC [ C`, `[ D`, `[ 5~` and `[ 1;5C` all redraw and the menu stays; bare Esc, Ctrl+C and Ctrl+D cancel and print the hint; Ctrl+Right followed by `d` then Enter ignores the noise, applies deny-all, and confirms. Added 6 unit tests over the extracted decoder covering those sequences, both arrow-key encodings, whole-sequence consumption, and the malformed-input bound.

## Checklist

- [x] If this PR introduces a major feature, capability, or security-relevant change, a corresponding NEP has been opened or accepted and is linked

### DIFF
--- a/crates/nono-cli/src/profile_save_runtime.rs
+++ b/crates/nono-cli/src/profile_save_runtime.rs
@@ -1087,13 +1087,76 @@ pub(crate) fn configure_prompt_termios(termios: &mut nix::sys::termios::Termios)
 
 // ─── Raw terminal mode for interactive selector ───────────────────────────
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum Key {
     Up,
     Down,
     Space,
     Enter,
     CtrlC,
+    CtrlD,
+    Esc,
     Char(char),
+    /// An input sequence the selector does not act on (left/right arrows,
+    /// Home/End, function keys, Alt-modified keys, mouse reports).
+    Unknown,
+}
+
+fn decode_plain_byte(byte: u8) -> Key {
+    match byte {
+        b' ' => Key::Space,
+        b'\r' | b'\n' => Key::Enter,
+        0x03 => Key::CtrlC,
+        0x04 => Key::CtrlD,
+        c => Key::Char(c as char),
+    }
+}
+
+/// Decode an escape sequence after its leading `ESC` byte.
+///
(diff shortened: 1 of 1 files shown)


SequeI (approved):
lgtm!

Issue #998, opened 2026-05-22
proxy: HTTPS CONNECT closes client socket silently when upstream TCP connect fails


attila:
### What happened?

If `nono-proxy` accepts a `CONNECT host:port` (allowlist passes, DNS resolves) but then can't establish the upstream TCP connection itself, it just drops the client socket. No HTTP response written, no audit log entry. From the outside this is indistinguishable from a nono bug, because the failing host isn't in `nono audit show <session> --json`'s `network_events` at all.

I hit this because Little Snitch was blocking `nono-proxy` from reaching a handful of hosts that my profile allowlisted. The network path was broken, fair enough, but the way nono surfaces that breakage sent me a long way down the wrong debugging path.

Functionally nothing is regressed. When the upstream connect works, behaviour is unchanged. This is purely a diagnostic gap.

### What did you expect to happen?

A `502 Bad Gateway` with the underlying error in the reason line, plus an `audit::log_denied` with `UpstreamConnectFailed`. Same shape as the existing "DNS resolution returned no addresses" branch sitting immediately above the broken code, which already does exactly this.

### Steps to reproduce

Any host firewall that blocks `nono-proxy` from reaching the allowlisted IP reproduces it. On macOS, Little Snitch with a deny rule is the cleanest version. Linux equivalent is `iptables -A OUTPUT -m owner --uid-owner <nono uid> -d <ip> -j REJECT`.

1. Add a Little Snitch deny: process `nono-proxy`, destination `registry.npmjs.org`, port 443.
2. Run a session with a profile that allowlists the host:
   ```jsonc
   { "network": { "allow_domain": ["registry.npmjs.org"] },
     "workdir":  { "access": "readwrite" } }
   ```
3. Inside the sandbox shell:
   ```sh
   curl -v --max-time 15 -w 'TIME: %{time_total}s\n' https://registry.npmjs.org/
   ```
4. No `HTTP/1.1 ...` line from the proxy, `curl: (56) Proxy CONNECT aborted` in ~25 ms, nothing in the audit log for that host.
5. Remove the rule, rerun, get a 200 in ~150 ms. For contrast, hit a non-allowlisted host like `example.com` and the proxy correctly returns a 403 and logs `host_denied`. The clean deny vs silent close difference is the diagnostic value the fix restores.

The same root cause has two client-visible shapes depending on whether the firewall rejects (RST) or drops the SYN:

| Filter behaviour | Symptom                                                | Elapsed                                         |
[...]

PR #1894, opened 2026-09-13
fix(proxy): fail closed when upstream DNS returns no addresses

### LINKED ISSUE #1891: HTTP and TLS intercept re-resolve hostname when first DNS answer is empty
## What happened

`nono-proxy` CONNECT already fails closed when `check_host` returns no resolved addresses (`connect.rs`). The shared HTTP and TLS-intercept path does not.

`ProxyFilter::check_host` allowlists the hostname first. If `lookup_host` fails, it still returns Allow with `resolved_addrs: []`. `open_tcp_upstream` then does `TcpStream::connect("host:port")`, which looks the name up again. That second lookup is not checked against the link-local metadata range.

CONNECT comments and the crate README say callers must connect to the already-checked addresses, not re-resolve.

This is a correctness bug against the documented metadata protection. Official profiles pin concrete hosts, so it is not a default-path advisory.

## Public trigger

`nono run` with a proxy (or `nono proxy` with an empty allowlist / `allow_domain: ["*"]`), then a cleartext `HTTP_PROXY` request to a hostname whose first DNS answer is empty.

## Observed vs inferred

From reading `crates/nono-proxy/src/forward.rs` (empty-addrs branch), `filter.rs` (`lookup_host` Err -> empty vec + Allow), and `connect.rs` (fail closed). CONNECT path is the existing correct sibling.

## Planned fix

Make `open_tcp_upstream` return `ProxyError::UpstreamConnect` with the same "DNS resolution returned no addresses" reason CONNECT already uses. Do not fall back to a hostname connect.

I plan to send a PR once this issue exists.

SebTardif:
## Linked Issue

Closes #1891

## Summary

`open_tcp_upstream` no longer re-resolves a hostname when `resolved_addrs` is empty. HTTP forward and TLS intercept now match CONNECT: empty DNS is `UpstreamConnect`, not a second `TcpStream::connect("host:port")`.

## The problem

`ProxyFilter::check_host` allowlists the name first. If `lookup_host` fails it still returns Allow with an empty address list. CONNECT already treats that as a 502 (`connect.rs`, issue #998). The shared HTTP/TLS helper did a fresh hostname connect, so the second lookup skipped the link-local metadata check.

Introduced in [#856](https://github.com/nolabs-ai/nono/pull/856) ([`149abde0`](https://github.com/nolabs-ai/nono/commit/149abde0)).

## The change

- `crates/nono-proxy/src/forward.rs`: empty `Direct` addrs return the same reason string CONNECT uses.
- `crates/nono-proxy/src/tls_intercept/h2_probe.rs`: drop the stale comment that described the hostname fallback.

## Test Plan

```
cargo test -p nono-proxy open_tcp_upstream_fails_closed -- --nocapture
```

Red (unfixed production): test failed with `failed to lookup address information`.
Green (this commit): test passed.

`cargo fmt` and `cargo clippy -p nono-proxy --all-targets -- -D warnings` are clean.

## Checklist

- [x] Public-facing changes are paired with documentation updates (none; fail-closed behavior only)
- [x] NEP not required (bug fix, no policy or API change)


SequeI (approved):
lgtm!

